Issue #30, opened 2021-09-11
Fix generics in traits


xldenis:
At the moment Creusot translates each trait declaration as a block of functions. Since generics are declared as opaque types within a module, that means that we essentially move from universal quantification to existential quantification for trait / instance methods. 
A more appropriate solution would be for each associated item to be declared in a standalone module.

However, doing so means that we would not load the laws of a trait / instance when we use one of its methods. I think that for the moment this is acceptable but would definitely need to be addressed, perhaps by a heuristic that automatically imports any laws along with trait methods.

PR #41, opened 2021-09-14
Maximally share clones 

### LINKED ISSUE #26: Avoid duplicate clones
Currently calling a function requires cloning, which is _generative_ in why3. This causes issues because we can clone the same function multiple times, with the same type parameters and why3 will not realize that they are in fact the same. 

Consider the following code:

```rust
#![feature(register_tool, rustc_attrs)]
#![register_tool(creusot)]
#![feature(proc_macro_hygiene, stmt_expr_attributes)]

use creusot_contracts::*;

enum List {
    Cons(u32, Box<List>),
    Nil,
}
use List::*;

impl List {
    logic! { fn len(self) -> Int {
          match self {
            Cons(_, ls) => 1 + ls.len(),
            Nil => 0,
          }
        }
    }
}

#[ensures(l.len() == 10)]
fn ensures_len(l : &List) { panic!() }

#[requires(l.len() == 10)]
fn requires_len(l : &List) { panic!() }

// Because each of requires_len & ensures_len clones `len` separately, 
// we cannto prove this obvious goal.
fn test(l: &List) {
    ensures_len(l);
    requires_len(l);
}

fn main() {}

```

which generates 

<details>
<summary>Why3 Code </summary>
```
module Type
  use mach.int.UInt32
  use mach.int.UInt64
  use prelude.Prelude
  type dummy_list  = 
    | Dummy_List_Cons(uint32, dummy_list)
    | Dummy_List_Nil
    
end
module Dummy_Impl0_Len
  use Ref
  use mach.int.Int
  use mach.int.Int32
  use mach.int.Int64
  use mach.int.UInt32
  use mach.int.UInt64
  use prelude.Prelude
  use Type
  let rec function len (self : Type.dummy_list) : int = 
    match (self) with
      | Type.Dummy_List_Cons(_, ls) -> Int32.to_int (1 : int32) + len ls
      | Type.Dummy_List_Nil -> Int32.to_int (0 : int32)
      end
end
module CreusotContracts_Builtins_Resolve
  use Ref
  use mach.int.Int
  use mach.int.Int32
  use mach.int.Int64
  use mach.int.UInt32
  use mach.int.UInt64
  use prelude.Prelude
  use Type
  type self   
  predicate resolve (self : self)
end
module Std_Panicking_BeginPanic
  use Ref
  use mach.int.Int
  use mach.int.Int32
  use mach.int.Int64
  use mach.int.UInt32
  use mach.int.UInt64
  use string.Char
  use floating_point.Single
  use floating_point.Double
  use prelude.Prelude
  use Type
  type m   
  val begin_panic (msg : m) : ()
    ensures { false }
    
end
module Dummy_EnsuresLen
  use Ref
  use mach.int.Int
  use mach.int.Int32
  use mach.int.Int64
  use mach.int.UInt32
  use mach.int.UInt64
  use string.Char
  use floating_point.Single
  use floating_point.Double
  use prelude.Prelude
  use Type
  clone CreusotContracts_Builtins_Resolve as Resolve0 with type self = Type.dummy_list
  clone Std_Panicking_BeginPanic as BeginPanic1 with type m = string
  clone Dummy_Impl0_Len as Len2
  let rec cfg ensures_len (l : Type.dummy_list) : ()
    ensures { Len2.len l = Int32.to_int (10 : int32) }
    
   = 
  var _0 : ();
  var l_1 : Type.dummy_list;
  var _2 : ();
  {
    l_1 <- l;
    goto BB0
  }
  BB0 {
    assume { Resolve0.resolve l_1 };
    absurd
  }
  
end
module Dummy_RequiresLen
  use Ref
  use mach.int.Int
  use mach.int.Int32
  use mach.int.Int64
  use mach.int.UInt32
  use mach.int.UInt64
  use string.Char
  use floating_point.Single
  use floating_point.Double
  use prelude.Prelude
  use Type
  clone CreusotContracts_Builtins_Resolve as Resolve0 with type self = Type.dummy_list
  clone Std_Panicking_BeginPanic as BeginPanic1 with type m = string
  clone Dummy_Impl0_Len as Len2
  let rec cfg requires_len (l : Type.dummy_list) : ()
    requires {Len2.len l = Int32.to_int (10 : int32)}
    
   = 
  var _0 : ();
  var l_1 : Type.dummy_list;
  var _2 : ();
  {
    l_1 <- l;
    goto BB0
  }
  BB0 {
    assume { Resolve0.resolve l_1 };
    absurd
  }
  
end
module Dummy_Test
  use Ref
  use mach.int.Int
  use mach.int.Int32
  use mach.int.Int64
  use mach.int.UInt32
  use mach.int.UInt64
  use string.Char
  use floating_point.Single
  use floating_point.Double
  use prelude.Prelude
  use Type
  clone Dummy_EnsuresLen as EnsuresLen0
  clone CreusotContracts_Builtins_Resolve as Resolve1 with type self = Type.dummy_list
  clone Dummy_RequiresLen as RequiresLen2
  let rec cfg test (l : Type.dummy_list) : () = 
  var _0 : ();
  var l_1 : Type.dummy_list;
  var _2 : ();
  var _3 : Type.dummy_list;
  var _4 : ();
  var _5 : Type.dummy_list;
  {
    l_1 <- l;
    goto BB0
  }
  BB0 {
    _3 <- l_1;
    _2 <- EnsuresLen0.ensures_len _3;
    goto BB1
  }
  BB1 {
    assume { Resolve1.resolve _3 };
    _5 <- l_1;
    assume { Resolve1.resolve l_1 };
    _4 <- RequiresLen2.requires_len _5;
    goto BB2
  }
  BB2 {
    assume { Resolve1.resolve _5 };
    _0 <- ();
    return _0
  }
  
end
module Dummy_Main
  use Ref
  use mach.int.Int
  use mach.int.Int32
  use mach.int.Int64
  use mach.int.UInt32
  use mach.int.UInt64
  use string.Char
  use floating_point.Single
  use floating_point.Double
  use prelude.Prelude
  use Type
  let rec cfg main () : () = 
  var _0 : ();
  {
    goto BB0
  }
  BB0 {
    _0 <- ();
    return _0
  }
  
end
```
</details>

`requires_len` and `ensures_len` both clone `len` independently but then the final caller, `test` cannot tell that they are the same, we end up with a proof obligation saying `len1 l = 10 -> len2 l = 10`, though the bodies of `len1` and `len2` are identical. 

The solution is to circumvent the generative nature of clones in Why3, one way or another, to only have a single `len` used by both modules here.

xldenis:
This PR closes #26.

In Why3, clones are _generative_, they create a fully fresh definition each time they are evaluated. Additionally, in Why3, terms do not respect alpha equivalence. If two definitions differ only by name, they are _different_ no matter what. 
The solution this PR implements is a mechanism to avoid cloning the same module multiple times at all, and instead _maximally_ share clones. 

# Context

As a quick reminder, in Creusot, each function is translated to a why3 module which clones all of its dependencies, _substituting_ type parameters in the clone.
This means that we will clone a function for each unique substitution it is applied to, and this PR does not attempt to change this. 
However, this PR solves the following issue, imagine we have:

```rust

logic! {
  fn my_precond(x: u32) -> bool {
   something_here
  }
}

#[requires(my_precond(x))]
fn a(x: u32)  { .. }

#[requires(my_precond(x))]
fn b(x : u32) { a(x) }
```

In the module for `b`, we will clone `my_precond`, a dependency from the contract, and `a` a dependency from the definition of `b`. However, by cloning `a`, we will also trigger a _second_ clone of `my_precond` from within the module of `a`. 
This leads to an issue when we attempt to prove the precondition of `a`. In our context we will have `my_precond0(x)` but we are asked to prove `my_precond1(x)`!

# Solution

The solution to this problem is in two parts. 

1. Introduce interfaces 
2. Maximal sharing of clones through said interfaces

Each function, logical or program, is given an _interface_. The interface describes what is visible from a caller: the function's signature and contract, nothing else. If a contract depends on other functions (ie: `my_precond`) then those interfaces are cloned. However, the interface is _abstract_, the actual symbols contained within are `val` in Why3, they have no body and can be _refined_. 

Additionally, each function is given an _implementation_, which defines the actual _body_ of the function. Here is where things get a little subtle as we behave differently for logic and program functions. 

- For logic functions, the body clones the interfaces of all dependencies. This leaves all dependencies in the body abstract. 
- For program functions, we clone the _interface_ of program functions but the _body_ of logic functions. This distinction occurs because we want logic functions to be transparent in proofs, we need to be able to reason on their definitions for them to be useful. 

So far, we have basically reproduced the previous situation but with extra complications, however, by introducing interfaces, we have opened a door to _sharing_. When performing a clone, why3 allows us to refine the definitions of abstract symbols by providing a concrete instantiation for them.

As we've always left dependencies abstract in interfaces and in the definitions of logic function bodies, that means that we can carefully construct a substitution which will share a single definition as widely as possible. 

To do this we construct a sharing graph in which nodes are function instantiations (DefId + Subst pairs), and edges represent dependencies. 
Then we traverse this graph in topological order, at each node, we look at all edges we depend on and add them to the substitution for the current node. 
This ensures that we fully share instantiations throughout the entire call graph. 

--- 

Because of how this is currently implemented we do not support calculating dependencies of functions that come from other crates, but I would rather address this in a follow up PR as this currently fixes a large amount of issues in current proofs. 
I have a relatively clear plan on how to restore support for multi-crate projects.

### DIFF
--- a/creusot/src/clone_map.rs
+++ b/creusot/src/clone_map.rs
@@ -1,18 +1,24 @@
+use std::ops::ControlFlow;
+
 use indexmap::{IndexMap, IndexSet};
 
+use petgraph::EdgeDirection::Incoming;
 use why3::declaration::{CloneKind, CloneSubst, Decl, DeclClone, Use};
 use why3::QName;
 
 use rustc_hir::def_id::DefId;
 use rustc_middle::ty::{
     self,
-    subst::{InternalSubsts, SubstsRef},
-    TyCtxt,
+    fold::{TypeFoldable, TypeVisitor},
+    subst::{InternalSubsts, Subst, SubstsRef},
+    AssocKind, ProjectionTy, Ty, TyCtxt, TyKind,
 };
 
 use heck::CamelCase;
 
 use crate::ctx::{self, *};
+use crate::translation::interface;
+use crate::util::{self, method_name};
 
 // Prelude modules
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
@@ -48,39 +54,44 @@ impl PreludeModule {
     }
 }
 
+#[derive(Clone)]
 pub struct CloneMap<'tcx> {
     tcx: TyCtxt<'tcx>,
     prelude: IndexSet<PreludeModule>,
     names: IndexMap<(DefId, SubstsRef<'tcx>), CloneInfo>,
     count: usize,
+    item_type: ItemType,
 }
 
 #[derive(Clone)]
 pub struct CloneInfo {
     name: String,
-    projs: Vec<CloneSubst>,
     hidden: bool,
 }
 
 impl CloneInfo {
     fn from_name(name: String) -> Self {
-        CloneInfo { name, projs: Vec::new(), hidden: false }
+        CloneInfo { name, hidden: false }
     }
 
     fn hidden(name: String) -> Self {
-        CloneInfo { name, projs: Vec::new(), hidden: true }
+        CloneInfo { name, hidden: true }
     }
 
     // TODO: When traits stop holding all functions we can remove the last two arguments
     pub fn qname(&self, tcx: TyCtxt, def_id: DefId) -> QName {
-        QName { module: vec![self.name.clone()], name: ctx::method_name(tcx, def_id) }
+        self.qname_raw(method_name(tcx, def_id))
+    }
+
+    fn qname_raw(&self, method: String) -> QName {
+        QName { module: vec![self.name.clone()], name: method }
     }
 }
 
 impl<'tcx> CloneMap<'tcx> {
-    pub fn new(tcx: TyCtxt<'tcx>) -> Self {
+    pub fn new(tcx: TyCtxt<'tcx>, item_type: ItemType) -> Self {
         let names = IndexMap::new();
-        CloneMap { tcx, names, count: 0, prelude: IndexSet::new() }
+        CloneMap { tcx, names, count: 0, prelude: IndexSet::new(), item_type }
     }
 
     pub fn insert(&mut self, mut def_id: DefId, subst: SubstsRef<'tcx>) -> &CloneInfo {
@@ -100,12 +111,6 @@ impl<'tcx> CloneMap<'tcx> {
         })
     }
 
-    pub fn qname_for_mut(&mut self, def_id: DefId, subst: SubstsRef<'tcx>) -> QName {
-        let tcx = self.tcx;
-        let module = self.insert(def_id, subst);
-        module.qname(tcx, def_id)
-    }
-
     pub fn clone_self(&mut self, self_id: DefId) {
         let mut modl = ctx::translate_value_id(self.tcx, self_id);
         let clone_name = if modl.module.is_empty() { modl.name } else { modl.module.remove(0) };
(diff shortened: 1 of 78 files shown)


shiatsumat:
Looking forward to this!!

creusot/src/clone_map.rs
```diff
@@ -48,72 +54,61 @@ impl PreludeModule {
...
+        CloneInfo { name, hidden: true }
     }
-}
 
-impl<'tcx> CloneMap<'tcx> {
-    pub fn new(tcx: TyCtxt<'tcx>) -> Self {
-        let names = IndexMap::new();
-        CloneMap { tcx, names, count: 0, prelude: IndexSet::new() }
+    // TODO: When traits stop holding all functions we can remove the last two arguments
+    pub fn qname(&self, tcx: TyCtxt, def_id: DefId) -> QName {
```
xldenis, line 82: The rather ugly type of this function will require #30 to be addressed first before it can be properly fixed.

creusot/src/clone_map.rs
```diff
@@ -48,72 +54,61 @@ impl PreludeModule {
...
-            self.names[&(def_id, subst)].name.clone()
-        }
-    }
-
-    pub fn qname_for(&self, def_id: DefId, subst: SubstsRef<'tcx>) -> Option<QName> {
-        let module = self.name_for(def_id, subst)?;
-        Some(QName { module: vec![module], name: ctx::method_name(self.tcx, def_id) })
-    }
-
-    pub fn qname_for_mut(&mut self, def_id: DefId, subst: SubstsRef<'tcx>) -> QName {
```
xldenis, line 114: I felt like it was better to move to an `Entry` like api, rather than directly returning a `QName`, we return a reference to a `CloneInfo` from which we can build a `QName`, perhaps that is overkill now. Originally, it was meant to allow a clone to be marked as 'public' (part of the interface) or not.

creusot/tests/should_succeed/binary_search.stdout
```diff
@@ -246,48 +289,80 @@ module BinarySearch_Impl0_Len
...
+  clone CreusotContracts_Builtins_Resolve as Resolve4 with type self = uint32
+  clone CreusotContracts_Builtins_Resolve as Resolve3 with type self = Type.binarysearch_list uint32
   clone BinarySearch_Get as Get11 with type t = uint32
+  clone BinarySearch_IsSorted as IsSorted10 with function Get0.get = Get11.get
+  clone BinarySearch_GetDefault as GetDefault1 with type t = uint32, function Get0.get = Get11.get
+  clone BinarySearch_LenLogic as LenLogic0 with type t = uint32
+  clone BinarySearch_Impl0_Index_Interface as Index8 with type t = uint32,
+  function LenLogic0.len_logic = LenLogic0.len_logic, function Get1.get = Get11.get
+  clone BinarySearch_Impl0_Len_Interface as Len2 with type t = uint32,
+  function LenLogic0.len_logic = LenLogic0.len_logic
```
xldenis, line 365: Here is the key bit: we share the definition of `Get1` between `IsSorted10`, `GetDefault1` and `Index8`!

creusot/src/clone_map.rs
xldenis: If each clone only introduces a single symbol, we can cache that symbol name in the `CloneInfo`, I believe that the solution to #30 will require splitting up traits / implementations into one module per symbol so this would address this issue.

xldenis:
@shiatsumat I'm going to merge this for the moment, your comments can be addressed post-hoc.

PR #55, opened 2021-09-15
Add test logic_funs

shiatsumat:


shiatsumat:
I noticed the existing test `logic_functions`. So this extension moved to #57.

PR #57, opened 2021-09-15
Add more examples to test logic_functions

shiatsumat:


xldenis:
I wonder: is it better for us to have many small test files (like for traits) or fewer larger ones? What's clearer, etc?

creusot/tests/should_succeed/logic_functions.rs
```diff
@@ -17,5 +17,22 @@ logic! {
...
+            -n + n - n * n / n
+        } else {
+            n
+        }
+    }
+}
+
+logic! {
+    fn deref_pat<'a>(p: &'a (Int, Int)) -> Int {
+        let (a, b) = p;
```
xldenis, line 32: I don't think this is testing `DerefPat`. It should say `(&a,&b)` instead.

xldenis (approved):
1 comment but lgtm

creusot/tests/should_succeed/logic_functions.rs
shiatsumat: This is actually using `DerefPat`, turning `& (Int, Int)` into `(&Int, &Int)`.

shiatsumat: But I'll update this test to pattern-match over `enum`.

shiatsumat:
Oh, finally it works. (I should go to bed now...)

creusot/tests/should_succeed/logic_functions.rs
```diff
@@ -6,16 +6,35 @@
...
+        }
+    }
 }
 
-#[ensures(logical())]
-fn main() {}
+logic! {
+    fn deref_pat<'a>(o: &'a Option<Int>) -> Int {
+        match o {
+            Some(a) => *a,
```
xldenis, line 36: are you certain that this triggers the `DerefPat` branch? I really thought it was from using `&` in a pattern.

creusot/tests/should_succeed/logic_functions.rs
xldenis: ahh ok! ignore my comments then.

PR #58, opened 2021-09-15
Modify lookup_builtin to support mul/div/neg_int

shiatsumat:
Follow-up for #56.

### DIFF
--- a/creusot/src/translation/builtins.rs
+++ b/creusot/src/translation/builtins.rs
@@ -41,6 +41,11 @@ pub fn lookup_builtin(
         let a = args.remove(0);
 
         return Some(Exp::UnaryOp(UnOp::Neg, box a));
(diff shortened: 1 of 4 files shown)


creusot/src/translation/builtins.rs
```diff
@@ -30,11 +30,17 @@ pub fn lookup_builtin(
         let r = args.remove(0);
 
         return Some(Exp::BinaryOp(BinOp::Sub, box l, box r));
-    } else if def_id == ctx.tcx.get_diagnostic_item(Symbol::intern("le_int")) {
```
xldenis, line 33: Don't forget to add support for `le` back!

xldenis (commented):
Fix the `le` translation and the diff will shrink quite a bit

creusot/src/translation/builtins.rs
shiatsumat: Yeah, I missed that. I have reverted the `le_int` case now.

PR #73, opened 2021-09-24
Fix README.md on rust-toolchain

shiatsumat:
In order to have `cargo` read `REPO/rust-toolchain`, we should be within `REPO`.
This may be a pitfall for users, so I added a note on this in `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ More examples are found in [creusot/tests/should_succeed](creusot/tests/should_s
     - [Install `rustup`](https://www.rust-lang.org/tools/install), to get the suitable Rust toolchain
 2. Build **Creusot**
     - Run `$ REPO/build`
+      * The current directory should be somewhere in `REPO` to have `cargo` read `REPO/rust-toolchain`.
 3. Set up **Why3**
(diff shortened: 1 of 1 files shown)


shiatsumat:
I tried to find a way to specify the path to `rust_toolchain` file for `cargo`, but I couldn't.
Probably just encouraging to be within `REPO` when we use `build` or `mlcfg` is good enough for now.

xldenis:
I'm not sure I understand the issue? Cargo returns to the root of a workspace to run commands so it should pick up the `rust-toolchain` file there.

shiatsumat:
The problem is we can't call `cargo` from outside `REPO`, say `../REPO`.

We can still specify the path to `Cargo.toml` and the target directory `target` by giving options `--manifest-path $SCRIPTPATH/Cargo.toml` and `--target-dir $SCRIPTPATH/target` to `cargo`,
but we can't specify `rust_toolchain`.

xldenis:
Ah! You could make your script move itself inside the correct path in a subshell. There's also the option of `cargo install`ing creusot so you have access to it globally.

shiatsumat:
Ah, I see. After some search, I found it nice to use `pushd`/`popd` for that purpose.
I'll update the `build/mlcfg` commands instead in another PR.

PR #74, opened 2021-09-24
Fix commands build and mlcfg to temporarily move into the base directory

shiatsumat:
In order to make `cargo` properly work for the commands `build` and `mlcfg`, we modified the scripts using `pushd`/`popd`.


xldenis (approved):
lgtm

PR #85, opened 2021-09-26
Add tests for Cell

xldenis:
cc @shiatsumat.

This has identified a minor issue in the handling of traits, it seems like Creusot doesn't select the instance of `Inv` for `Even` in `add_two`. However, apart from this minor detail (and machine arith issues), this works :D


xldenis:
I'm going to merge this basic example as a proof of concept that Cell can work for now. @shiatsumat if you want to add more examples feel free!

Issue #28, opened 2021-09-09
Add a `#[pure]` annotation 


xldenis:
Some functions are pure and could be used from both the logic and the program code. To support these functions we should add an annotation `#[pure]` which verifies this property and makes the symbol available in the logic and program. 
These functions would not have access to purely logical operations like quantifiers or `^`. 

It turns out that `logic!` does most of this already, it emits a `let rec function` which does exactly this in why3 but it also erroneously gives access to logical operations. 
`logic!` should be changed to emit only a `function` declaration and we should add a `#[pure]` annotation which emits the previous `let rec function`.

xldenis:
I meant to open a PR for this but 8537c7ec48b015232692b11fc1b91eadac92e288 fixes the first part of this issue, `logic!` now generates only logical symbols. 

I think that adding a `#[pure]` annotation should wait until I've fixed the issue that is plaging function calls / polymorphism.

shiatsumat:
Is the `#[pure]` function what is introduced as a `hybrid` function in the current `README.md`?

I also wonder if the current `logic!` macro can be updated into an attribute macro like `#[pure]`.
https://doc.rust-lang.org/reference/procedural-macros.html#attribute-macros

PR #93, opened 2021-09-28
Add `inc-some-*` tests

shiatsumat:


shiatsumat:
Because I wanted to use `rand::random` in tests, I let `creusot-contracts` re-export the `rand` crate. I believe this is convenient.

xldenis:
That seems alright to me.

creusot/tests/should_succeed/inc_some_tree.rs
```diff
@@ -0,0 +1,66 @@
...
+    Node(tl, a, tr) => sum_tree(*tl) + Int::from(a) + sum_tree(*tr),
+    Leaf => 0,
+  }
+  }
+}
+
+/* TODO: prove this lemma */
+#[trusted]
+#[ensures(sum_tree(*t) >= 0)]
+fn lemma_sum_tree_nonneg(t: &Tree) {}
```
xldenis, line 26: i think you should be able to prove these trees by doing induction in the function body. If you do 

```rust
fn lemma_sum_tree_nonneg(t: &Tree) {
match t {
Leaf => (),
Node(box l, n, box r) =>{ lemma_sum_tree_nonneg(l); lemma_sum_tree_nonneg(r) }
}
```

shiatsumat: Ah, I see.

shiatsumat: The problem is that this makes `lemma_sum_tree_nonneg` actually non-trivial as Rust code, taking unboundedly many program steps.
It is nice if we can mark lemma functions like `lemma_sum_tree_nonneg` as ghost code and eliminate them in the actual executable.

shiatsumat: Is this going to be supported in Creusot?

xldenis: >  It is nice if we can mark lemma functions like lemma_sum_tree_nonneg as ghost code and eliminate them in the actual executable.

Agreed. We (Claude, JH and me) discussed this this morning. I have a clear plan on how to add 'logic functions with contracts' which is what you need.

xldenis: The solution is actually what's outlined in #28, but we only use the logical part.

xldenis: I'd say give it 1-2 weeks for addition but if you write your "lemma"s in this style you would only need to add a `#[pure]` attribute to be able to use them in a _true_ ghost context.

shiatsumat: OK, then I'll write lemmas like that and just comment out the `#[pure]` annotations for now.

PR #96, opened 2021-09-28
Clean up associativity and precedence in printing

### LINKED ISSUE #92: Invalid MLCFG outputs due to improper printing of `let ... in ...`
Currently, Creusot sometimes outputs invalid mlcfgs due to improper syntax printing of `let ... in ...`.

For example, when Creusot inputs the following Rust code,
```rust
#[ensures(^x.0 == *x.0)]
fn drop_pair(x: (&mut u32, &mut u32)) {}
```
it outputs the following MLCFG.
```ocaml
module DropPair_DropPair_Interface
  use prelude.Prelude
  use mach.int.Int
  use mach.int.UInt32
  val drop_pair (x : (borrowed uint32, borrowed uint32)) : ()
    ensures {  ^ let (_, a) = x in a =  * let (_, a) = x in a }
```
This is invalid, because the body of the first `let` has wrongly become `a =  * let (_, a) = x in a`,
while what we intend is actually `^ (let (_, a) = x in a) = * (let (_, a) = x in a)`.

As seen here, usually we should add parentheses around `let ... in ...`.

xldenis:
Closes #92. I copied the precedence levels from the why3 manual so we should be in much better hands now.

### DIFF
--- a/creusot/tests/should_succeed/sum.stdout
+++ b/creusot/tests/should_succeed/sum.stdout
@@ -18,7 +18,7 @@ module Sum_SumFirstN_Interface
   use mach.int.Int
   use mach.int.UInt32
   val sum_first_n (n : uint32) : uint32
-    ensures { result = div (n * n + (1 : uint32)) (2 : uint32) }
+    ensures { result = div (n * (n + (1 : uint32))) (2 : uint32) }
     
 end
 module Sum_SumFirstN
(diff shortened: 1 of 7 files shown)


creusot/tests/should_succeed/sum.stdout
```diff
@@ -28,7 +28,7 @@ module Sum_SumFirstN
   clone CreusotContracts_Builtins_Resolve as Resolve1 with type self = bool
   clone CreusotContracts_Builtins_Resolve as Resolve0 with type self = uint32
   let rec cfg sum_first_n (n : uint32) : uint32
-    ensures { result = div (n * n + (1 : uint32)) (2 : uint32) }
+    ensures { result = div (n * (n + (1 : uint32))) (2 : uint32) }
```
xldenis, line 31: huh, that explains things: the printer was outputting the wrong expressions before!

PR #98, opened 2021-09-28
Fix the format of tests

shiatsumat:
Fix the format of tests.
I did `rustfmt **/*.rs` in the top directory (this automatically reads `rustfmt.toml`) and then manually modified some tests.

I wonder if there's a way to enforce formatting over tests by `cargo fmt` (whereby we can let CI check the format of tests).


xldenis:
You can add the `rustfmt` call to `.github/workflows/rust.yml` add it after the `cargo fmt` one

.github/workflows/rust.yml
```diff
@@ -15,7 +15,9 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Rust fmt
-      run: cargo fmt -- --check
+      run: |
+        cargo fmt -- --check
+        rustfmt **/*.rs --check
```
xldenis, line 20: Maybe you need to split this into two separate `run` invocations?

shiatsumat:
I got ``file `**/*.rs` does not exist`` when I added `rustfmt **/*.rs --check` to CI's `fmt` job.

.github/workflows/rust.yml
shiatsumat: That's not what causes the error.

shiatsumat:
OK, let's just use `cargo fmt` for now.

xldenis:
Ah dang.

xldenis:
I feel like it's a strong escaping issue... maybe quotes would fix it?

xldenis:
You can merge this now if you want though

shiatsumat:
I finally found out the problem and the solution: the globstar `**` is not active by default, and to enable that we just need to call `shopt -s globstar`.

Note that `shopt` is supported in Unix but not in Mac OS.
I added the `fmt` command for formatting, where I decided to use `zsh` instead of `bash` to enable the globstar `**` by default. This requires `zsh` but I think it is pretty portable.

PR #100, opened 2021-09-29
Re-export the rand crate in creusot-contracts

shiatsumat:
Re-export the rand crate in the crate `creusot-contracts` for convenience.
This lets us naturally use `rand::random` in benchmarks.
Also add the test `rand` to check that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -138,6 +138,7 @@ name = "creusot-contracts"
 version = "0.1.0"
 dependencies = [
  "creusot-contracts-proc",
+ "rand",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


xldenis (approved):
wfm. i suppose this is particularily because you might want to actually execute some of these examples?

shiatsumat:
Yeah, I'd like to keep benchmarks executable as much as possible.
Although we can write fake random functions like `fn fake_random() -> bool { true }`, it's nice if we can use an actual random function `rand::random`.

That said, the re-export proposed here is a quick workaround and subject to future changes.

xldenis:
yea, that's totally fine with me, we should probably add a passthrough mode to all the logic macros so that when you compile with normal rustc they disappear. (That's an easy `#[cfg(creusot)]` on to switch the definitions in `creusot-contracts-proc`)

Issue #103, opened 2021-09-30
Update to newer Rust nightly


shiatsumat:
Creusot currently supports `nightly-2021-08-30`, but things change fast in Rust.
It would be nice if Creusot is updated to the newer Rust nightly.
This is particularly because `aarch64-apple-darwin` (which I mainly use recently) is better supported in the latest versions.
It seems that Rustc has undergone some big changes, such as split of the `rustc_mir` crate into `rustc_mir_build`, `rustc_mir_dataflow`, and `rustc_mir_transform`.

xldenis:
oh fun. but yes agreed, I wanted to do the update so I could use `def_path_hash_to_def_id` anyways.

shiatsumat:
Finished by #104.

PR #104, opened 2021-09-30
Update rustc to 25-09-2021

xldenis:
This PR updates the toolchain to 25-09-2021, addressing #103. 

This was a slightly challenging update as some crates where broken up and apis were shuffled around. 
However, the test updates are fairly important. It leads me to think if there isn't a way we could have a little more stable ordering of definitions in the test output, we're currently too affected by internal rustc changes.

cc @shiatsumat


shiatsumat (approved):
Wow it works! Great!

shiatsumat:
Thank you so much for the very quick update!!
Now `rls` and `rustc-src` work in `aarch64-apple-darwin`, which really helps me.

xldenis:
@shiatsumat i recommend switching from `rls` to `rust-analyzer` it's the replacement and has many more features

shiatsumat:
I switched the VSCode extension from `rust-lang.rust` to `matklad.rust-analyzer`.
Yeah, `rust-analyzer` really provides many nice features! Thanks!

PR #106, opened 2021-10-01
Move pearlite parser to library

xldenis:
This PR creates a `pearlite-syn` crate which implements the `syn` based parser from pearlite. This is massive and ugly, but now we can at least in _theory_ update `syn` :). 

I think we can improve the grammar of `pearlite` a bit, cut unused cases, etc... but this is a decent start.


shiatsumat (approved):
Big bold step! Looks good to me.

Issue #63, opened 2021-09-20
Add `#[modeled_by=X]` attribute


xldenis:
When using 'unsafe types' we want to give them a model, or _representation value_ in the formalization which will be used during proofs. 

The way I'm proposing to do this at the moment is to add an attribute which can be attached to type declarations which specifies which type models a given type. This does not create new verification conditions, and is meant to be used only in cases where we have separately proved the model in RustHornBelt or some other mechanism. Additionally, it should be subsumed by a proper ghost field & type invariant system (once we figure out how that works). 

What does this attribute do? Well,...

```rust
#[model=List<T>]
struct Vec<T> { 
  spooky_pointers: *mut T
}
```

Simply generates an opaque predicate `model` on Vec:

```rust
struct Vec<T> { ... }

impl Vec<T> {
  predicate! { fn model(self) -> List<T>; }
}
```

This allows us to refer to the model of a vector in contracts like so:

```rust

impl Vec<T> {
  #[ensures(^self.model() = *.self.model().snoc(e))]
   fn push(&mut self, e: T) -> { ... } 
  ...
}
```

xldenis:
I believe that this can simply be a design pattern if we have a `#[trusted]` attribute. if we mark a predicate as `#[trusted]` it would generate an opaque predicate in why3 which is exactly what we want here.

shiatsumat:
How is this going?

xldenis:
I think the approach I've settled on for now is to do this through a design pattern instead. ie:

```rust
impl Vec<T> {
  #[trusted]
  logic! { fn model(self) -> Seq<T> { panic!() } }
}
```

shiatsumat:
Yeah, the `model` function workaround perfectly works, but I was wondering if adding `modeled_by` is doable.
We prefer to use Creusot's actual specs for the paper, and it's great if the specs are free of the `.model()` parts.

xldenis:
Perhaps, I'm just worried that it would be ad-hoc. Let's discuss this on Friday, I think jh had a related concern

xldenis:
Yes I understand, I agree with that approach as its more lightweight than something like `modeled_by` which would require _deep_ integration into creusot. The advantage of this is that we only need to adjust the `pearlite_syn` parser to do some rewriting for us.

PR #121, opened 2021-10-07
Introduce the `Model` trait and add the `@` notation

shiatsumat:
Introduce the `Model` trait and add the `@` notation, as mentioned in the recent comments of #63.
I think this provides a really good interface for handling `model`s.


xldenis:
Ok I've identified the issue, I have no idea why it only appeared now... it seems like it should have been affecting us much earlier.

shiatsumat:
I also tried making `model` local (translating `@x` simply as `model(x)` without no quantifier) in the last commit 33e30e9.
Again, this doesn't work. Compilation to MLCFG works but Why3 emits the same type error as before.
```
This term has type Model_Seven_Interface.Model0.Model0.model,
but is expected to have type int
```
So the problem is not fixed.

shiatsumat:
Just to check, I tried the no-trait approach. It means that we only have a notation `@x` standing for `x.model()` and have to locally write stuff like the following.
```rust
impl Seven {
    logic! {
        #[trusted]
        fn model(self) -> Int {
            panic!()
        }
    }
}
```
This is not very satisfying.
In particular, when we use a method `x.model()` instead of `...::model(x)`, we have tricky type inference problems.
(E.g., I had to write `@*(old_v.model(): &mut MyVec<u32>)` instead of `@*@old_v` in `vector/01.rs`.)

jhjourdan:
Well but the obvious downside of this is that we cannot define a model for types where we actually want to define a model.

But anyway, why do we want to have a default implementation at all? Cannot we simply require a method `model` in the trait `Model` (without a default), and trivially implement it with `#[trusted]` in e.g., vectors? It's not like this trivial definition is going to take much space anyway...

xldenis:
you need to use `logic! ` in the trait declaration as well.

Incidentally I found the source of the crash with `pub`. it's rather hilarious.. it comes from `panic!()` the actual error is that it can't parse the macro invocation!!! if you replace it with `std::process::abort` it compiles.

xldenis:
> > you need to use logic! in the trait declaration as well.
> 
> Would it be possible to have a check for that in Creusot (i.e., an error if the trait declaration and implementation are not in sync wrt. `logic!`)?
> 

Yes, it should definitely be enforced by creusot and is a fairly simple feature to add, we just need to check the attributes of the declaration and definitions.

> > Incidentally I found the source of the crash with pub. it's rather hilarious.. it comes from panic!() the actual error is that it can't parse the macro invocation!!! if you replace it with std::process::abort it compiles.
> > I'll open an issue to allow macros in pearlite terms.
> 
> In any case, we should produce an error message if an unsupported Rust feature is used in a logic function.

Yes, I've opened an issue for this. The first issue here is that we were swallowing the _good_ error message.

jhjourdan:
As for `pub` on logic functions: the issue, I think, is that there are two kinds of visibility that we might want:
- Either we make public both the function and its content
- Or we make public the function, but hide its content to the outside world (this is basically what we want for model).

What kind of syntax should we use to do this distinction? Should we use the `pub` keyword at the beginning of the body of the function?

xldenis:
Perhaps something like `#[opaque]`? I feel like this discussion belongs to a separate issue.

PR #129, opened 2021-10-08
Add `#[logic_rust]` and `#[predicate_rust]` attributes

shiatsumat:
This PR adds `#[logic]` attribute and renames `logic!` to `logic_fn!`.
You can use `#[logic]` when the function body does not use pearlite syntax. In this way, we can get better IDE support.
When we want to use pearlite syntax, we use `logic_fn!` macro (old `logic!` macro).

I also updated `README.md` to give better explanation on `logic` and `pure`.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -151,7 +151,7 @@ Instead, we replace the `discriminant` / `switchInt` pair with a match directly 
 
 ## Logical functions
 
-Logical functions are created by `#[logic]`/`logic_fn!`, `#[pure]`, or `#[predicate]`/`predicate_fn!`.
+Logical functions are created by `#[logic_rust]`/`logic!`, `#[pure]`, or `#[predicate_rust]`/`predicate!`.
 
 ## Specifications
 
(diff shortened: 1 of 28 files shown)


jhjourdan:
I'm not so sure that's a good idea: if we do so, the user will need to remember what is part of pearlite syntax and what is part of simple Rust syntax, which surely will lead to confusion. Moreover, if the user uses `#[logic]` while using pearlite syntax, she will get a Rust syntax error, which will probably not going to help.

xldenis:
Can you do the same for predicate? 

I was also thinking: it might be better to just introduce a `pearlite` macro which parses the body of a function/an expression.

xldenis:
Related: I want to submit an RFC to rust which weakens the requirements for proc macro attributes to allow syntactically invalid (for rust) items.

shiatsumat:
By the way, how is `predicate!` different from `logic!`?
Currently `README.md` doesn't mention `predicate!`, so I'd like to add some nice explanation.

xldenis:
Ok, I'm sorry I'm kind of waffling on this issue: when do we want to use `#[logic]` but can't use `#[pure]`? Is it just when we manipulate a 'logical type'?

xldenis:
> First I guess #[logic] makes MLCFG simpler than #[pure] and helps verification more or less.

I don't think this is true, granted more mlcfg code is generated but for verification all the bits about the 'program' part of `pure` will be thrown away.

> Also, a function that calls other logic functions can be written in proper Rust but should be marked #[logic].

Right a `pure` function still can't call `logic` functions.

xldenis:
I'd prefer that, I hope this also alleviates some of @jhjourdan's concern.

PR #130, opened 2021-10-08
Add `Model` instances for references and integers

### LINKED ISSUE #128: Improvements to the ergonomics of `Model`
I wonder if we could improve the ergonomics of `Model` a little, in particular by adding an instance for `& T` which forwards to `T`? 

This could let us transform: 
```rust
#[ensures(result.into() === (@*self).len())]
``` 

into:

```rust
 #[ensures(result.into() === (@self).len())]
```

Which seems a little more readable to me?

shiatsumat:
This PR adds and uses `Model` instances for `&T` and integer types, which closes #128.
I decided not to add an instance for `&mut T`, to keep symmetry between `@*x` and `@^x`.

### DIFF
--- a/creusot/tests/should_succeed/vector/01.rs
+++ b/creusot/tests/should_succeed/vector/01.rs
@@ -30,7 +30,7 @@ impl<T> GhostRecord<T> {
     }
 }
 
-impl<T> Model for MyVec<T> {
+impl<T: ?Sized> Model for MyVec<T> {
     type ModelTy = Seq<T>;
     #[logic_rust]
     #[trusted]
--- a/creusot/tests/should_succeed/vector/01.stdout
+++ b/creusot/tests/should_succeed/vector/01.stdout
@@ -108,6 +108,26 @@ module CreusotContracts_Builtins_Model_Impl0_Model
   function model (self : t) : Model0.modelty = 
     Model0.model self
 end
+module C01_Impl2_Interface
+  type t   
+  use Type
+  use seq.Seq
+  clone export C01_Impl2_Model_Interface with type t = t
+  type modelty  = 
+    Seq.seq t
+  clone export CreusotContracts_Builtins_Model_Model with type self = Type.c01_myvec t, type modelty = modelty,
+  function model = model
+end
+module C01_Impl2
+  type t   
+  use Type
+  use seq.Seq
+  clone export C01_Impl2_Model with type t = t
+  type modelty  = 
+    Seq.seq t
+  clone export CreusotContracts_Builtins_Model_Model with type self = Type.c01_myvec t, type modelty = modelty,
+  function model = model
+end
 module C01_Impl3_Len_Interface
   type t   
   use seq.Seq
@@ -115,7 +135,7 @@ module C01_Impl3_Len_Interface
   use Type
   use mach.int.Int
   use mach.int.UInt64
-  clone CreusotContracts_Builtins_Model_Model as Model1 with type self = Type.c01_myvec t
+  clone C01_Impl2_Interface as Model1 with type t = t
   clone CreusotContracts_Builtins_Model_Impl0_Model_Interface as Model0 with type t = Type.c01_myvec t,
   type Model0.modelty = Model1.modelty, function Model0.model = Model1.model
   val len (self : Type.c01_myvec t) : usize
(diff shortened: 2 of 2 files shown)


shiatsumat:
Even after we replace `Type.creusotcontracts_builtins_int` with `int`, we still get errors at the Why3 step.

The errors seem to come from dereferencing instances on `&T` and `&mut T`.
For example, we get the following error for `inc_some_list.stdout`.
```
File "inc_some_list.stdout", line 195, characters 4-21:
This term has type CreusotContracts_Builtins_Impl1_Model.Model1.Model0.
modelty, but is expected to have type CreusotContracts_Builtins_Impl1_Model.
Model0.modelty
```
The MLCFG code around it is as follows.
```ocaml
module CreusotContracts_Builtins_Impl1_Model_Interface
  type t
  use prelude.Prelude
  clone CreusotContracts_Builtins_Model as Model0 with type self = t
  function model (self : borrowed t) : Model0.modelty
end
module CreusotContracts_Builtins_Impl1_Model
  type t
  use prelude.Prelude
  clone CreusotContracts_Builtins_Model as Model0 with type self = t
  clone CreusotContracts_Builtins_Impl1_Model_Interface as Model1 with type t = t
  function model (self : borrowed t) : Model0.modelty =
    Model1.model self
end
```

xldenis:
Sorry! I accidentally pushed some code to your branch. I undid it though. Good news is i've gotten specialization working in projection types.

shiatsumat:
I added `rustc_diagnostic_item` to `model`s on integer types and modified `lookup_builtin`.
This erases most `model` calls, except for calls on a mutable reference to an integer like `&mut u32`.
I think this is really an improvement and remaining `model` calls on `u32` etc. through `&mut u32` etc. is acceptable.

xldenis:
>  But merging the master (pushed as the commit 17c256b) still doesn't work at the Why3 step.

Werid! I did the merge and loaded `model.stdout` without error, did you also change the test case?

xldenis:
I've identified a fix to the second problem that was brought up by this PR. The issue arises from the fact that implementations _define_ associated types, and thus cannot be refined further. When during cloning we replace a dependency on a _trait_ by a dependency on an _implementation_ of that trait, this issue arises where we attempt to refine the associated type of a trait. 

We clone a function `F`, which was written generically against `Tr`, and during cloning we instantiate `F` with a type that provides a concrete instance of `Tr`, called `I`. The key change here is that we must refine all the symbols of `Tr` using those of `I` but, not attempt to refine symbols of `I` further. This is done by using the _refinable symbols_ of `Tr` during cloning and providing them values of `I`. Additionally, we ensure that implementations do expose types as refinable symbols.

shiatsumat:
Thanks Xavier for many updates! Now it still has a bug with the test `vector/01`, after merging the master (commit 419eadf).
Why3 emits the following error message.
```
File "vector/01.stdout", line 122, characters 34-53:
This term has type C01_Impl3_Len_Interface.Model1.modelty,
but is expected to have type seq.Seq.seq 'xi
```
The `C01_Impl3_Len_Interface` module looks as follows.
```ocaml
module C01_Impl3_Len_Interface
  type t   
  use seq.Seq
  use prelude.Prelude
  use Type
  use mach.int.Int
  use mach.int.UInt64
  clone CreusotContracts_Builtins_Model_Model as Model1 with type self = Type.c01_myvec t
  clone CreusotContracts_Builtins_Model_Impl0_Model_Interface as Model0 with type t = Type.c01_myvec t,
  type Model0.modelty = Model1.modelty, function Model0.model = Model1.model
  val len (self : Type.c01_myvec t) : usize
    ensures { result = Seq.length (Model0.model self) }
    
end
```

shiatsumat:
OK, I identified the bug. I opened Why3-IDE and digged into the `Task` of `3 [postcondition]`, which is as follows.
```
--------------------------- Local Context ---------------------------

function model4 (c01_myvec uint32) : seq uint32

function model3 (c01_myvec uint32) : seq uint32

type modelty2 = seq uint32

function model2 (self:borrowed (c01_myvec uint32)) : seq uint32 =
  model3 (* self)

...

------------------------------- Goal --------------------------------

goal all_zero'vc : length3 (model2 v) = length3 (model4 (^ v))
```
Clearly the problem is that we have two versions for the model function for `MyVec<u32>`, `model4` (the one directly used) and `model3` (the one used by `&mut MyVec<u32>`).

PR #145, opened 2021-10-13
gnome sort

xldenis:
work on gnome sort example for creusot paper.


xldenis:
I can't turn on CI proving because _one_ of the obligations times out with z3 (locally it passes using cvc4).

PR #155, opened 2021-10-16
Export a standard version of Vec

### LINKED ISSUE #135: Confusing issue with sized bounds in traits.
A confusing issue related to `Sized` bounds can appear when writing traits that include logic functions that accept `self`. The immediate solution is to add a `?Sized` bound to the type parameters in the trait instance. Otherwise, it seems we get stuck attempting to resolve `T : Sized`. 

There are two underlying problems:
- Fully understand where issue comes from. 
- Either fix or find a way to emit a proper, helpful error message when this issue arises.

xldenis:
This PR provides a standard implementation of `Vec` as part of the `creusot_contracts` crate. I believe we can then build up a set of wrapper types and traits which are commonly used. 
Perhaps in the future we can provide a mechanism to directly attach specs to foreign items but this provides most of the benefit with very little work.

Along the way to implementing this I had to fix #135 and migrate fully to the binary metadata format.

### DIFF
--- a/creusot-contracts/src/builtins/seq.rs
+++ b/creusot-contracts/src/builtins/seq.rs
@@ -4,7 +4,7 @@ use crate::Int;
 use std::ops::Index;
 
 #[creusot::builtins = "seq.Seq.seq"]
-pub struct Seq<T>(std::marker::PhantomData<T>);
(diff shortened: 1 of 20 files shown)


xldenis:
This required fixing #135 along the way.

PR #193, opened 2021-10-29
Encode pure sub-expressions in logic proof terms

xldenis:
To express pure mathematical functions in `let function` / `let predicat`s in why3 we can use the (undocumented) `pure { .. }` expression. This comes with several key restrictions: we cannot perform recursive calls inside (since they are considered 'program' and not 'pure'). At the same time we may also need to wrap calls to logical functions inside `pure` as well. To handle this we let bind any impure sub-expressions before performing calls to pure functions or operators. We must also avoid being too eager about introducing letbindings or we may break why3's type inference for literals. 

I believe there might be a bug with recursive calls, but bizarrely in the current tests of `06_logic_functions_with_contracts` it seems to work just fine...

cc @jhjourdan.


creusot/tests/should_succeed/syntax/06_logic_function_contracts.stdout
```diff
@@ -70,9 +70,35 @@ module C06LogicFunctionContracts_AllZero_Impl
     variant {Seq.length seq}
     
    = 
-    if Seq.length seq = 0 then
+    if pure {Seq.length seq = 0} then
       true
     else
-      Seq.get seq (Seq.length seq - 1) = 0 && all_zero (SeqExt.subsequence seq 0 (Seq.length seq - 1))
+      pure {Seq.get seq (Seq.length seq - 1) = 0} && all_zero (SeqExt.subsequence seq 0 (Seq.length seq - 1))
```
xldenis, line 76: I don't understand why we don't need to let bind the call to `SeqExt.subsequence` here, it's defined as just a `function`. I think this is another bizarre thing about how `let functions` work (we must be allowed to _call_ logic functions).

Issue #156, opened 2021-10-17
Logical trait heirarchy


xldenis:
In this issue I want to document ideas, proposals and issues identified with providing a standard interface for `std` types in Creusot. 

# Trait Hierarchy

Probably the most fundamental issue we face with exposing a useful set of `std` wrappers is going to be designing the appropriate trait hierarchy. I believe we should have a separate _logical_ hierarchy of traits and then require that each program trait _realizes_ its logical counterpart. This approach is interesting because it also allows us to define operations on purely logical types like `Seq`. 

To start, we could provide the following hierarchy: 

```
 Eq ◄────────── LogicalEq          Index ◄──── LogicalIndex
  │                 │
  │                 │
  │                 │
  ▼                 ▼
Ord ◄────────── LogicalOrd

```

This provides a clear interpretation for operators in Pearlite: they translate to whatever the logical version of the Rust trait would be. 

## Equality

As it stands we currently provide mathematical equality `===` for all types in Pearlite. However, this is not always the correct choice, for some types we want to use an extensional equality which may be provided as an axiom, for example, when dealing with `Seq`. 

Perhaps we should introduce a `LogicalEq` trait which provides one function, with a default implementation in in terms of mathematical equality but which specific types can then override. Then, we would use `===` as sugar for this equality relation. 

- **Laws**: Equality relation: transitivity, reflexivity, symmetry.
- **Realization**: `Eq`, we ask that the result of `Eq` matches with whatever `LogicalEq::eq` says.

## Ordering 

`Ord` is a little trickier because the program trait allows an implementation to provide custom optimized versions of `le`, `lt`, etc.. 
I feel like `LogicalOrd` should only provide `cmp`, but then that leaves a question of how to desugar operations like `<` and `>` in Pearlite.

There are a few possibilities I can see:

1. Also provide `le`, `lt`, etc... in `LogicalOrd` but require them by law to be coherent with `cmp`: they should implement exactly what we expect them to. Perhaps we can even provide a default implementation.
2. Define `le`, etc.. as top level functions that take two values of type `T : LogicalOrd`. This is unsatisfactory as the desugaring for `Ord` is thus fairly ad-hoc but would allow us to guarantee that there is only a single definition of `<` , `<=`, etc... 
[...]

PR #199, opened 2021-11-01
Add `Ord` trait to `creusot_contracts::std`

xldenis:
I still need to come up with some sufficiently challenging test cases to flex this specification.

### DIFF
--- a/creusot-contracts/src/logic/eq.rs
+++ b/creusot-contracts/src/logic/eq.rs
@@ -1,0 +1,70 @@
+use crate as creusot_contracts;
+use crate::logic::Int;
+use creusot_contracts_proc::*;
+
+#[allow(unused)]
+pub trait EqLogic {
+    #[predicate]
+    fn log_eq(self, _: Self) -> bool;
+
+    #[predicate]
+    fn log_ne(self, _: Self) -> bool;
+
+    #[logic]
+    #[ensures(!(a.log_eq(b) === a.log_ne(b)))]
+    fn eq_ne(a: Self, b: Self);
+
+    #[logic]
+    #[ensures(x == x)]
+    fn refl(x: Self);
+
+    #[logic]
+    #[requires(x == y)]
+    #[ensures(y == x)]
+    fn symmetry(x: Self, y: Self);
+
+    #[logic]
+    #[requires(x == y)]
+    #[requires(y == z)]
+    #[ensures(x == z)]
+    fn transitivity(x: Self, y: Self, z: Self);
+}
+
+macro_rules! eq_logic_impl {
+    ($t:ty) => {
+        impl EqLogic for $t {
+            #[predicate]
+            #[creusot::builtins = "=="]
+            fn log_eq(self, _: Self) -> bool {
+                true
+            }
+
+            #[predicate]
+            #[creusot::builtins = "!="]
+            fn log_ne(self, _: Self) -> bool {
+                false
+            }
+
+            #[logic]
+            fn eq_ne(_: Self, _: Self) {}
+
+            #[logic]
+            fn refl(_: Self) {}
+
+            #[logic]
+            fn symmetry(_: Self, _: Self) {}
+
+            #[logic]
+            fn transitivity(_: Self, _: Self, _: Self) {}
+        }
+    };
+}
+
+eq_logic_impl!(Int);
+eq_logic_impl!(usize);
+eq_logic_impl!(u64);
+eq_logic_impl!(u32);
+eq_logic_impl!(isize);
+eq_logic_impl!(i64);
+eq_logic_impl!(i32);
+eq_logic_impl!(bool);
--- a/creusot-contracts/src/logic/ord.rs
+++ b/creusot-contracts/src/logic/ord.rs
@@ -1,0 +1,164 @@
+use crate as creusot_contracts;
+use crate::logic::EqLogic;
+use crate::logic::Int;
+use creusot_contracts_proc::*;
+pub use std::cmp::Ordering;
+
+#[allow(unused)]
+pub trait OrdLogic: EqLogic {
+    #[logic]
+    fn cmp_log(self, _: Self) -> Ordering;
+
+    #[logic]
+    fn le_log(self, o: Self) -> bool {
+        match self.cmp_log(o) {
+            Ordering::Greater => false,
+            _ => true,
+        }
+    }
+
+    #[logic]
+    #[ensures(x.le_log(y) === ! (x.cmp_log(y) === Ordering::Greater))]
+    fn cmp_le_log(x: Self, y: Self);
+
+    #[logic]
+    fn lt_log(self, o: Self) -> bool {
+        match self.cmp_log(o) {
+            Ordering::Less => true,
+            _ => false,
+        }
+    }
+
+    #[logic]
+    #[ensures(x.le_log(y) === (x.cmp_log(y) === Ordering::Less))]
+    fn cmp_lt_log(x: Self, y: Self);
+
+    #[logic]
+    fn ge_log(self, o: Self) -> bool {
+        match self.cmp_log(o) {
+            Ordering::Less => false,
+            _ => true,
+        }
+    }
+
+    #[logic]
+    #[ensures(x.le_log(y) === ! (x.cmp_log(y) === Ordering::Less))]
+    fn cmp_ge_log(x: Self, y: Self);
+
+    #[logic]
+    fn gt_log(self, o: Self) -> bool {
+        match self.cmp_log(o) {
+            Ordering::Greater => true,
+            _ => false,
+        }
+    }
+
+    #[logic]
+    #[ensures(x.le_log(y) === (x.cmp_log(y) === Ordering::Greater))]
+    fn cmp_gt_log(x: Self, y: Self);
+
+    #[logic]
+    #[ensures(x.cmp_log(x) === Ordering::Equal)]
+    fn refl(x: Self);
+
+    #[logic]
+    #[requires(x.cmp_log(y) === o)]
+    #[requires(y.cmp_log(z) === o)]
+    #[ensures(x.cmp_log(z) === o)]
+    fn trans(x: Self, y: Self, z: Self, o: Ordering);
+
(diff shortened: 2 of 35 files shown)


creusot-contracts/src/std/ord.rs
```diff
@@ -0,0 +1,101 @@
+use crate as creusot_contracts;
+use creusot_contracts_proc::*;
+
+use std::cmp::Ordering;
+
+pub trait Ord {
+    #[logic]
+    fn cmp_log(self, _: Self) -> Ordering;
```
xldenis, line 8: We probably will want to separate this out into `OrdLogic` eventually so we can have orderings for purely logical types (`Int`).

creusot/src/translation/ty.rs
```diff
@@ -363,7 +363,8 @@ fn intty_to_ty(
             names.import_prelude_module(PreludeModule::Int64);
             isize_ty()
         }
-        I8 => unimplemented!(),
+        // for now
+        I8 => MlT::Tuple(vec![]),
```
xldenis, line 367: Fix!

creusot/tests/should_succeed/ord_trait.rs
```diff
@@ -0,0 +1,18 @@
...
+
+#[ensures(result === true)]
+fn x<T: Ord>(x: &T) -> bool {
+    proof_assert! { {Ord::refl(*x); true} };
+    x.le(x)
+}
+
+#[ensures(result === y.le_log(*x))]
+fn gt_or_le<T: Ord>(x: &T, y: &T) -> bool {
+    proof_assert! { {Ord::antisym1(*x, *y); true} };
```
xldenis, line 14: Definitely need a `#[law]` which autoloads.

creusot-contracts/src/std/ord.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    #[logic]
+    #[requires(x.cmp_log(y) === Ordering::Greater)]
+    #[ensures(y.cmp_log(x) === Ordering::Less)]
+    fn antisym2(x: Self, y: Self);
+
+    #[logic]
+    #[requires(x.cmp_log(y) === Ordering::Equal)]
+    #[ensures(y.cmp_log(x) === Ordering::Equal)]
+    fn symmetry(x: Self, y: Self);
```
jhjourdan, line 100: Same thing here: this should not be called `symmetry`. The whole point of order relations is that they are *not* symetric.

creusot-contracts/src/std/ord.rs
```diff
@@ -0,0 +1,101 @@
+use crate as creusot_contracts;
+use creusot_contracts_proc::*;
+
+use std::cmp::Ordering;
+
+pub trait Ord {
```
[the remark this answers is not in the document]
xldenis: Yes, we should definitely extend this to be a subtrait of `Eq`.

creusot-contracts/src/std/ord.rs
```diff
@@ -0,0 +1,101 @@
...
+    #[logic]
+    fn cmp_log(self, _: Self) -> Ordering;
+
+    #[ensures(result === self.cmp_log(*o))]
+    fn cmp(&self, o: &Self) -> Ordering;
+
+    #[logic]
+    fn le_log(self, o: Self) -> bool {
```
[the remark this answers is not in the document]
xldenis: I discussed this possibility in #156, but I think you're right.

creusot-contracts/src/std/ord.rs
```diff
@@ -0,0 +1,101 @@
...
+    #[logic]
+    #[requires(x.cmp_log(y) === o)]
+    #[requires(y.cmp_log(x) === o)]
+    #[ensures(x.cmp_log(z) === o)]
+    fn trans(x: Self, y: Self, z: Self, o: Ordering);
+
+    #[logic]
+    #[requires(x.cmp_log(y) === Ordering::Less)]
+    #[ensures(y.cmp_log(x) === Ordering::Greater)]
+    fn antisym1(x: Self, y: Self);
```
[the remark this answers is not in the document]
xldenis: What would you call this then?

creusot-contracts/src/logic/eq.rs
```diff
@@ -0,0 +1,22 @@
+use crate as creusot_contracts;
+use creusot_contracts_proc::*;
+
+pub trait EqLogic {
```
xldenis, line 4: Do we require `log_eq` to be extensional?

xldenis: If so, do we need any other laws?

creusot-contracts/src/logic/ord.rs
```diff
@@ -0,0 +1,66 @@
...
+
+#[logic]
+pub fn ge_log<T: OrdLogic + ?Sized>(a: T, o: T) -> bool {
+    match a.cmp_log(o) {
+        Ordering::Less => false,
+        _ => true,
+    }
+}
+#[logic]
+pub fn gt_log<T: OrdLogic + ?Sized>(a: T, o: T) -> bool {
```
xldenis, line 23: the big disadvantage of moving this to a top-level function is that it's no longer possible to override it for integer types so that we emit `<` instead.

xldenis: (this is a **very** minor issue though.

xldenis: I wish why3 had the equivalent of Coq's notations so at least we could pretty-print in proof tasks.

creusot-contracts/src/logic/eq.rs
jhjourdan: Do you mean by "extensional" that it should be equivalent to `===`? I don't think this is a great idea. For some types, we really need an equality which is not `===`.

If you want, we can also have a trait for types where `===` coincides with `==`.

creusot-contracts/src/logic/ord.rs
jhjourdan: That's indeed a very important point, since Why3's `<` is translated to the provers' native comparison operator over integers, and I guess this is crucial for good performance of provers. So I guess this forces us to use default implementations, and require a lemma that `<` is indeed equivalent to `cmp`.

BTW, we should be very cautious about default specifications of default implementations, as it can cause unsoundness. Indeed, if the specification of a default implementation of a lemma/program function depends on a logic function/predicate which has a default implementation, then we should ask the user to reprove the specification of the lemma/program funciton as soon as either the lemma/program function OR logic function/predicate is given a non-default implementation.

In this particular case, let's call `cmp_gt_log` the lemma stating that `cmp` and `gt_log` are equivalent. This means that we should make sure that we generate VCs for `cmp_gt_log` as soon as we give a non-default implementation for `cmp_gt_log` (of course) OR `gt_log` (which is less obvious).

xldenis: Agreed, if we provide default logical methods we need to ensure coherence between definitions.

xldenis: Fwiw, I think we should *always* generate vcs, but if you don't change any of the definitions involved the proof should be trivial (just what was proven for the trait).

jhjourdan: That's a good first implementation, yes.

But imagine a default implementation of program function. Its body can be quite complex, so a large number of VCs could be generated from that, and perhaps the proof is not even automatic. If we don't change anything, then we'd like to avoid re-doing the proof (that's the whole point of default implementations !).

The problem here is that we should make sure the spec does not depend on a reimplemented default implementation, transitively (if it depends on a default implementation which depends on a method which had a defaulf implementation which has been overridden, then the proof is out-of-date). So I guess this criterion is rather subtle to implement.

xldenis: If we don't change anything we literally clone the default substitution so we inherit the proof as well!

xldenis: Ah i see your concern, yes we could sneak in unsoundness during this clone process if another default implementation was also changed.

creusot-contracts/src/logic/eq.rs
xldenis: Yea that's what I meant but it's probably better to separate the two like `PartialOrd` and `Ord` are

creusot-contracts/src/logic/ord.rs
xldenis: > The problem here is that we should make sure the spec does not depend on a reimplemented default implementation, transitively (if it depends on a default implementation which depends on a method which had a defaulf implementation which has been overridden, then the proof is out-of-date). So I guess this criterion is rather subtle to implement.

I believe we can implement this by asking if the proof (transitively) depends on the implementation. If it does we somehow force a new proof.

creusot-contracts/src/logic/ord.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    #[logic]
+    #[requires(x.cmp_log(y) === Ordering::Greater)]
+    #[ensures(y.cmp_log(x) === Ordering::Less)]
+    fn antisym2(x: Self, y: Self);
+
+    #[logic]
+    #[requires(x.cmp_log(y) === Ordering::Equal)]
+    #[ensures(y.cmp_log(x) === Ordering::Equal)]
+    fn symmetry(x: Self, y: Self);
```
jhjourdan, line 82: Instead of that, we should perhaps have a lemma relating `cmp_log` with `eq_log`, and rely on symmetry of equality instead.

jhjourdan:
Another comment: we should also change examples using `Ord` such as `gnome_sort`, which still use its own custom version.

xldenis:
> Another comment: we should also change examples using Ord such as gnome_sort, which still use its own custom version.

Agreed. However, keep in mind that while we have access to the `< / <=` syntax in Pearlite, we will not have access to it for programs.

Fixing that will require making a custom version of the `std` crate.

jhjourdan:
> Fixing that will require making a custom version of the std crate.

Or having a feature to add in a subtrait specifications to functions of the super trait. That's probably less heavyweight than rewriting core.

PR #202, opened 2021-11-02
Make knapsack test fully automatic (if provers are givzn enough time).

jhjourdan:


xldenis:
you can write `// WHY3PROVE Z3` (or CVC4) at the top of the test files to run the provers during CI. It will perform a `split_vc` and then run the prover(s) on it.

jhjourdan:
Yeah, I know. One of the VCs for knapsack_full takes ~42s to prove with Why3, so I am not sure we want to add this to CI.

In any case, what's the policy about activating `WHY3PROVE` in CI? Most of our tests don't have this, even if they are fully automatic.

xldenis:
I think we should make it the default as much as possible. We should also go back over older tests and enable it where appropriate, but tests don't all have proof obligations, some are merely testing the syntactic validity of the translation.

jhjourdan:
So... the knapsack.rs test fail in CI because apparently GitHubs machines are less efficient than mine...

BTW, don't you think that we should use `why3 replay` rather then `why3 prove`, so that we can have why3 sessions with multiple different provers?

xldenis:
> BTW, don't you think that we should use why3 replay rather then why3 prove, so that we can have why3 sessions with multiple different provers?

Perhaps, but that would require committing why3 sessions (maybe that's not a bad idea?).

jhjourdan:
> Perhaps, but that would require committing why3 sessions (maybe that's not a bad idea?).

That's what Why3 is doing for its own CI.

xldenis:
Yea, you're right we should transition to the same model for creusot.

PR #215, opened 2021-11-09
Knight's Tour

xldenis:
Proves the (complete) safety of the Knight's tour, improving on Prusti's version by not axiomitizing the board access / setting and other auxiliary definitions.

There's currently one remaining goal which is driving me mad: proving that `i * i` doesn't overflow for unsigned 64 bit integers less than 1000!


xldenis:
Proving full functional correctness actually shouldn't be _too_ difficult. The main challenge is defining a logical function which extracts the hamiltonian path from the board.

PR #225, opened 2021-11-11
Trait Laws

xldenis:
Adds laws which auto-load whenever a trait or impl item is used.

TODO: Load the law when an impl item is used.

Unresolved Questions
-------------------- 

What are we supposed to do when the law has additional generic parameters?

I think for now we should just forbid laws from having additional generics, as we would not be able to guess what values to give them at clone-time.


jhjourdan:
> What are we supposed to do when the law has additional generic parameters?

> I think for now we should just forbid laws from having additional generics, as we would not be able to guess what values to give them at clone-time.

Agreed.

PR #228, opened 2021-11-12
Cleanup filter_positive test.

jhjourdan:


jhjourdan:
Also added support for if-then expressions in Pearlite.

PR #229, opened 2021-11-12
Macro Passthrough

xldenis:
Adds a set of dummy macros which are on by default when compiling with normal `rustc`. This makes it possible to actually compile Creusot programs and execute them.

This still requires being a little careful around imports of things from `creusot_contracts::logic` but maybe we can think up a more coherent story. 

Additionally, when compiling with rustc it is necessary to manually set the `expr_stmt_attributes` and `proc_macro_hygiene` flags if any loop invariants are used, nothing I can do about that sadly. 

cc @claudemarche @sarsko I know both of you have wanted this feature.

### DIFF
--- a/creusot-contracts/src/lib.rs
+++ b/creusot-contracts/src/lib.rs
@@ -1,7 +1,95 @@
 #![cfg_attr(feature = "contracts", feature(unsized_fn_params))]
 #![cfg_attr(feature = "typechecker", feature(rustc_private), feature(box_patterns, box_syntax))]
 
-pub use creusot_contracts_proc::*;
+#[cfg(feature = "contracts")]
+mod macros {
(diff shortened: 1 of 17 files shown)


creusot-contracts/src/lib.rs
```diff
@@ -1,19 +1,124 @@
...
+}
+
+#[cfg(not(feature = "contracts"))]
+pub mod logic {
+    pub struct Ghost<T>(std::marker::PhantomData<T>)
+    where
+        T: ?Sized;
+
+    impl<T> Ghost<T> {
+        pub fn record(_: &T) -> Ghost<T> {
```
xldenis, line 115: Since `Ghost` appears in Rust programs, I added a dummy implementation of the type which doesn't do anything.

PR #232, opened 2021-11-14
Add selection_sort_i32

sarsko:
Based on the one from Prusti/Rosetta.

Split it up into functions. There is a case to be made for reverting to nested while-loops. I feel like this is more readable, whereas nested loops yield an overall shorter program.


creusot/tests/should_succeed/selection_sort_i32.rs
```diff
@@ -0,0 +1,135 @@
...
+    }
+}
+
+impl<T> Ghost<T> {
+    #[trusted]
+    #[ensures(@result === *a)]
+    fn record(a: &T) -> Ghost<T> {
+        panic!()
+    }
+}
```
xldenis, line 26: We now have `Ghost` in the standard library, so you can delete this bit.

creusot/tests/should_succeed/selection_sort_i32.rs
```diff
@@ -0,0 +1,135 @@
...
+#[requires(selection_sort_invariant(@v, @i))]
+#[requires(forall<j: Int> @i <= j && j < (@v).len() ==> (@v)[@min] <= (@v)[j])] // min smaller
+#[ensures((@v)[@i] === (@^v)[@min] && (@v)[@min] === (@^v)[@i])] // i and min swapped
+#[ensures(forall<j: Int> 0 <= j && j < (@v).len() && j != @i && j != @min ==> (@v)[j] === (@^v)[j])] // rest are unchanged
+#[ensures((@^v).permutation_of(@v))]
+#[ensures(selection_sort_invariant(@^v, @i+1))]
+#[ensures(sorted_range(@^v, 0, @i+1))]
+fn swap(v: &mut Vec<i32>, i: usize, min: usize) {
+    v.swap(i, min);
+}
```
xldenis, line 118: I don't understand why you need this at all?

creusot/tests/should_succeed/selection_sort_i32.rs
```diff
@@ -0,0 +1,135 @@
...
+    all_larger(s, i) && all_smaller(s, i)
+}
+
+#[predicate]
+fn min_larger(s: Seq<i32>, u: Int, idx: Int) -> bool {
+    pearlite! {forall<j: Int> 0 <= j && j < u ==> s[j] <= s[idx] }
+}
+
+#[requires(min_larger(@*v, @i, @i))]
+#[requires(sorted_range(@v, 0, @i))]
```
xldenis, line 84: I don't think we care that the vector is sorted?

creusot/tests/should_succeed/selection_sort_i32.rs
```diff
@@ -0,0 +1,135 @@
...
+#[requires(selection_sort_invariant(@v, @i))]
+#[requires((@v).len() <= @0xffff_ffff_ffff_ffffusize)]
+#[requires(0 <= @i && @i < (@v).len())]
+#[ensures(min_larger(@v, @i, @result))]
+#[ensures(forall<j: Int> @i <= j && j < (@v).len() ==> (@v)[@result] <= (@v)[j])] // min smaller
```
xldenis, line 89: As your comment notices, I think that this could be generalized? We're asking that `result` is a the minimum of the subsequence from `i, v.len()`

creusot/tests/should_succeed/selection_sort_i32.rs
```diff
@@ -0,0 +1,135 @@
...
+    }
+}
+
+#[predicate]
+fn larger_than(s: Seq<i32>, e: i32, u: Int) -> bool {
+    pearlite! {
+        forall<i: Int> 0 <= i && i < u && i < s.len() ==>
+            e >= s[i]
+    }
+}
```
xldenis, line 56: Can we define `larger_than` as not `smaller_than`?

xldenis, line 56: I wonder if these couldn't be combined? It seems like they are actually stating something about the partition of the sequence around `e`?

creusot/tests/should_succeed/selection_sort_i32.rs
```diff
@@ -0,0 +1,135 @@
...
+#[ensures((@^v).permutation_of(@v))]
+#[ensures(selection_sort_invariant(@^v, @i+1))]
+#[ensures(sorted_range(@^v, 0, @i+1))]
+fn swap(v: &mut Vec<i32>, i: usize, min: usize) {
+    v.swap(i, min);
+}
+
+#[ensures(sorted(@^v))]
+#[ensures((@^v).permutation_of(@v))]
+fn selection_sort(v: &mut Vec<i32>) {
```
xldenis, line 122: What happens when we replace `i32` by `T : Ord`? It should be equivalent, but I'm curious to know what happens.

creusot/tests/should_succeed/selection_sort_i32.rs
```diff
@@ -0,0 +1,135 @@
...
+    #[invariant(j_bound, @i <= @j && @j <= (@v).len())]
+    #[invariant(min_bound, @i <= @min && @min < (@v).len() && @min <= @j)]
+    while j < v.len() {
+        if v[j] < v[min] {
+            min = j;
+        }
+        j += 1;
+    }
+    return min;
+}
```
xldenis, line 104: I think it might be better to combine the two functions, it's a more traditional presentation of selection sort that way

xldenis (commented):
Thanks for the submission Sarek! I took a quick look over it and made some initial comments. I haven't taken a look at the proofs to know if all my suggestions work / make sense so let me know what you think.

creusot/tests/should_succeed/selection_sort_i32.rs
sarsko: Abstracted it away into `min_smaller`

creusot/tests/should_succeed/selection_sort_i32.rs
sarsko: Perhaps. They are used to state that everything in the sorted part is smaller than or equal to everything in the larger part, and vice versa. Probably possible to merge, but I'm not sure that it will become more readable/prettier/faster.

sarsko:
Thanks for the feedback. I implemented the suggested changes except changing the inners of `selection_sort_invariant`. Haven't looked at making it generic. New version is a tad slower than the previous version(~5 secs -> ~6.5 secs).  Haven't found a way to remove the assertions without making it even slower.

xldenis:
Great, I'll take a further look later tonight.

PR #234, opened 2021-11-15
Selection sort generic

sarsko:
A generic mirror of the i32-version. There is a case to be made for keeping the monomorphic one, as it proves over twice as fast, and is enables an apples-to-apples comparison between Creusot/Prusti.


xldenis:
I'm merging and editing this further by hand.

PR #249, opened 2021-12-06
start prototyping an actual error struct

xldenis:


xldenis:
Im going to go ahead and merge this and attempt to progressively move things over to this model.

PR #254, opened 2022-01-10
Extern Specs

xldenis:
Need to clean things up a lot, but lays the groundwork for extern specs.

### DIFF
--- a/creusot-contracts-dummy/src/lib.rs
+++ b/creusot-contracts-dummy/src/lib.rs
@@ -51,3 +51,8 @@ pub fn law(_: TS1, _: TS1) -> TS1 {
 pub fn trusted(_: TS1, tokens: TS1) -> TS1 {
     tokens
 }
+
+#[proc_macro]
+pub fn extern_spec(_: TS1) -> TS1 {
+    TS1::new()
+}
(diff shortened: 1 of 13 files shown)


extern_test.rs
```diff
@@ -0,0 +1,13 @@
+extern crate creusot_contracts;
+use creusot_contracts::*;
+
+#[creusot::no_translate]
+#[creusot::extern_spec]
```
jhjourdan, line 5: Would it be possible to make `extern_spec` imply `no_translate` automatically?

extern_test.rs
```diff
@@ -0,0 +1,13 @@
...
+#[requires(true === true)]
+fn extern_xxxx<T>() -> Vec<T> {
```
jhjourdan, line 7: I don't understand the syntax here. Is it supposed to be the spec of function `xxxx`? Why does this function have a body?

extern_test.rs
xldenis: Yes, one of the final steps will be to wrap this all up in a macro. What you're seeing here is the desugared syntax

extern_test.rs
xldenis: `xxxx` represents a random uuid. We attach the spec to `extern_xxxx` and inside have a call to the function we really want to specify. 

This structure is partially caused by the fact we can only resolve names after/during type checking. To even identify `Vec::new` we must run the type checker.

PR #272, opened 2022-02-03
namespace hacking

xldenis:


xldenis:
I really wish i had gotten this working.. the idea was to use proc macro spans to place logic and program functions in separate namespaces. It _almost_ worked but there seemed to be a bug in rustc's handling of span contexts (forgot what though).

Issue #283, opened 2022-02-21
'anomaly: Failure("stackify is not compiled")' error message


safinaskar:
Hi. After several try-and-error steps I was able to create Dockerfile, which installs creusot. Here it is:
```dockerfile
FROM debian:bullseye
ENV LC_ALL C.UTF-8
RUN apt-get update && apt-get install -y apt-utils gcc curl less nano mc whiptail locales man-db wget procps sudo apt-file pkg-config

# User
RUN useradd --create-home --uid 1000 --shell /bin/bash user
RUN echo 'user ALL=(ALL) NOPASSWD: ALL' > /etc/sudoers.d/a
USER user
WORKDIR /home/user

# Rust
RUN curl --proto '=https' --tlsv1.2 https://sh.rustup.rs -sSf | bash -s -- -y --default-toolchain none
RUN . "$HOME/.cargo/env"; rustup toolchain install nightly --allow-downgrade

USER root
RUN apt-get update && apt-get install -y git
USER user

RUN . "$HOME/.cargo/env"; mkdir ~/opt && cd ~/opt && git clone https://github.com/xldenis/creusot && creusot/build

USER root
RUN apt-get update && apt-get install -y make bubblewrap
USER user

RUN mkdir ~/bin && wget https://github.com/ocaml/opam/releases/download/2.1.2/opam-2.1.2-x86_64-linux -O ~/bin/opam && chmod +x ~/bin/opam
ENV PATH /home/user/bin:/usr/local/bin:/usr/bin:/bin
RUN opam init --auto-setup --disable-sandboxing

RUN opam pin --confirm-level=unsafe-yes add why3 https://gitlab.inria.fr/why3/why3.git
RUN opam install --confirm-level=unsafe-yes why3 why3-ide

USER root
RUN apt-get update && apt-get install -y z3
USER user

RUN eval $(opam env); why3 config detect

RUN echo 'fn main(){}' > ~/a.rs
RUN . "$HOME/.cargo/env"; eval $(opam env); ~/opt/creusot/mlcfg ~/a.rs > ~/a.mlcfg
RUN . "$HOME/.cargo/env"; eval $(opam env); ~/opt/creusot/prove -P z3 ~/a.mlcfg # This line fails
```
(By the way it would be good if you will provide similar Dockerfile or at least automatic install script.)

Okey, so I created this Dockerfile and it doesn't work. The very last line fails with message `anomaly: Failure("stackify is not compiled")`

xldenis:
[...]

PR #285, opened 2022-02-21
trivial vector test

xldenis:
This adds a test case with the simple example from #283, using vectors (not slices).


xldenis:
asdfasdfasf wtf why are... some of the tests failing, consistently? but not locally?

PR #289, opened 2022-02-21
trivial vector test

xldenis:
- Ensure all local traits are translated
- Add read-write vector example
- going crazy
- make me fucking crazy


xldenis:
>  DefId(20:350 ~ creusot_contracts[5fad]::logic::ord::OrdLogic::cmp_log) ([Self, Self]; c_variadic: false)->fn(i32, i32) -> bool {<i32 as creusot_contracts::OrdLogic>::lt_log}

This is insane... somehow the `cmp_log` is seen as **returning** `lt_log` from the outside... but not when we compile `creusot_contracts`... it works just fine there??

PR #291, opened 2022-02-22
Add a module for mappings in logic

claudemarche:
A first draft of a type of mappings in logic, with a simple test


creusot-contracts/src/logic/mapping.rs
```diff
@@ -0,0 +1,36 @@
...
+    #[logic]
+    #[creusot::builtins = "map.Const.const"]
+    pub fn cst(_: B) -> Self {
+        std::process::abort()
+    }
+
+    #[trusted]
+    #[logic]
+    #[creusot::builtins = "map.MapExt.(==)"]
+    pub fn eq(self, _: Self) -> bool {
```
xldenis, line 32: You should implement `LogicEq` instead, that way we can use the symbol `==` in proofs.

claudemarche: Agreed!

creusot/tests/should_succeed/mapping_test.rs
```diff
@@ -20,16 +18,16 @@ impl Model for T {
...
         std::process::abort()
     }
-
 }
 
 #[requires( 0 <= @((*t).a) )] // otherwise its wrong !
 #[requires( @((*t).a) < 1000 )] // to prevent overflow
-#[ensures( (@^t).eq((@*t).set(@((*t).a),1)) )]
-fn incr(t:&mut T) {
+#[ensures( @^t === (@*t).set(@((*t).a),1) )]
```
xldenis, line 25: adding `LogicalEq` means you can use `==`. this is key since your equality needs to be extensional unlike `===`.

creusot/tests/should_succeed/mapping_test.rs
```diff
@@ -20,16 +18,16 @@ impl Model for T {
...
+fn incr(t: &mut T) {
     let old_t = Ghost::record(t);
     (*t).a += 1;
-    proof_assert!( (@^t).eq((@@old_t).set(@((@old_t).a),1)) );
+    // extensional equality of mappings
+    proof_assert!( @^t == (@@old_t).set(@((@old_t).a),1) );
```
xldenis, line 30: :+1:

creusot/tests/should_succeed/mapping_test.rs
claudemarche: I'm not sure what is your point. I really want to use `===` in the post-condition, since it will allow callers to make use of this fact using logical equality reasoning (e.g., perform rewrite steps)

creusot/tests/should_succeed/mapping_test.rs
claudemarche: On the other hand, proving the post-condition is not trivial, and the intermediate step through `==` in the assertion helps. Note that the fact `f == g => f === g` is already stated (and proved!) in Why3.

claudemarche:
Apparently I forgot to update `mapping_test.stdout`.

creusot-contracts/src/logic/mapping.rs
```diff
@@ -25,12 +26,38 @@ impl<A,B> Mapping<A,B> {
...
     }
 
+    #[logic]
+    fn refl(_: Self) {
+        ()
+    }
+
+    #[logic]
+    fn symmetry(_: Self, _: Self) {
+        ()
```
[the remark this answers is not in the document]
claudemarche: OK, will do it. By the way, I believe there are `()` like this elsewhere in the library, I copied that from somewhere, probably sequences

creusot/tests/should_succeed/mapping_test.rs
xldenis: Ah, interesting, I would have figured it would be better to consistently use `==`

xldenis: Doesn't the definition of `==` have something like `a == b -> a = b`?

jhjourdan: Note that I would not expect `===` and `==` to be equivalent for mappings.

`===` should be logical, Leibniz equality, while `==` should be pointwise `==` equality. If `===` and `==` are not equivalent on the underlying type, then `===` and `==` should not be equivalent (even if mappings are extensional).

This is what I would expect from the notations. Question is: is this really what we want? In particular, this means that we do a different distinction between `===` and `==` than Why3 is doing...

xldenis: We do have extensionality for `==` on mappings:

```
lemma extensionality:
    forall m1 m2: 'a -> 'b. m1 == m2 -> m1 = m2
```

Where `==` is defined pointwise using `=`. 

> This is what I would expect from the notations. Question is: is this really what we want? In particular, this means that we do a different distinction between === and == than Why3 is doing...

That is the real question, and we actually _can_ do something different (perhaps even *better*) as we can rely on traits. Why3 can't / doesn't parameterize definitions of `==` by the `==` on the underlying type, but we can.

xldenis: But I had the idea of defining `===` as a default for `LogicalEq` and using specialization to allow overriding it with more specific notions of equality but is that appropriate?

xldenis:
```
impl<A, B: LogicalEq> LogicalEq for Map<A, B> { 
  #[predicate]
  fn eq_log(self, o: Self) -> bool {
    forall<a : A> self.get(a) == o.get(a)
  }
}
```

jhjourdan: Yeah, that's what I would expect. But then this does not meet the notion in Why3, and perhaps we will not leverage the lemmas that are proved in Why3's standard library about its `==`...

claudemarche: In short, `===` (that is `=` in Why3) is built-in in provers, whereas `==` is not, that's why I'd like to use `===` whenever possible.

And there is the fact that extensionality lemmas like to one for mappings above is *not* applied easily by provers (where as the converse is a triviality)

claudemarche:
Could we merge this PR now? @xldenis

creusot/tests/should_succeed/mapping_test.rs
claudemarche: > Note that I would not expect `===` and `==` to be equivalent for mappings.
> 
> `===` should be logical, Leibniz equality, while `==` should be pointwise `==` equality. If `===` and `==` are not equivalent on the underlying type, then `===` and `==` should not be equivalent (even if mappings are extensional).
> 
> This is what I would expect from the notations. Question is: is this really what we want? In particular, this means that we do a different distinction between `===` and `==` than Why3 is doing...

In the Why3 logic, and its theory of mappings, `===` and `==` are proved equivalent. The situation is different from Coq (as it is different regarding excluded middle, inhabitation of types, etc.)

xldenis: > leverage the lemmas that are proved in Why3's standard library about its ==...

There's only one and given the comment in the library, i'm not even sure it is proven in the first place.

> In short, === (that is = in Why3) is built-in in provers, whereas == is not, that's why I'd like to use === whenever possible.
> And there is the fact that extensionality lemmas like to one for mappings above is not applied easily by provers (where as the converse is a triviality)

This makes sense, perhaps the 'idiom' would be to use `==` inside a function and `===` in its external spec? Or could there be ways to better hint to provers to more systematically attempt extensionality?

xldenis:
Sure, I think we should return to the question of `EqLogic` though. I'll mark it on the agenda for next week.

creusot/tests/should_succeed/mapping_test.rs
jhjourdan: > In the Why3 logic, and its theory of mappings, === and == are proved equivalent.

That is not true if the equality on which `==` is defined is not the logical equality over elements.

claudemarche: > Or could there be ways to better hint to provers to more systematically attempt extensionality?

This is a long-standing question, there are tickets about that in Why3...

xldenis: > That is not true if the equality on which == is defined is not the logical equality over elements.

Should we make extensionality a required law of `LogicalEq` ?

claudemarche: > That is not true if the equality on which `==` is defined is not the logical equality over elements.

Indeed, but `==` on mappings in why3 *is* the point-wise logical equality of elements!

claudemarche: > Should we make extensionality a required law of `LogicalEq` ?

What does it mean ??

claudemarche:
> Sure, I think we should return to the question of `EqLogic` though. I'll mark it on the agenda for next week.

Next week I'm on vacation, next next week please

creusot/tests/should_succeed/mapping_test.rs
jhjourdan: That would mean making `===` and `==` equivalent for every type in Creusot.

I am not so sure this is what we want. This would really defeat the point of `LogicalEq`: this trait is here to define an alternative notion of equality if needed.

xldenis: > This would really defeat the point of LogicalEq: this trait is here to define an alternative notion of equality if needed.

Not necessarily defeat, as it allows us to sidestep the intensionality of `=`, for example: on `seq`, you often don't want `=` as you find yourself incapable to prove simple properties when you build / modify the sequence. That's why why3 comes with a definition of `==` which is extensional.

PR #295, opened 2022-02-27
Slices

xldenis:
Adds _very minimal_ support for slices: in particular we can perform the 3 core operations: immutable indexing, mutable indexing and getting the length. We can then expand this as needed.


creusot-contracts/src/logic/model.rs
```diff
@@ -21,3 +21,14 @@ impl<T: Model> Model for &mut T {
         (*self).model()
     }
 }
+
+impl<'a, T> Model for &'a [T] {
```
xldenis, line 25: I would like to define this directly on `[T]` but it introduces a bunch of issues related to `Sized`...

PR #348, opened 2022-05-16
Get rid of the === operator.

jhjourdan:
We agreed that `==` should refer to mathematical equality, so there is no need for the === syntax.


jhjourdan:
This is a rather large change, so I prefer not merging myself before @xldenis approves explicitely.

xldenis (approved):
lgtm

PR #356, opened 2022-05-18
[EXPERIMENT] [DO NOT MERGE] Tool Lib

xldenis:
As part of [Project SMIR](https://github.com/rust-lang/project-stable-mir), I am experimenting with a `tool_lib` which would provide a minimal and stable interface to Rust for compiler tools to use. 

In this PR, I attempt to port Creusot to this too lib and identify which additional apis are required. 
Afterwards, the challenge will be identifying which apis should _not_ be used and having `tool-lib` provide alternatives or stable wrappers.


xldenis:
out of date

Issue #210, opened 2021-11-03
Store why3 sessions of tests for replaying in CI


xldenis:
Rather than using `why3 prove` during CI we should use `why3 replay`. This will allow us to run more subtle proofs than `split_vc` + prover in CI. Additionally, it would allow having specific goals solved by a specific prover rather than forcing it to be uniform.

PR #365, opened 2022-05-23
Install Alt-Ergo in CI

xldenis:


xldenis:
Turns out if we want to add support for checking `binary_search` then we need more than just alt-ergo. `len_logic` is not solveable after a split but is trivially before one... We really need to fix #210

PR #371, opened 2022-05-23
Fix #127

### LINKED ISSUE #127: Make it possible to hide the body of public logic function
@xldenis  proposed to have an `#[opaque]` attribute. That's a possibility, indeed.

Then the default is that the body is public when the function is public, which is good, after all.

When the body of a logic function uses private symbols, what should we do? We should either emit an error or make the function opaque automatically. I guess the less surprising behavior is to emit an error.

(I guess Creusot is already doing something when the spec of a function (logic or program) mentions a private symbol (type, function, trait, ....), isn't it?)

### LINKED ISSUE #222: Unsound interaction between default logical methods and trait implementations
By combining default logical methods and implementation refinement it is possible to introduce unsoundness. Quoting:

> That's indeed a very important point, since Why3's < is translated to the provers' native comparison operator over integers, and I guess this is crucial for good performance of provers. So I guess this forces us to use default implementations, and require a lemma that < is indeed equivalent to cmp.
>
> BTW, we should be very cautious about default specifications of default implementations, as it can cause unsoundness. Indeed, if the specification of a default implementation of a lemma/program function depends on a logic function/predicate which has a default implementation, then we should ask the user to reprove the specification of the lemma/program funciton as soon as either the lemma/program function OR logic function/predicate is given a non-default implementation.
>
> In this particular case, let's call cmp_gt_log the lemma stating that cmp and gt_log are equivalent. This means that we should make sure that we generate VCs for cmp_gt_log as soon as we give a non-default implementation for cmp_gt_log (of course) OR gt_log (which is less obvious).

We should ensure that providing an alternate implementation for a default invalidates all dependent proof obligations. I'm not yet sure what the proper way of doing this should be.

xldenis:
Adds a `#[creusot::decl::opaque]` annotation which prevents using the body of a definition in dependencies.

### DIFF
--- a/creusot/src/clone_map.rs
+++ b/creusot/src/clone_map.rs
@@ -600,7 +600,8 @@ fn cloneable_name(tcx: TyCtxt, def_id: DefId, interface: bool) -> QName {
     // TODO: Refactor.
     match util::item_type(tcx, def_id) {
         Logic | Predicate | Impl => {
-            if interface {
+            let is_opaque = util::is_opaque(tcx, def_id);
+            if interface || is_opaque {
                 // TODO: this should directly be a function...
                 QName { module: Vec::new(), name: interface::interface_name(tcx, def_id) }
             } else {
--- a/creusot/src/util.rs
+++ b/creusot/src/util.rs
@@ -73,6 +73,10 @@ pub(crate) fn is_trusted(tcx: TyCtxt, def_id: DefId) -> bool {
     get_attr(tcx.get_attrs_unchecked(def_id), &["creusot", "decl", "trusted"]).is_some()
 }
 
+pub(crate) fn is_opaque(tcx: TyCtxt, def_id: DefId) -> bool {
+    get_attr(tcx.get_attrs_unchecked(def_id), &["creusot", "decl", "opaque"]).is_some()
+}
+
 pub(crate) fn is_law(tcx: TyCtxt, def_id: DefId) -> bool {
     get_attr(tcx.get_attrs_unchecked(def_id), &["creusot", "decl", "law"]).is_some()
 }
--- a/creusot/tests/should_succeed/opaque.rs
+++ b/creusot/tests/should_succeed/opaque.rs
@@ -1,0 +1,16 @@
+extern crate creusot_contracts;
+use creusot_contracts::*;
+
+#[logic]
+#[creusot::decl::opaque]
+#[ensures(result == 2)]
+fn two() -> Int {
+    pearlite! {
+     1 + 1 - 1 + 1 - 1 + 1 - 1 + 1
+    }
+}
+
+#[ensures(@result == two())]
+fn x() -> u64 {
+    2
+}
--- a/creusot/tests/should_succeed/opaque.stdout
+++ b/creusot/tests/should_succeed/opaque.stdout
@@ -1,0 +1,63 @@
+module Type
+  use Ref
+  use mach.int.Int
+  use prelude.Int8
+  use prelude.Int16
+  use mach.int.Int32
+  use mach.int.Int64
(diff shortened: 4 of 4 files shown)


jhjourdan:
What do you mean by "dependencies"? Ideally, this should have the same semantics as the `pub` keyword...

xldenis:
Good point, we need to do a check similar to with `pub` for this to be actually useful. I propose that we only allow it to be visible in child modules, as that is by far simpler to implement than the `pub(super)`, `pub(crate)` etc.. variants.

jhjourdan:
Actually, I have an idea: we could simply decide that the body of a logical definition is private as soon as it mentions a private "dummy" symbol. It is then easy to make a logical definition public but opaque, by making it body mention a private symbol (this could be automatized with a macro).

This way, we do not have to reinvent a new system of opacity in parallel to the default one of Rust. And by carefully managing the opacity of the "dummy" symbol, we can give the desired opacity to the definition.

What do you think?

xldenis:
By body do you mean the recursive expansion of every referenced function? If not how is this different than just tagging each function with an 'opaque' attribute?

jhjourdan:
> By body do you mean the recursive expansion of every referenced function?

Not the "recursive" expansion. I simply mean the body of the definition. I.e., the body of a definition can be made public if it mentions public bu opaque symbols.

> If not how is this different than just tagging each function with an 'opaque' attribute?

I don't really see what you mean, but my point is that by doing what I propose, we can rely on Rust's visibility mechanism, which is simple yet quite expressive thanks to the `pub use` construct, which makes it possible to re-export a symbol with different visibility.

Perhaps something which could be complicated if we want to implement that in Creusot is whether there exists in rustc's API a function to tell whether a symbol is public, either directly or using a different (re-exported) name. Is there such an an API?

xldenis:
> I don't really see what you mean, but my point is that by doing what I propose, we can rely on Rust's visibility mechanism, which is simple yet quite expressive thanks to the pub use construct, which makes it possible to re-export a symbol with different visibility.

What I'm suggesting and what is implemented in this PR does just that. All we do is tag functions as `opaque` with a special attribute, which seems to be equivalent to checking if it contains a `opaque_dummy` symbol in its body?

xldenis:
> Perhaps something which could be complicated if we want to implement that in Creusot is whether there exists in rustc's API a function to tell whether a symbol is public, either directly or using a different (re-exported) name. Is there such an an API?

Yes, `rustc` calculates a `DefId` visibility graph/tree which can answer the following query: is "A visible from B?"

jhjourdan:
> What I'm suggesting and what is implemented in this PR does just that. All we do is tag functions as opaque with a special attribute, which seems to be equivalent to checking if it contains a opaque_dummy symbol in its body?

Right, this has the same semantics as if `opaque_dummy` was fully private. This is surely useful.

But to really handle visibility correctly, we need to do the check of visibility of symbols used in the body (am I correct that we still do not do that?). This is required because we don't want to make visible a private symbol used in a public definition. This is useful, because thanks to the trick I described above, we can really fine-tune visibility in the cases the opaque attribute does not do want we want.

xldenis:
> This is required because we don't want to make visible a private symbol used in a public definition. 

Ok, sorry I understand what you mean now, I wasn't getting the point you're driving at.
I think the ideal solution is for us to only clone the interface of functions which should be private so that a public body can still refer to private symbols.

jhjourdan:
We should be cautious, though, because AFAIR, the interface module of a logical definition does not contain its contracts, right? An the contracts need to be public even for opaque definitions

xldenis:
Quoting myself: 

> I would still like this change to happen in the future, but I will do it properly using actual function contracts instead.

PR #372, opened 2022-05-23
Remove Model::id

### LINKED ISSUE #366: creusot::builtins does not work on trait instance functions.

xldenis:
Fixes #366


xldenis:
Closing since we actually want to use `id` so that we can attach specs to the model of slices.

PR #382, opened 2022-05-26
Update toolchain

xldenis:
Updates creusot to use the `rustc_smir` crate. However, we still need it to be merged.


xldenis:
out of date

PR #389, opened 2022-05-27
In Rust, vec cannot be larger than isize::MAX

jhjourdan:


xldenis:
Is this actually part of the spec or just an implementation detail?

jhjourdan:
The documentation of `Vec` is full of things like:

> Panics if the new capacity exceeds isize::MAX bytes.

jhjourdan:
Ah... I have no idea what happens with zero-sized types, though.

jhjourdan:
You're right, for ZST, length can reach `usize::MAX`...

Issue #387, opened 2022-05-27
Type mismatch when using extern spec for a default implementation


jhjourdan:
Here is the code:

```rust
extern crate creusot_contracts;

use creusot_contracts::*;

pub struct Tree(Option<Box<Node>>);

#[allow(dead_code)]
struct Node {
    left: Tree,
    val: u32,
    right: Tree,
}

// To force the translation of `Tree`
pub fn use_tree(_: &Tree) {}

extern_spec! {
    #[ensures(@result >= @self_)]
    #[ensures(@result >= @o)]
    #[ensures(result == self_ || result == o)]
    #[ensures(@self_ <= @o ==> result == o)]
    #[ensures(@o < @self_ ==> result == self_)]
    fn std::cmp::Ord::max<Self_>(self_: Self_, o: Self_) -> Self_ where
        Self_: Ord + Model,
        Self_::ModelTy: OrdLogic
}

impl Tree {
    pub fn height(&self) -> u64 {
        match self {
            Tree(None) => 0,
            Tree(Some(n)) => n.left.height().max(n.right.height()) + 1,
        }
    }
}
```

It gives the following error in Why3:

```
Illegal instantiation for predicate ge_log:
type mismatch between C10MutualRecTypes_Impl0_Height.ModelTy1.modelTy and int
```

xldenis:
Is it caused by the use-site or the extern spec definition?

jhjourdan:
This is caused at the use-site. It seems like it is a problem when specializing the extern spec to the type u64.

PR #394, opened 2022-05-29
New extern_spec parser

xldenis:
This PR completely overhauls the `extern_spec!` parser and provides something much more intuitive to use instead. 

Rather than having to specify functions using UFCS, users can write a set of modules, trait impls and function signatures and specify their contracts using that instead.

This means that we will be able to basically create 'interface files' for whole subtrees of Rust at once.

As an example, here is what the new version of slice specs will look like:

```rust

extern_spec! {
    impl<T> [T] {
        #[ensures((@self).len() == @result)]
        fn len(&self) -> usize;

        #[requires(@i < (@s).len())]
        #[requires(@j < (@s).len())]
        #[ensures((@^s).exchange(@*s, @i, @j))]
        fn swap(&mut self, i: usize, j: usize);
}
```

As a further improvement, when specifying `impl` or `trait` blocks where clauses can be attached to the trait/impl block and they will be properly added to the children during typecheckign / translation.

-----

However, this PR is very much unfinished:

- [x] Complete clean up of `extern_spec`. 
- [x] Substitute `self` parameters in specs.
- [x] Add more tests


xldenis:
@jhjourdan I'm giving you last word on when this should be merged as it almost certainly will conflict heavily with #388

jhjourdan:
Please merge. #388 is stuck because of the Why3 bug anyway.

xldenis:
The bug #387 should be fixed

PR #408, opened 2022-05-31
Supress unused variable warnings for laws

### LINKED ISSUE #402: For logic symbols/predicates/laws, make sure that parameters are not warned as unused when they are used in the specification.

xldenis:
Should fix #402.

The best I could do was entirely supress unused variables in trait declarations, since in the eyes of Rust there is nothing linking a law with its contracts. 

@jhjourdan: does this fix all the cases you've encountered? I couldn't find any other kinds from a cursory scan of the tests.

### DIFF
--- a/creusot-contracts-proc/src/lib.rs
+++ b/creusot-contracts-proc/src/lib.rs
@@ -83,6 +83,18 @@ impl ContractItem {
             ContractItem::Closure(_) => "closure".to_string(),
         }
     }
+
+    fn mark_unused(&mut self) {
+        if let ContractItem::TraitSig(sig) = self {
+            for arg in sig.sig.inputs.iter_mut() {
+                let attrs = match arg {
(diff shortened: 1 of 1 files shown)


jhjourdan:
Okay, this indeed fixes the issue, and let's say it's enough for now.

(However, I don't see why the approach I proposed does not work...)

Thanks!

xldenis:
> (However, I don't see why the approach I proposed does not work...)

The issue is only with _trait declarations_ in that case we have to do something a little special:

we turn

```rust
#[ensures(P)]
fn signature(...);
```

into 

```
fn ensures_uuid() { P }

fn signature(...);
```

jhjourdan:
Okay, but the unused variable warning appeared at the point of implementations, not at the point of trait declarations.

xldenis:
I wasn't seeing that, I was seeing it at the declaration site not the implementation site (using `iterators/01_range` as my guiding example).

PR #416, opened 2022-06-05
Add If-Then-Else Normalization Test

Liby99:
Add an If-Then-Else normalization test. Proved at SSFT22.


xldenis:
Thanks Ziyang, I hope you had fun playing with Creusot!

PR #425, opened 2022-06-18
Normalize line endings and file paths for Windows

Wardenfar:
Small changes for using Creusot in Windows.

There is a last problem : symlinks are not supported on windows (REPO/rust-toolchain.toml)
I can open an issue to find an other way (without symlinks)

### DIFF
--- a/creusot/tests/ui.rs
+++ b/creusot/tests/ui.rs
@@ -47,16 +47,21 @@ fn run_creusot(file: &Path, contracts: &str) -> Option<std::process::Command> {
     base_path.push("target");
     base_path.push("debug");
 
-    let mut creusot_contract_path = base_path.clone();
-    creusot_contract_path.push("libcreusot_contracts.rlib");
+    let creusot_contract_path = base_path.join("libcreusot_contracts.rlib");
+    let creusot_contract_path =
+        creusot_contract_path.to_str().expect("invalid utf-8 in contract path");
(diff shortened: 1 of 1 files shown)


xldenis:
Thanks! For the tool chain file I think we can just remove the version in `ci` entirely and just have it in the root. I can do that later if you prefer.

dewert99:
> Thanks! For the tool chain file I think we can just remove the version in `ci` entirely and just have it in the root. I can do that later if you prefer.

The symlinks also lead to another issue when using creusot-rustc
```
Caused by:
  process didn't exit successfully: `D:\Users\dewert\.cargo\bin\creusot-rustc rustc - --crate-name ___ --print=file-names --crate-type bin --crate-type rlib --crate-type dylib --crate-type cdylib --crate-type staticlib --crate-type proc-macro --print=sysroot --print=cfg` (exit code: 101)
  --- stderr
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { inner: ErrorInner { kind: Unexpected('/'), line: Some(0), col: 2, at: Some(2), message: "", key: [] } }', creusot\src\bin\creusot-rustc.rs:101:90
```
This could probably be solved by changing the include_str! on that line to `include_str!("../../../ci/rust-toolchain")`, but avoiding the symlink help prevent these types of errors more in general.

theo-certix:
In windows the symlink file is a normal file with the linked path for content.

Creusot try to parse it as toml and it obviously failed here

PR #433, opened 2022-06-22
Add theory of (infinite) sets

xldenis:
Adds a theory of (infinite) sets.

We should also add `FSet` the theory of finite sites which notably would have an induction principle and a cardinality.


jhjourdan:
Would it make sense to have a type of both finite and infinite sets, and a function of cardinality which returns an option (i.e., None when the set is infinite)?

Then it's easy to do induction on the finite set, by assuming that the cardinality is some integer.

jhjourdan:
(I don't know whether this is a good idea or not, I'm just suggesting.)

creusot-contracts/src/logic/set.rs
```diff
@@ -9,15 +9,54 @@ use std::ops::Index;
...
-    #[creusot::builtins = "set.Set.add"]
+    #[creusot::builtins = "set.Set.mem"]
+    fn mem(e: T, set: Self) -> bool {
+        pearlite! { absurd }
+    }
+
+    #[logic]
+    #[why3::attr = "inline:trivial"]
     pub fn insert(self, e: T) -> Self {
+        Self::add(e, self)
```
jhjourdan, line 30: I'm not such a great fan of having plenty of notions which are aliases from each other: insert/add remove/rem mem/contains. Why do we need these aliases?

xldenis: `mem` is private to the `Set` module, it is not exported as part of the API. It's needed as must match the signature of the `why3` function which is flipped wrt Rust conventions.

xldenis:
> Would it make sense to have a type of both finite and infinite sets, and a function of cardinality which returns an option (i.e., None when the set is infinite)?

That's an interesting idea, Why3 does this by adding a second module `Cardinal` which provides an `is_finite` predicate a set can have and then allows access to the cardinality. 

I wonder if there are underlying reasons for this? Perhaps we should just avoid depending on why3 theories entirely.

xldenis:
I'm going to merge this for now, we can revisit it to add finite sets (either via optional cardinality or seperate type).

xldenis:
Problem: this seems to break the red black tree proofs. it might be necessary to be a bit smarter about `use set.Set` in `Types`...

PR #434, opened 2022-06-22
Get rid of some trivial goals

jhjourdan:
- Use ghost!{ qname_of_lemma } instead of proof_assert!{ dummy.name_of_lemma(dummies); true } to introduce a dependency to an external lemma.
- Remove all detached goals in session files

I had to add support for restricted use of function pointers. I am not completely sure that what I have done is correct, so I'd be happy if @xldenis you could have a look to my changes.


creusot/tests/should_succeed/sparse_array.rs
```diff
@@ -119,7 +119,7 @@ impl<T> Sparse<T> {
         let index = self.idx[i];
         if !(index < self.n && self.back[index] == i) {
             // the hard assertion!
-            proof_assert!( self.lemma_permutation(@i); true );
+            ghost!( Self::lemma_permutation );
```
xldenis, line 122: Is there any way we could actually call `lemma_permutation` with valid arugments here?

creusot/tests/should_succeed/red_black_tree.rs
```diff
@@ -574,7 +574,7 @@ where
         self.node.as_mut().unwrap().color = Black;
         proof_assert! { forall<h: Int> old_self.has_height(h) ==>
         (*self).has_height(h) || (*self).has_height(h+1) }
-        proof_assert! { (*self).has_mapping_model(@key /* dummy */); true }
+        ghost! { Self::has_mapping_model };
```
xldenis, line 577: From what I can tell, the reason you need to translate`FnDef` is that you have bare function calls in these ghost macros, I wonder if we could avoid that.

creusot/src/translation/constant.rs
```diff
@@ -162,6 +163,11 @@ fn try_to_bits<'tcx, C: ToBits<'tcx>>(
             }
         }
         _ if ty.is_unit() => Exp::Tuple(Vec::new()),
+        FnDef(def_id, subst) => {
+            let method = resolve_assoc_item_opt(ctx.tcx, env, *def_id, subst).unwrap_or((*def_id, subst));
```
xldenis, line 167: unneeded

creusot/tests/should_succeed/red_black_tree.rs
jhjourdan: So you don't think having these bare functions is a good thing?

As far as I'm concerned, I don't see why having to translate FnDef is a problem.

xldenis: If we had non-linear ghost code then it would be easy to handle since we could move out of self to provide the arguments to `has_mapping_model`.

creusot/tests/should_succeed/sparse_array.rs
xldenis: Well we shouldn't be able to use `forall` or `exists` either, in fact the *only* syntax we would want to admit is `@`.

creusot/tests/should_succeed/red_black_tree.rs
xldenis: > As far as I'm concerned, I don't see why having to translate FnDef is a problem.

I have an issue with translating them to `()`, if we could translate them to function types I wouldn't really care.

jhjourdan: Well, the thing is `FnDef` types are really singletons. So it really makes sense to translate them to unit.

xldenis: Oh, true, perharps `FnPtr` could (one day) be translated as function types.

xldenis: Can you add a comment above that line in `ty.rs`?

creusot/tests/should_succeed/sparse_array.rs
xldenis: You can't use `forall` or `exists` in why3 ghost code either afaik, it's just normal why3 code with the erasure property. That's also the reason why I chose to keep borrow checking for `ghost!` and ties into the broader discussion we have been having about having tiers of ghost code.

creusot/src/translation/constant.rs
jhjourdan: Ok then why are there calls to `resolve_assoc_item_opt` here and there in the translation?

creusot/tests/should_succeed/sparse_array.rs
xldenis: > On the other hand, there is something we allow that should be forbidden: unrestricted use of logic functions are dangerous since they can themselve use prophecies.

They can't meaningfully use them since they can't perform any mutation or assumptions, and we cannot mutate borrows in ghost code. 

I do agree that we need to actually define the semantics of ghost code carefully though.

creusot/src/translation/constant.rs
xldenis: Yea, I guess we still need to do this manually here. 

This is one of the things I would like to clean up by introducing an actual IR.

jhjourdan: So should I leave this or not?

xldenis: Leave it.

creusot/tests/should_succeed/sparse_array.rs
jhjourdan: > They can't meaningfully use them since they can't perform any mutation or assumptions, and we cannot mutate borrows in ghost code.

Actually I think they can be used for unsoundness: imagine you define a logic function which returns the prophecy of a borrow. Then if you call this in ghost code, you effectively have access to `^`. So you can create a causality loop in ghost code (for a mutable borrow of a ghost variable, for example).

xldenis: how do you close the loop?

jhjourdan: Something like this:

```
let x = &mut ghost! { 1 };
*x = ghost! { ^x + 1 };
```

Then , if you resolve x at this point, you assume x = x+1. Boom.

xldenis: You can't write to mutable borrows in ghost code.

jhjourdan: The point being: for any kind of information:

(a) either you decide that it can depend on prophetic information, and then it should not be allowed to take borrows of this information (i.e., the thing is in the clairvoyant monad)
(b) or you decide that it only depends on the operational semantics, and then borrows are allowed.

Before introducing this kind of ghost code, the situation was clear: all pearlite code was (a) and all compiled code was (b). Now that we have ghost code, we have to decide whether it belongs to (a) or (b). To me, ghost code can only be really useful if it is part of (b), because then you can place ghost information in structs etc. So then you need to make sure that you cannot use prophetic information in ghost code.

jhjourdan: > You can't write to mutable borrows (modulo a bug) in ghost code.

I'm not doing this. I am writing in actual code to a variable of type `Ghost`.

xldenis: > I'm not doing this. I am writing in actual code to a variable of type Ghost.

Oh that is mind breaking.

jhjourdan: Let's discuss this in #436

PR #444, opened 2022-06-28
Redblacktree

jhjourdan:


creusot/tests/should_succeed/red_black_tree.rs
```diff
@@ -120,22 +120,26 @@ impl<K: Model, V> Tree<K, V> {
     {
         pearlite! { {
             self.has_mapping_model(k);
-            match (@self).get(k) {
-                None => (),
-                Some(_v) => ()
-            }
+            match (@self).get(k) { None => (), Some(_v) => () }
```
xldenis, line 123: I assume the match is needed for triggers?

jhjourdan: Yeah, this guides the proof. I don't remember whether that's for SMT solvers or for the automated transformations, though.

PR #448, opened 2022-06-29
Rustc smir

xldenis:
As part of the `rustc_smir` work, I'm moving all the imports in creusot to a separate `creusot-rustc` crate. Over time I would like to move them to happen exclusively through the `rustc_smir` crate, but this will take further work.


xldenis:
Since this is a big PR but that has no functional changes I'm just going to merge it now. If we wait it _will_ cause unnecessary rebase conflicts, and further changes can happen later.

PR #457, opened 2022-07-08
Use postorder DFS for clones

xldenis:
Use a postorder dfs traversal rather than topological. 

Also adds a `Drop` instance on `CloneMap` to ensure that we have cloned ever entry in the map.

This is because I somehow encountered an issue where we didn't perform clones in CDSAT and this caused the generated output to fail to type.


jhjourdan:
What is the difference between a post-order DFS and a topological order ?

xldenis:
> What is the difference between a post-order DFS and a topological order ?

Postorder DFS handles cycles. This is important because I needed to add a reflexive edge on the root node. 

 It also allows us to flip the arrows in the graph to be more intuitive: edges now go from user to dependency rather than the other way around.

xldenis:
Bizarrely enough, if all i do is swap out the dfs post order traversal for the topological one it still causes failures in CDSAT.

xldenis:
On further investigation it's still possible to use a topological sort, and I've updated the PR to do that.

xldenis:
Also fixes #464

PR #477, opened 2022-07-14
Make spans relative by default

xldenis:
Makes spans relative by default.


jhjourdan:
I'm trying to avoid using Z3 in RBT, because it is so impracticable in CI... So please avoid reintroducing it in PRs.

#479 shoudl fix the initial issue.

jhjourdan:
> Yea, but its often the best solver for the job too :/

Well, for RBT, it turns out it is not that good actually.

PR #484, opened 2022-07-15
Fix pure neq

jhjourdan:
The logical `!=` operator was not properly translated in impure context.


xldenis:
Ah, interesting bug, I think this solution is but I wonder if we couldn't just eliminate the `EqualsOrNot` constructor entirely. Perhaps leave a `TODO` expressing that intent?

jhjourdan:
I don't understand. What do you want to replace `EqualsOrNot` with? Why don't you like it?

xldenis:
> I don't understand. What do you want to replace EqualsOrNot with? Why don't you like it?

We already have a `BinaryOp` constructor, so we should use that one for everything instead.

jhjourdan:
#490 : wish granted.

PR #488, opened 2022-07-17
ghost types

xldenis:
Enables ghost fields in types. 

Also adds `FSet` the theory of finite sets.


xldenis:
I also had to change the constructor of `Ghost` to not take a parameter in passthrough mode as this prevented many programs using ghost from succesfully compiling with rustc (since we can't actually furnish the appropriate ghost value). The side-effect of this is that we may need to explicitly annotate the ghost bindings quite often.

PR #501, opened 2022-07-26
Specify `split_at_mut`

### LINKED ISSUE #499: Contract for `split_at_mut`
[split_at_mut](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.split_at_mut) a very useful method on slices, which is impossible to implement in safe rust.

suhr:
For some reason, equality in this module is on the values rather than on models of values.
Mistakes are likely, I don't have much Creusot experience.

Closes #499.

### DIFF
--- a/creusot-contracts/src/std/slice.rs
+++ b/creusot-contracts/src/std/slice.rs
@@ -104,12 +104,10 @@ extern_spec! {
         #[ensures({
             let (l,r) = result;  let sl = (@self).len();
             ((@^self).len() == sl) &&
-            ((@l).len() == @mid && (@r).len() == sl - @mid) &&
-            ((@^l).len() == @mid && (@^r).len() == sl - @mid) &&
-            (forall<i:usize> @i < @mid
-                ==> (@l)[@i] == (@self)[@i] && (@^l)[@i] == (@^self)[@i]) &&
-            (forall<i:usize> @i >= @mid && @i < sl
-                ==> (@r)[@i] == (@self)[@mid-@i] && (@^r)[@i] == (@^self)[@mid-@i])
+            (@self).subsequence(0, @mid).ext_eq(@l) &&
+            (@self).subsequence(@mid, sl).ext_eq(@r) &&
+            (@^self).subsequence(0, @mid).ext_eq(@^l) &&
+            (@^self).subsequence(@mid, sl).ext_eq(@^r)
         })]
         fn split_at_mut(&mut self, mid: usize) -> (&mut [T], &mut [T]);
 
--- a/creusot-contracts/src/std/slice.rs
+++ b/creusot-contracts/src/std/slice.rs
@@ -100,6 +100,19 @@ extern_spec! {
         })]
         fn get<I : SliceIndexSpec<[T]>>(&self, ix: I) -> Option<&<I as SliceIndex<[T]>>::Output>;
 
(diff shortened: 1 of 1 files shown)


creusot-contracts/src/std/slice.rs
```diff
@@ -100,6 +100,19 @@ extern_spec! {
...
+        #[requires(@mid <= (@self).len())]
+        #[ensures({
+            let (l,r) = result;  let sl = (@self).len();
+            ((@^self).len() == sl) &&
+            ((@l).len() == @mid && (@r).len() == sl - @mid) &&
+            ((@^l).len() == @mid && (@^r).len() == sl - @mid) &&
+            (forall<i:usize> @i < @mid
+                ==> (@l)[@i] == (@self)[@i] && (@^l)[@i] == (@^self)[@i]) &&
+            (forall<i:usize> @i >= @mid && @i < sl
+                ==> (@r)[@i] == (@self)[@mid-@i] && (@^r)[@i] == (@^self)[@mid-@i])
```
jhjourdan, line 112: The usual style is to quantify over `i: Int` and require that `0 <= i`. This makes the quantifiers easier to instantiate by provers.

jhjourdan (approved):
Minor comment on the spec (please fix), appart from that, this seems to be OK to me.

creusot-contracts/src/std/slice.rs
```diff
@@ -100,6 +100,19 @@ extern_spec! {
         })]
         fn get<I : SliceIndexSpec<[T]>>(&self, ix: I) -> Option<&<I as SliceIndex<[T]>>::Output>;
 
+        #[requires(@mid <= (@self).len())]
+        #[ensures({
+            let (l,r) = result;  let sl = (@self).len();
+            ((@^self).len() == sl) &&
+            ((@l).len() == @mid && (@r).len() == sl - @mid) &&
+            ((@^l).len() == @mid && (@^r).len() == sl - @mid) &&
+            (forall<i:usize> @i < @mid
```
xldenis, line 109: could you use `Seq::subsequence` here? Like for [IteratorSpec](https://github.com/xldenis/creusot/blob/master/creusot-contracts/src/std/slice.rs#L172)

jhjourdan:
> For some reason, equality in this module is on the values rather than on models of values.

That is expected: `==` denotes logical equality in pearlite, and equality on models in Rust code.

creusot-contracts/src/std/slice.rs
jhjourdan: Oh, indeed this is simpler. And then you don't even need to say much about the lengths.

xldenis (approved):
thanks! This looks correct to me!

jhjourdan:
Actually, I think a simpler specification (and perhaps more effective for provers, who knows...) would be to say that self is the concatenation of l and r, and that l and r have the right sizes.

I don't know whether the provers would like this more, but at least this better captures the intent, I think.

xldenis:
> Actually, I think a simpler specification (and perhaps more effective for provers, who knows...) would be to say that self is the concatenation of l and r, and that l and r have the right sizes.

That's a good point, I wish i had a better intuition for which spec would be preferred by provers. 
This is a further reason that supporting SMT sequences in Why3 would be a huge improvement..

PR #504, opened 2022-07-26
Specify more operations on `Option`

LevitatingLion:


xldenis:
Could you split the specs involving closures to a separate PR? I'd like to review those more closely.

creusot-contracts/src/std/option.rs
```diff
@@ -1,25 +1,213 @@
...
                 fn unwrap(self) -> T;
 
+                #[requires(self != None)]
+                #[ensures(Some(result) == self)]
+                fn expect(self, msg: &str) -> T;
+
+                #[ensures(result == match self {
+                    Some(inner) => inner,
+                    None => default,
+                })]
```
jhjourdan, line 26: Pattern matching are usually a bad idea in post-conditions, they are badly handled by provers. It's usually better to have two post-conditions, one for None and one for Some (with no quantifier or an existential quantifier if possible, or a forall quantifier.).

creusot-contracts/src/std/option.rs
```diff
@@ -1,25 +1,213 @@
...
+                fn unwrap_or(self, default: T) -> T;
+
+                #[requires(self != None || f.precondition(()))]
+                #[ensures(match self {
+                    Some(inner) => result == inner,
+                    None => f.postcondition_once((), result),
+                })]
```
jhjourdan, line 33: Idem.

creusot-contracts/src/std/option.rs
```diff
@@ -1,25 +1,213 @@
...
-                #[ensures(*self == None || exists<r: &mut T> result == Some(r) && *self == Some(*r))]
+                #[ensures(
+                    *self == None || exists<r: &mut T> result == Some(r) && *self == Some(*r)
+                )]
                 fn as_ref(&self) -> Option<&T>;
+
+                #[ensures(result == match self {
+                    Some(inner) => Ok(inner),
+                    None => Err(err),
+                })]
```
jhjourdan, line 54: Idem.

creusot-contracts/src/std/option.rs
```diff
@@ -1,25 +1,213 @@
...
+                fn ok_or<E>(self, err: E) -> Result<T, E>;
+
+                #[requires(self != None || err.precondition(()))]
+                #[ensures(match self {
+                    Some(inner) => result == Ok(inner),
+                    None => exists<e: E> err.postcondition_once((), e) && result == Err(e),
+                })]
```
jhjourdan, line 61: Idem.

creusot-contracts/src/std/option.rs
```diff
@@ -1,25 +1,213 @@
...
+
+                #[requires(match self {
+                    Some(inner) => f.precondition((inner,)),
+                    None => true,
+                })]
+                #[ensures(match self {
+                    Some(inner) =>
+                        exists<r: U> f.postcondition_once((inner,), r) && result == Some(r),
+                    None => result == None,
+                })]
```
jhjourdan, line 74: Idem.

creusot-contracts/src/std/option.rs
```diff
@@ -1,25 +1,213 @@
...
+                where
+                    F: FnOnce(T) -> U;
+
+                #[requires(match self {
+                    Some(inner) => f.precondition((inner,)),
+                    None => true,
+                })]
+                #[ensures(match self {
+                    Some(inner) => f.postcondition_once((inner,), result),
+                    None => result == default,
```
jhjourdan, line 85: Idem.

jhjourdan (commented):
Thanks for your proposal! It is usually a bad idea to use pattern matching in post-conditions, because they are badly handled by provers. Could you please update the specs?

jhjourdan:
Have you tested these specs? I've been slightly lazy introducing new specs like these, because I'm afraid the provers handle them badly and that we need to reformulate them.

LevitatingLion:
I reformulated the post-conditions to avoid pattern-matching, and added a test that exercises all functions.

The tests of the functions involving closures are currently commented out, because they cause invalid mlcfg to be generated. I didn't really look into it, but it seems specific to closures.

xldenis:
> The tests of the functions involving closures are currently commented out, because they cause invalid mlcfg to be generated. I didn't really look into it, but it seems specific to closures.

Yea, that doesn't surprise me that much, there's still work to do fixing bugs related to closures. Can you remove the relevant specifications and put them in a different PR?

LevitatingLion:
I removed the specs related to closures, and will open another PR for them once this is merged

xldenis (approved):
lgtm

PR #505, opened 2022-07-26
Specify checked, wrapping, overflowing, saturating operations on integer types

LevitatingLion:
And add the builtins `int.EuclideanDivision.div` as `Int::div_euclid` and `int.EuclideanDivision.mod` as `Int::rem_euclid`.


xldenis:
Awesome! Thanks for all these contributions. Could you add a test case verifying these specs work as intended?

creusot-contracts/src/std/num.rs
```diff
@@ -0,0 +1,85 @@
...
+macro_rules! spec_op {
+    (
+        $type:ty,
+        $op:tt,
+        $checked:ident,
+        $wrapping:ident,
+        $saturating:ident,
+        $overflowing:ident $(,)?
+    ) => {
+        extern_spec! {
```
xldenis, line 27: macros generating macros generating macros 😄

creusot-contracts/src/std/num.rs
```diff
@@ -0,0 +1,85 @@
...
+                #[allow(dead_code)]
+                #[ensures(
+                    (result == None)
+                    == ((@self $op @rhs) < @$type::MIN || (@self $op @rhs) > @$type::MAX)
+                )]
+                #[ensures(forall<r: $type> result == Some(r) ==> @r == (@self $op @rhs))]
+                fn $checked(self, rhs: $type) -> Option<$type>;
+
+                // Wrapping: performs the operation on `Int` and converts back to `$type`
+                #[allow(dead_code)]
```
xldenis, line 41: You shouldn't need to `allow(dead_code)`.

creusot-contracts/src/std/num.rs
```diff
@@ -0,0 +1,85 @@
...
+                #[ensures(
+                    @result == (@self $op @rhs).rem_euclid(2.pow(@$type::BITS)) + @$type::MIN
```
xldenis, line 43: I wonder how usable this specification will be in practice. Perhaps there's a good way to lower this to SMT so that they will be able to reason (bit vectors?)

jhjourdan: Probably we should have a post-condition saying its in bounds, and another saying they have the same value modulo  2^n, i.e., their difference is a multiple.

We should also have a post-condition saying that if self is in bounds, then the result is equal.

LevitatingLion:
I split division from the `spec_op!` macro since it has the additional precondition that the divisor is non-zero, and it only overflows in the single case of `MIN / -1` (which the prover is not able to deduce from the general specification).

I also added additional postconditions to `wrapping_*` and `overflowing_*`, asserting that the result is just the result of the operation if that is in range, and that it is offset by a multiple of the type's range otherwise (as suggested by @jhjourdan, and the prover is unable to verify all the testcases without these postconditions).

And I added some tests, which can be proved by CVC4 or Z3 except for the two `*_overflowing_mul`, which CVC4 is unable to prove (let me know if you have an idea how to fix that).

I agree that the `#[allow(dead_code)]` shouldn't be needed, but I do get warnings without them; probably caused by a specific interaction between `extern_spec!` and that the code inside is generated by a macro.

xldenis:
Thanks for addressing the comments and for the extensive test cases. I think this PR is ready to merge.

xldenis:
Looking at CI, I'll need to replay your proofs with a slightly older Z3, and longer term we need a CI image where we control the versions more tightly.

LevitatingLion:
I only tested the proofs with my local why3 installation, not in the CI image. I can do that tonight, maybe it fixes the version mismatch.

xldenis:
> I only tested the proofs with my local why3 installation, not in the CI image. I can do that tonight, maybe it fixes the version mismatch.

Ill just push a commit fixing them.

LevitatingLion:
Awesome!

xldenis:
> Awesome!

Thanks to you! I'm looking forward to getting your other PRs merged soon as well.

PR #506, opened 2022-07-26
Specify indexing slices with ranges

LevitatingLion:
`RangeInclusive` is missing because its members are private, so we cannot access its start and end in specifications. The changes in tests only adjust changing `impl` numbers.

This also fixes a bug in `[T]::get`: It should not have a precondition on the index being in bounds.


xldenis:
Awesome! Do you think you could add a test case which uses these specifications? Doesn't have to be anything crazy but I'd feel more comfortable with a test case verifying that it all works as expected.

xldenis:
> This also fixes a bug in [T]::get: It should not have a precondition on the index being in bounds.

😬 my bad, thanks for catching that.

creusot-contracts/src/std/slice.rs
```diff
@@ -83,6 +83,96 @@ impl<T> SliceIndexSpec<[T]> for usize {
...
+
+    #[predicate]
+    fn has_value(self, seq: Seq<T>, out: [T]) -> bool {
+        pearlite! { seq.subsequence(@self.start, @self.end) == @out }
+    }
+
+    #[predicate]
+    fn resolve_elswhere(self, old: Seq<T>, fin: Seq<T>) -> bool {
+        pearlite! {
+            forall<i : Int> 0 <= i && (i < @self.start || @self.end <= i) && i < old.len()
```
xldenis, line 100: As a pragmatic issue I wonder if provers would prefer that this is specified using two seperate quantifiers, one for the lower segment and one for the upper segment?

I don't know the answer nor do I have a clear idea on the best way to test it..

creusot-contracts/src/std/slice.rs
```diff
@@ -83,6 +83,96 @@ impl<T> SliceIndexSpec<[T]> for usize {
...
+    }
+
+    #[predicate]
+    fn has_value(self, seq: Seq<T>, out: [T]) -> bool {
+        pearlite! { seq == @out }
+    }
+
+    #[predicate]
+    fn resolve_elswhere(self, old: Seq<T>, fin: Seq<T>) -> bool {
+        pearlite! { old == fin }
```
xldenis, line 155: I think this should actually be `true` since 'elsehwere' is the empty sequence.

xldenis:
@LevitatingLion will you have time soon to address my comment?  Or would you rather I take over?

LevitatingLion:
I fixed `resolve_elswhere` for `RangeFull` and added some tests for all the ranges. I also reformulated the postcondition of `[T]::get` to get the reading part of the tests to succeed.

The test cases that write to part of the array currently fail; the provers seem to be unable to prove that the part of the array that was mutably borrowed but not written to through that borrow remains the same (e.g. line 68: `arr[3]` is part of the mutably borrowed slice, but was not written to and thus remains `3`). Any ideas on how to fix that?

creusot-contracts/src/std/slice.rs
LevitatingLion: I have no idea what is more beneficial to provers, but once this PR is done we could run the test with both options, to see if one of them results in consistently fewer steps for the provers?

xldenis:
> The test cases that write to part of the array currently fail; the provers seem to be unable to prove that the part of the array that was mutably borrowed but not written to through that borrow remains the same (e.g. line 68: arr[3] is part of the mutably borrowed slice, but was not written to and thus remains 3). Any ideas on how to fix that?

I'll take a look at this now.

xldenis:
At a quick glance it seems like it's specifically the index just after the end of the `Range` which is causing an issue, which leads me to think there's an off-by-one error in `resolve_elsewhere`.

xldenis:
I was able to get it to pass by manually asserting that `s[2] == 3`. This indicates to me that all the information is _there_ but perhaps not exploitable by the provers for whatever reason. I'll see if I can adjust definitions so that they accept them more willingly.

xldenis:
I _think_ this is caused by an interaction with `resolve_elsewhere with `usize` indices. 

After the first two calls the solver knows that all the indices of `s` which are not 0 or 1 are unchanged, but on its own it won't figure out that this means that 2 is unchanged (it's not going to pull 2 out of thin air to instantiate the quantifier). And if you don't know that `s[2]` is unchanged then it won't figure out that `arr[3]` is unchanged. 

I don't know what we can do apart from use the real SMT theory of sequences in why3 which should allow us to get rid of all these quantifiers.

xldenis:
I would say: add an `assert!` or `proof_assert` and a comment explaining _why_ and we'll call it a day.

LevitatingLion:
Great, I added an assert like you described and now Z3 is able to prove the tests easily

PR #509, opened 2022-07-26
Adds support for the Model trait and "logic only" types when compiling without "contracts"

dewert99:


creusot-contracts/src/lib.rs
```diff
@@ -162,6 +162,20 @@ pub mod logic {
...
+
+    pub struct Mapping<A: ?Sized, B: ?Sized>(std::marker::PhantomData<A>, std::marker::PhantomData<B>);
+
+    pub struct Seq<T: ?Sized>(std::marker::PhantomData<T>);
+
+    pub struct FSet<T: ?Sized>(std::marker::PhantomData<T>);
+
+    pub struct Set<T: ?Sized>(std::marker::PhantomData<T>);
+
+    pub struct Int;
```
xldenis, line 174: Can you instead remove the conditional flag on the relevant modules? They should compile just fine with `contracts` turned off (since `#[predicate]` and `#[logic]` will turn to no-ops.

dewert99: I tried doing that but it quite that easy, most types had `creusot::builtins = ...` attributes which needed to be disabled, the `EMPTY` constants also needed to be disabled since they had attributes but not `#[predicate]` or `#[logic]`, and implementations of standard library traits needed to be disabled since they did have `#[logic]` functions got removed leaving incomplete implementations.

xldenis: > I tried doing that but it quite that easy, most types had creusot::builtins = ... attributes which needed to be disabled, the EMPTY constants also needed to be disabled since they had attributes but not #[predicate] or #[logic], and implementations of standard library traits needed to be disabled since they did have #[logic] functions got removed leaving incomplete implementations.

Right, that's annoying.

xldenis (approved):
Thanks, this looks good! I'm going to merge this now.

PR #515, opened 2022-08-04
Specify operations on `Result`

LevitatingLion:
Similar to #504, but now for `Result<T, E>`.


xldenis:
Something that would be nice especially for these kinds of safe apis would be to actually prove their specs on a re-implementation. These functions typically are 1-2 lines each, and this would boost confidence in their specifications.

xldenis:
:wave: would you like me to take this over?

PR #526, opened 2022-08-17
Add test for hillel exercises

voidc:
This also adds specifications for `Vec::insert` and `abs_diff`.


creusot/tests/should_succeed/hillel.rs
```diff
@@ -0,0 +1,206 @@
...
+
+#[predicate]
+fn is_unique<T: Eq + Model>(s: Seq<T>) -> bool {
+    pearlite! {
+        forall<i: Int, j :Int> 0 <= i && i < s.len() && 0 <= j && j < s.len() ==> @(s[i]) == @(s[j]) ==> i == j
+    }
+}
+
+#[predicate]
+fn contains<T: Eq + Model>(seq: Seq<T>, elem: T) -> bool {
```
xldenis, line 56: Exists in `Seq` now.

creusot/tests/should_succeed/hillel.rs
```diff
@@ -0,0 +1,206 @@
...
+    pearlite! {
+        exists<i: Int> 0 <= i && i < seq.len() && @(seq[i]) == @elem
+    }
+}
+
+#[predicate]
+fn is_subset<T: Eq + Model>(sub: Seq<T>, sup: Seq<T>) -> bool {
```
xldenis, line 63: The `Eq` and `Model` bounds should be redundant.

xldenis (commented):
Thanks, overall this looks pretty good. A few comments and ci needs fixing but then we'll get this merged!

creusot-contracts/src/std/num.rs
```diff
@@ -157,6 +157,35 @@ macro_rules! spec_op_common {
...
+    if a < b {
+        b - a
+    } else {
+        a - b
+    }
+}
+
+/// Adds specifications for the abs_diff operation on the given pair of signed
+/// and unsigned integer types
+macro_rules! spec_abs_diff {
```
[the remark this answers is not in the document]
voidc: I don't think so because abs_diff requires knowing both the signed and the unsigned type (the signed variant of the function returns the unsigned type).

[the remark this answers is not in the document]
xldenis: Right, I was just looking at he contracts.

creusot/tests/should_succeed/hillel.rs
voidc: Good to know! However, there is a slight difference between the two functions: my contains checks for equality of the elements' models while the one in `Seq` directly compares the elements. WIth the new contains function, the [assertion](https://github.com/xldenis/creusot/pull/526/files#diff-62b5c0619a5e5b84a19bf7edfeeb213af2ef0ddb2e67817cfbac6e1870853eafR87) `proof_assert! { contains(@vec, elem) };` no longer holds after checking `vec[i] == elem`.

xldenis: Ah yea, that's a good point, god models are such a pain sometimes...

creusot/tests/should_succeed/hillel.rs
voidc: I removed some of the `Eq` bounds but `Model` is still needed for `contains`.

xldenis:
@voidc can you run "Clean" on your session? It will remove all the failed goals and allow CI to pass.

PR #527, opened 2022-08-18
Correct GTK sourceview dependency name

jhaye:
I'm not familiar with OCaml at all, but the dependency as listed does not resolve and with these changes it [does](https://opam.ocaml.org/packages/lablgtk3-sourceview3/).


xldenis:
Thanks for correcting that!

Issue #510, opened 2022-08-01
Add support for casting booleans to integer types


sarsko:
Casting a `bool` to any integer type, as in for instance:
```rust
let bing = false as u8;
```
currently yields `error[creusot]: Non integral casts are currently unsupported`

gmorenz:
This now panics

```
    Checking testbin v0.1.0 (/home/greg/p/games/bb/testbin)
warning: unused variable: `bing`
 --> src/main.rs:3:9
  |
3 |     let bing = false as u8;
  |         ^^^^ help: if this is intentional, prefix it with an underscore: `_bing`
  |
  = note: `#[warn(unused_variables)]` on by default

thread 'rustc' panicked at 'internal error: entered unreachable code', creusot/src/translation/fmir.rs:164:26
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

note: Creusot has panic-ed!
  |
  = note: Oops, that shouldn't have happened, sorry about that.
  = note: Please report this bug over here: https://github.com/xldenis/creusot/issues/new
```

Issue #482, opened 2022-07-15
Better error message when using `char`


suhr:
Essential for leftpad verification. I guess it makes sense to interpret it as an integer from 0 to 0x10FFFF. This allows for char ranges, for example.

xldenis:
> Essential for leftpad verification. I guess it makes sense to interpret it as an integer from 0 to 0x10FFFF. This allows for char ranges, for example.

You can define leftpad in a generic manner over any slice of type `T`. 

I think char will require a new abstract mathematical type since we don't want to allow things like addition of characters which is ill-defined.

xldenis:
Since we already compile `char` to why3 `char.Char` I believe the suitable model here is just the identity.

jhjourdan:
Why3's char is unicode?

xldenis:
No, it isn't so we won't actually be able to use it. The SMT Char is though.

gmorenz:
I hit two char related panics. They're both just unimplemented branches, but since they asked me to report them and I don't see any direct mentions in the issue  I thought I'd make a note here - it seems to be the relevant issue for replacing implementing them.

```rust
use creusot_contracts::trusted;

#[trusted]
fn foo() -> char {
    'a'
}

fn main() {
    // Panic 1: thread 'rustc' panicked at 'not implemented', creusot/src/translation/function/terminator.rs:313:14
    match foo() {
        'a' => (),
        _ => (),
    }

    // Panic 2: thread 'rustc' panicked at 'internal error: entered unreachable code', creusot/src/translation/fmir.rs:170:26
    22 as char;
}
```

jhjourdan:
We agreed that proper support for `char` is not in our priorities, so we should instead just have good error messages.

PR #544, opened 2022-09-06
Improve casts and char handlign

xldenis:
Improve basic casts and handling of `char`. 
Also explicitly marks `char` as partial / experimental since Why3 unfortunately uses ASCII strings...


xldenis:
Fixes #510

xldenis:
Partially fixes #482

PR #546, opened 2022-09-06
Fix #545

### LINKED ISSUE #545: Dropping negative signs
Creusot appears to be dropping negative signs when translating requirements to whyml

This code

```rust
use creusot_contracts::{ensures, requires, OrdLogic};

#[requires(a < 100i32)]
#[requires(b < 100i32)]
#[requires(a > -100i32)]
#[requires(b > -100i32)]
#[ensures(result == a - b)]
fn sub(a: i32, b: i32) -> i32 {
    7
}
```

Produces this output, with the lower bounds now `100` instead of `-100`.

```why3
module Testbin_Sub_Interface
  use mach.int.Int
  use mach.int.Int32
  val sub [@cfg:stackify] (a : int32) (b : int32) : int32
    requires {[#"../src/main.rs" 3 11 3 21] a < (100 : int32)}
    requires {[#"../src/main.rs" 4 11 4 21] b < (100 : int32)}
    requires {[#"../src/main.rs" 5 11 5 22] a > (100 : int32)}
    requires {[#"../src/main.rs" 6 11 6 22] b > (100 : int32)}
    ensures { [#"../src/main.rs" 7 10 7 25] result = a - b }
    
end
module Testbin_Sub
  use mach.int.Int
  use mach.int.Int32
  let rec cfg sub [@cfg:stackify] [#"../src/main.rs" 8 0 8 29] (a : int32) (b : int32) : int32
    requires {[#"../src/main.rs" 3 11 3 21] a < (100 : int32)}
    requires {[#"../src/main.rs" 4 11 4 21] b < (100 : int32)}
    requires {[#"../src/main.rs" 5 11 5 22] a > (100 : int32)}
    requires {[#"../src/main.rs" 6 11 6 22] b > (100 : int32)}
    ensures { [#"../src/main.rs" 7 10 7 25] result = a - b }
    
  var _0 : int32;
  var a_1 : int32;
  var b_2 : int32;
  {
    a_1 <- a;
    b_2 <- b;
    goto BB0
  }
  BB0 {
    _0 <- (7 : int32);
    return _0
  }
  
end
```

It seems like I can work around this by replacing `-100` with `0 - 100`.

xldenis:
Fixes an issue where the negative sign of literals was dropped in when translating pearlite.


gmorenz:
This will overflow on `-170141183460469231731687303715884105728` (`i128::MIN`)

xldenis:
Argh right. Hmm, I suppose we have to manually construct the correct signed representation in two's complement

gmorenz:
Just replacing `-` with `.wrapping_neg()` should work, will submit a pr

PR #547, opened 2022-09-06
Fix overflow on i128::MIN

gmorenz:


gmorenz:
I manually tested that  the following produces the expected value, and with `-` instead of `wrapping_neg` it panics.

```rust
let x = 170_141_183_460_469_231_731_687_303_715_884_105_728u128;
println!("{}", (x as i128).wrapping_neg());
```

xldenis (approved):
thanks!

Issue #552, opened 2022-09-09
Modules emitted in the wrong order (Library file not found: why3.Testbin_Impl0_Transition_Interface)


gmorenz:
The following code ends up emitting `Testbin_Impl0_Transition_Interface` before it's used in `Testbin_Impl1_Step`. This results in the error message `Library file not found: why3.Testbin_Impl0_Transition_Interface` when trying to open it in why3. Simply cut-and-pasting the Interface to before it's used in the file fixes the problem.

```rust
use creusot_contracts::{logic, ensures};

pub struct Transition;
pub struct Machine;

#[logic]
pub fn transition_logical(this: &Machine) -> bool {
    this.description_logical();
    true
}

impl Machine {
    #[ensures(transition_logical(self))]
    pub fn transition(&self) -> Transition {
        Transition
    }
}

pub trait MachineTrait {
    #[logic]
    fn description_logical(self);

    fn step(&mut self);
}

impl MachineTrait for Machine {
    #[logic]
    fn description_logical(self) {}

    fn step(&mut self) {
        self.transition();
    }
}

fn main() {}
```

[And the .mlcfg (with .log added to the filename because github filters file uploads by extension](https://github.com/xldenis/creusot/files/9537655/testbin-bin.mlcfg.log)

Related to #457.

xldenis:
There must be a missing dependency edge. I really want to overhaul the handling of all of this soon..

gmorenz:
Better reproduction, this one not just emits things in the wrong order, it also fails to emit a `with` clause which makes it un-provable - which means I can't just work around this with a script that re-arranges modules.

```rust
use creusot_contracts::{predicate, maintains, requires, ensures};

pub struct Transition;
pub struct Machine;

pub trait MachineTrait {
    #[predicate]
    fn invariants(self) -> bool;

    #[maintains((mut self).invariants())]
    fn step(&mut self) -> bool;
}

impl MachineTrait for Machine {
    #[predicate]
    fn invariants(self) -> bool {
        true
    }

    #[maintains((mut self).invariants())]
    fn step(&mut self) -> bool {
        self.transition(); // Comment out this line and it works
        false
    }
}

impl Machine {
    #[requires(self.invariants())]
    pub fn transition(&self) -> Transition {
        Transition
    }
}
```

My current understand of the problem is that:

[...]

PR #554, opened 2022-09-12
Split deps out from TranslatedItems

gmorenz:
This splits dependencies away from translated items, so we can store the dependencies for a function before recursively translating function it calls.

See discussion in #552, but while this does fix an issue where dependencies aren't properly substituted in a cloned theorem, it doesn't fix the the main ordering issue in #552.

---

Unfortunately I don't have a good test case for what this does fix, because my reproducing code still outputs the modules in the wrong order, they just have correct contents now. This is specifically the difference between the following outputs

Old:
```
module Testbin_Impl0
 [...]  
clone Testbin_Impl0_Step_Interface as Step0
  [...]
end
```

New:

```
module Testbin_Impl0
  [...]
  clone Testbin_Impl0_Step_Interface as Step0 with predicate Invariants0.invariants = Invariants0.invariants
  [...]
end
```

on this code (the second reproducer in #552)

```rust
use creusot_contracts::{predicate, maintains, requires, ensures, logic};

pub struct Transition;
pub struct Machine;

pub trait MachineTrait {
    #[predicate]
    fn invariants(self) -> bool;

    #[maintains((mut self).invariants())]
    fn step(&mut self) -> bool;
}

impl MachineTrait for Machine {
    #[predicate]
    fn invariants(self) -> bool {
        true
    }

    #[maintains((mut self).invariants())]
    fn step(&mut self) -> bool {
        self.transition();
        false
    }
}

impl Machine {
    #[requires(self.invariants())]
    pub fn transition(&self) -> Transition {
        Transition
    }
}
```

### DIFF
--- a/creusot/src/ctx.rs
+++ b/creusot/src/ctx.rs
@@ -39,7 +39,8 @@ pub struct TranslationCtx<'sess, 'tcx> {
     pub tcx: TyCtxt<'tcx>,
     pub translated_items: IndexSet<DefId>,
     ty_binding_groups: HashMap<DefId, DefId>, // maps type ids to their 'representative type'
-    functions: IndexMap<DefId, TranslatedItem<'tcx>>,
+    functions: IndexMap<DefId, TranslatedItem>,
+    dependencies: IndexMap<DefId, CloneSummary<'tcx>>,
     terms: IndexMap<DefId, Term<'tcx>>,
     pub externs: Metadata<'tcx>,
     pub(crate) opts: &'sess Options,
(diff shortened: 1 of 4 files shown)


xldenis:
I'm fine with merging this as is if there are no regressions, it's a monotonic improvement in code quality. Can you run `./fmt` fist though?

gmorenz:
Oops, didn't realize that was distinct from `cargo fmt`, or maybe I forgot to run `cargo fmt` after fixing a late nit...

xldenis:
> Oops, didn't realize that was distinct from cargo fmt, or maybe I forgot to run cargo fmt after fixing a late nit...

It's because `cargo fmt` won't format the test suite otherwise :| it just runs `rustfmt **/*.rs`

PR #555, opened 2022-09-13
Map

xldenis:
Some more proofs about iterators.


xldenis:
I've significantly weakend the conditions on the invariant of `Map`, but this comes at a cost: the proof of `next` is _extremely_ painful to get to go through, and it seems like instantiating `map` for closures is trickier than with the simplified versions. If we can remove ` self.iter.produces(initial.produced.subsequence(self.produced.len(), initial.produced.len()), initial.iter) ==>` then it _significantly_ speeds things up. Otherwise perhaps some refactoring could help guide provers?

xldenis:
I've proved that an identity map and incrementing map are safe but have not yet checked their results (i'm scared of looking at the proofs).

xldenis:
@jhjourdan I'm going to merge `Map` despite not having working benchmarks. We can fine tune the spec of `map` afterwards.

PR #556, opened 2022-09-15
Iterators: skip

jhjourdan:


creusot/tests/should_succeed/iterators/01_range.rs
```diff
@@ -39,10 +42,9 @@ impl Iterator for Range {
     #[ensures(a.produces(ab.concat(bc), c))]
     fn produces_trans(a: Self, ab: Seq<Self::Item>, b: Self, bc: Seq<Self::Item>, c: Self) {}
 
-    // #[requires(!(*self).completed())]
     #[ensures(match result {
-      None => (^self).completed() && self.resolve(),
-      Some(v) => (*self).produces(Seq::singleton(v), ^self) && !(*self).completed()
+      None => (*self).completed(^self),
```
xldenis, line 46: Can we instead make `completed` take a `&mut Self`?

jhjourdan: I considered doing that, but:
- It is not uniform with produces
- it's not easy to write the spec for Skip, because we don't actually have this borrow in hand.

xldenis: > It is not uniform with produces

Produces needs that since we are talking about arbitrary numbers of steps

> it's not easy to write the spec for Skip, because we don't actually have this borrow in hand.

In your spec you have `(*self).completed(^self)`, don't you just need to replace that by `self.completed()`?

xldenis:
To avoid having to regenerate a bunch of sessions for trivial things like alt-ergo patch version differences, I recommend using this command: `cargo test --test why3 -- --diff-from=master --replay=obsolete` it will only regenerate the sessions for `mlcfg` files which have changed.

creusot/tests/should_succeed/iterators/01_range.rs
jhjourdan: Nope:

```rust
   #[predicate]
    fn completed(self, o: Self) -> bool {
        pearlite! {
            @o.n == 0 &&
            exists<s: Seq<Self::Item>, i: I>
                s.len() <= @self.n &&
                self.iter.produces(s, i) &&
                i.completed(o.iter)
        }
    }
```

There is this `i.completed(o.iter)`. I could existantially quantify over a mutable borrow, though. It just felt a bit simpler that way. As you prefer.

xldenis: Ah right, hrm... I _think_ I would prefer the signature to be 'simpler' but having to pull a borrow out of thin air also sucks.

jhjourdan: This borrow does not comes from nowhere either. It actually exists when executing the `next` function of `Skip`. So I am doing this change.

jhjourdan:
That seems ready. Merging.

PR #565, opened 2022-09-16
Only emit implies in terms, and don't nest Pure blocks

### LINKED ISSUE #564: Why3 "syntax error" due to misplaced `pure`
The placement of `pure { ... }` seems to be wrong, resulting in why3 returning "syntax error".

Here are two reproductions (I'm guessing that these have the same root cause):

```rust
#[allow(unused_imports)]
use creusot_contracts::{requires, pearlite, logic};

#[logic]
#[requires(invariants())]
fn step_logical() -> bool {
    pearlite! {
        (1i32 == 1i32) ==> true
        // Results in
        // pure {(1 : int32) = (1 : int32)} -> true
        // Should be
        // pure {(1 : int32) = (1 : int32) -> true}

        // Similar, but with two pures instead of a misplaced pure.
        // Which also results in syntax_error.
        // (1i32 == 1i32) == true
        // Resutls in
        // pure {pure {(1 : int32) = (1 : int32)} = true}
        // Should be
        // pure { (1 : int32) = (1 : int32) = true}
    }
}

#[logic]
fn invariants() -> bool {
    true
}

fn main() {}
```

Context from which the why3 snippets above are taken

```whyml
module Testbin_StepLogical_Impl
  use mach.int.Int
  use mach.int.Int32
  clone Testbin_Invariants as Invariants0
  let rec ghost function step_logical (_ : ()) : bool
    requires {[#"../src/main.rs" 5 11 5 23] Invariants0.invariants ()}

    [#"../src/main.rs" 7 4 17 5] pure {(1 : int32) = (1 : int32) -> true}
end
```

gmorenz:
Fixes #564

Putting these commits in one PR because my test for the first fix relies on the second fix as well (the test is only included in the second commit)... and it seems harmless enough. 

I'm not sure why BinOps handle nested exprs differently from everything else? They first lower them, and then use a variable to get around the issue that they might need to be pure while having been lowered in an impure fashion. Everything else just lowers their sub-expressions in a context with the correct version of purity. The latter method seems more elegant, but I may be missing some complexity.

### DIFF
--- a/creusot/src/translation/specification/lower.rs
+++ b/creusot/src/translation/specification/lower.rs
@@ -235,7 +235,12 @@ impl<'tcx> Lower<'_, '_, 'tcx> {
                 Exp::Final(box self.lower_term(term))
             }
             TermKind::Impl { box lhs, box rhs } => {
-                Exp::Impl(box self.lower_term(lhs), box self.lower_term(rhs))
+                let exp = Exp::Impl(box self.lower_term(lhs), box self.lower_term(rhs));
+                if Purity::Program == self.pure {
+                    Exp::Pure(box exp)
+                } else {
+                    exp
+                }
             }
             TermKind::Old { box term } => Exp::Old(box self.lower_term(term)),
             TermKind::Match { box scrutinee, mut arms } => {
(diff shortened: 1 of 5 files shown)


xldenis:
The reason that BinOp is treated differently is that `pure` affects type inference! 

For example: `(a : int32) + (b : int32)` has type `int32` while `pure { (a : int32) + (b : int32) }` has type `int`. So the code factors out arguments into a let binding when necessary.

xldenis:
Thanks for the bug fixes! I think `Impl` needs the same treatment as `Eq`, but really, it should just be a `BinOp` as well!

PR #566, opened 2022-09-16
Refactor code that creates pure sub-exps out

gmorenz:
Adds a single commit on top of #565 to refactor some shared code out. I prefer it this way, but it's definitely a stylistic choice and I don't know if you will. Thought I'd push this separately to make it easy for you to choose to include it, or not.

GitHub doesn't have proper stacked PRs, but I think if you merge #565 first this should automatically reduce itself to the last commit.


xldenis:
This is definitely better, I don't know why I didn't originally do it..

PR #571, opened 2022-09-17
Add dependency on parent type when indexing fields

### LINKED ISSUE #570: Why3 "unbound function or predicate symbol" when accessing nested struct fields
Something broke `a.b.c` in the last few commits to master. I'll try and see if I can fix this.

Reproduction:

```rust
struct S1 {
    f: i32,
}

struct S2 {
    s1: S1,
}

fn test(s: S2) {
    s.s1.f;
}

fn main() {}
```

Error

```
unbound function or predicate symbol 'Testbin_S1_Type.s1_f'
```

Relevant Why3 Module

```
module Testbin_Test
  use Testbin_S2_Type as Testbin_S2_Type
  use mach.int.Int
  use mach.int.Int32
  let rec cfg test [@cfg:stackify] [#"../src/main.rs" 9 0 9 14] (s : Testbin_S2_Type.t_s2) : ()
  var _0 : ();
  var s_1 : Testbin_S2_Type.t_s2;
  var _2 : int32;
  {
    s_1 <- s;
    goto BB0
  }
  BB0 {
    _2 <- Testbin_S1_Type.s1_f (Testbin_S2_Type.s2_s1 s_1);
    _0 <- ();
    return _0
  }
  
end
```

Can be manually fixed by adding the following

```
  use Testbin_S1_Type as Testbin_S1_Type
```

gmorenz:
Fixes #570.

This takes the approach of just importing the modules that we use explicitly, which of the 3 approaches mentioned in #570 feels like the cleanest to me... but since I just posted a question there about whether I even understand how `use` is supposed to work in why3 correctly, you should probably check that you agree with me before merging this.

 I'm suspicious that I might be missing similar issues by doing

```rust
// Filter out enum variants, and maybe other things...
if !matches!(util::item_type(ctx.tcx, variant.def_id), ItemType::Unsupported(_))
```

 I haven't yet gone out of my way to try to find (or convince myself of the lack of) those similar issues. If you remove the check, a bunch of tests break with "unsupported definition kind" - but there might be a more principled version of the check available too.
 
 This PR is arguably a bit premature given the above, but I'm not going to have time to work on it for the next few days, so I thought I'd push it anyways.

### DIFF
--- a/creusot/src/translation/function/place.rs
+++ b/creusot/src/translation/function/place.rs
@@ -70,7 +70,7 @@ pub(crate) fn create_assign_inner<'tcx>(
                 }
             }
             Field(ix, _) => match place_ty.ty.kind() {
-                TyKind::Adt(def, _) => {
+                TyKind::Adt(def, subst) => {
                     let variant_id = place_ty.variant_index.unwrap_or_else(|| 0u32.into());
                     let variant = &def.variants()[variant_id];
                     let var_size = variant.fields.len();
@@ -86,6 +86,7 @@ pub(crate) fn create_assign_inner<'tcx>(
 
                     let tyname = constructor_qname(ctx, variant);
 
+                    names.insert(def.did(), subst);
                     inner = Let {
                         pattern: ConsP(tyname.clone(), field_pats),
                         arg: box translate_rplace_inner(ctx, names, body, lhs.local, stump),
@@ -180,10 +181,11 @@ pub(crate) fn translate_rplace_inner<'tcx>(
                 }
             }
             Field(ix, _) => match place_ty.ty.kind() {
-                TyKind::Adt(def, _) => {
+                TyKind::Adt(def, subst) => {
                     let variant_id = place_ty.variant_index.unwrap_or_else(|| 0u32.into());
                     let variant = &def.variants()[variant_id];
 
+                    names.insert(def.did(), subst);
                     ctx.translate_accessor(def.variants()[variant_id].fields[ix.as_usize()].did);
                     let accessor_name =
                         variant_accessor_name(ctx, def.did(), variant, ix.as_usize());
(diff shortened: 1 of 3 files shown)


creusot/src/translation/function/place.rs
```diff
@@ -180,11 +180,18 @@ pub fn translate_rplace_inner<'tcx>(
...
                     let variant_id = place_ty.variant_index.unwrap_or_else(|| 0u32.into());
                     let variant = &def.variants()[variant_id];
 
-                    ctx.translate_accessor(def.variants()[variant_id].fields[ix.as_usize()].did);
+                    let field_id = variant.fields[ix.as_usize()].did;
+                    // Filter out enum variants, and maybe other things...
+                    if !matches!(util::item_type(ctx.tcx, variant.def_id), ItemType::Unsupported(_))
+                    {
+                        names.insert(ctx.tcx.parent(field_id), subst);
+                    }
```
xldenis, line 193: this can just be replaced by `names.insert(def.did(), subst)`, which will correctly add the edge for the datatype itself.
```suggestion
                    
                    names.insert(def.did(), subst);
                    
```

gmorenz:
Finally got back to this :)

Fixed the suggestion in the comment, also fixed another case of the same issue, and added a more complete test.

I haven't checked if #584 will completely obsolete this? Even if it does, it might be nice to get these expressions in the test suite so they don't break in the future.

creusot/tests/should_succeed/bug/570.stderr
```diff
@@ -0,0 +1,34 @@
...
+  |
+4 | struct S1 {
+  |        ^^
+  |
+  = note: `#[warn(dead_code)]` on by default
+
+warning: struct `S2` is never constructed
+ --> 570.rs:8:8
+  |
+8 | struct S2 {
```
xldenis, line 12: Can you fix these warnings by making things public?

xldenis (approved):
One last detail. 

As for #584, I wouldn't wait on  it. I'm going to land individual bits over the coming weeks, but I doubt #584 itself  will be merged

creusot/tests/should_succeed/bug/570.stderr
gmorenz: Done

PR #576, opened 2022-09-20
Fix missing impl refinements

xldenis:
Major bug in #573 (predictable given the title). This fixes it.


xldenis:
The solution would be to not expose internal methods like `translate_trait` publicly.

PR #579, opened 2022-09-21
Implement Iterator::fuse

xldenis:
Adds `Fuse<I>`, `fuse` and `FusedIterator`. These are useful in general when working with iterators because reasoning about non-fused iterators is extremely difficult in general (see `Map`).


xldenis:
Not here yet :)

xldenis:
why is it saying that `07_fuse` is `obsolete`???

creusot/tests/should_succeed/iterators/07_fuse.stderr
```diff
@@ -0,0 +1,23 @@
...
+  --> 07_fuse.rs:29:9
+   |
+29 |         (pearlite! { (exists<x :_> (^self).iter == Err(x)) })
+   |         ^^^^^^^^^^^^^                                     ^^^
+   |
+   = note: `#[warn(unused_parens)]` on by default
+help: remove these parentheses
+   |
+29 -         (pearlite! { (exists<x :_> (^self).iter == Err(x)) })
+29 +         (exists<x :_> (^self).iter == Err(x))
```
xldenis, line 19: So close, yet so wrong!

xldenis:
`FusedIterators` are useful when attempting to generically `map` an iterator as the required 're-initialization' property is hard to satisfy in general for arbitrary iterators.

PR #582, opened 2022-09-22
Fix default trait methods being transparent

xldenis:


xldenis:
Fixes #222 again, ensuring that we never clone the body of a logical method which could be further refined by specialization or default impls.

PR #593, opened 2022-10-03
Add support for non-associated constants

gmorenz:
This feels a bit too easy, but it seems like it just works? I've had this change locally for awhile now, and haven't run into any issues.

I wasn't sure if the test should go under `lang` or `syntax`?

### DIFF
--- a/creusot/src/util.rs
+++ b/creusot/src/util.rs
@@ -287,7 +287,7 @@ pub(crate) fn item_type(tcx: TyCtxt<'_>, def_id: DefId) -> ItemType {
                 ItemType::Program
             }
         }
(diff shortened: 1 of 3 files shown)


xldenis:
Seems about right! I actually think const fn parameters should also be pretty easy but that requires touching `CloneMap`.

As for the test:

I really need to reorganize and cleanup the test suite. My idea was that `lang` tested support for rust features while `syntax` (needs renaming) was about the Creusot specific "extensions" to rust.

xldenis:
More generally I'd like to go over all the tests, eliminate useless or trivial ones, organize the rest and document what they test. It can be hard to tell *why* a test matters today.

xldenis (approved):

Issue #594, opened 2022-10-03
Patterns in signatures


dewert99:
The following program
```rust
#[logic]
fn test((x, y): (u32, u32)) -> u32 {
    x
}

#[ensures(result == x)]
fn test2((x, y): (u32, u32)) -> u32 {
    x
}
```
produces the following why3 file
```
module CreusotExperiments_Fail_Test_Stub
  use mach.int.Int
  use mach.int.UInt32
  function test (_1' : (uint32, uint32)) : uint32
end
module CreusotExperiments_Fail_Test_Interface
  use mach.int.Int
  use mach.int.UInt32
  function test (_1' : (uint32, uint32)) : uint32
end
module CreusotExperiments_Fail_Test
  use mach.int.Int
  use mach.int.UInt32
  function test [#"../src\fail.rs" 4 0 4 34] (_1' : (uint32, uint32)) : uint32 =
    [#"../src\fail.rs" 5 4 5 5] x
  val test (_1' : (uint32, uint32)) : uint32
    ensures { result = test _1' }
    
end
module CreusotContracts_Logic_Resolve_Resolve_Resolve_Stub
  type self
  predicate resolve (self : self)
end
module CreusotContracts_Logic_Resolve_Resolve_Resolve_Interface
  type self
  predicate resolve (self : self)
end
module CreusotContracts_Logic_Resolve_Resolve_Resolve
  type self
  predicate resolve (self : self)
  val resolve (self : self) : bool
    ensures { result = resolve self }
    
end
module CreusotContracts_Logic_Resolve_Impl0_Resolve_Stub
  type t1
  type t2
  predicate resolve (self : (t1, t2))
end
module CreusotContracts_Logic_Resolve_Impl0_Resolve_Interface
  type t1
  type t2
  predicate resolve (self : (t1, t2))
end
module CreusotContracts_Logic_Resolve_Impl0_Resolve
  type t1
  type t2
  clone CreusotContracts_Logic_Resolve_Resolve_Resolve_Stub as Resolve1 with type self = t2
  clone CreusotContracts_Logic_Resolve_Resolve_Resolve_Stub as Resolve0 with type self = t1
  predicate resolve (self : (t1, t2)) =
    Resolve0.resolve (let (a, _) = self in a) /\ Resolve1.resolve (let (_, a) = self in a)
  val resolve (self : (t1, t2)) : bool
    ensures { result = resolve self }
    
end
[...]

PR #599, opened 2022-10-04
Updated toolchain to nightly-2022-10-04

dewert99:
Updates the toolchain to help with #594.  This seems to have modified the behaviour of `def_span`.
In the following test:
```rust
use creusot_contracts::*;

#[logic]
#[ensures(true && false)]
#[creusot::builtins = "dummy_function"]
fn builtin_with_contract() {}
```
it now gives a span from line for leading to the error:
```
error[creusot]: cannot specify both `creusot::builtins` and a contract on the same definition
 --> builtin_with_contract.rs:4:1
  |
4 | #[logic]
  | ^^^^^^^^
  |
  = note: this error originates in the attribute macro `logic` (in Nightly builds, run with -Z macro-backtrace for more info)

error: aborting due to previous error
```
instead of a span from line 7 leading to the error:
```
warning: function `builtin_with_contract` is never used
 --> builtin_with_contract.rs:7:4
  |
7 | fn builtin_with_contract() {}
  |    ^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(dead_code)]` on by default

error[creusot]: cannot specify both `creusot::builtins` and a contract on the same definition
 --> builtin_with_contract.rs:7:1
  |
7 | fn builtin_with_contract() {}
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to previous error; 1 warning emitted
```

### DIFF
--- a/creusot/src/translation/constant.rs
+++ b/creusot/src/translation/constant.rs
@@ -8,7 +8,10 @@ use crate::{
 use creusot_rustc::{
     hir::def_id::DefId,
     middle::{
-        mir::interpret::{AllocRange, ConstValue},
+        mir::{
+            interpret::{AllocRange, ConstValue},
+            UnevaluatedConst,
+        },
         ty::{Const, ConstKind, ParamEnv, Ty, TyCtxt},
     },
     smir::mir::ConstantKind,
@@ -68,6 +71,10 @@ pub(crate) fn from_mir_constant_kind<'tcx>(
         }
     }
 
+    if let ConstantKind::Unevaluated(UnevaluatedConst { promoted: Some(p), .. }, _) = ck {
+        return Expr::Exp(Exp::impure_var(format!("promoted{:?}", p.as_usize()).into()));
+    }
+
     return Expr::Constant(try_to_bits(ctx, names, env, ck.ty(), span, ck));
 }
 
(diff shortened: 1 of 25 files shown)


creusot/src/translation/constant.rs
```diff
@@ -84,9 +87,9 @@ pub(crate) fn from_ty_const<'tcx>(
             return Expr::Exp(Exp::pure_qvar(nm.without_search_path()));
     };
 
-    if let ConstKind::Unevaluated(Unevaluated { promoted: Some(p), .. }) = c.kind() {
-        return Expr::Exp(Exp::impure_var(format!("promoted{:?}", p.as_usize()).into()));
-    }
+    // if let ConstKind::Unevaluated(UnevaluatedConst{ promoted: Some(p), .. }) = c.kind() {
+    //     return Expr::Exp(Exp::impure_var(format!("promoted{:?}", p.as_usize()).into()));
+    // }
```
xldenis, line 92: Interesting, any idea where promoted consts have moved to then?

creusot/src/translation/function/statement.rs
```diff
@@ -28,9 +28,9 @@ impl<'tcx> BodyTranslator<'_, 'tcx> {
             StorageDead(_) | StorageLive(_) | Nop => {}
             // Not real instructions
             FakeRead(_) | AscribeUserType(_, _) | Retag(_, _) | Coverage(_) => {}
-            CopyNonOverlapping(_) => self.ctx.crash_and_error(
+            Intrinsic(_) => self.ctx.crash_and_error(
```
xldenis, line 31: Ooo, I wonder what operations are possible here.

xldenis (commented):
I'll take a look at that issue, there's no reason the spans should have changed, they're provided using stable proc macro apis..

xldenis:
I did a quick bisection: using the `09-28-2022` nightly makes this issue disappear. I think it might be a consequence of https://github.com/rust-lang/rust/pull/102538

xldenis:
Ok, I'm 90% sure the issue is due to that PR, but the fix is simple: use `quote_spanned!` in `creusot-contracts-proc`.

xldenis:
I've taken the liberty of adding a commit fixing this issue. It uncovers more issues as the `should_succeed` suite then runs.

dewert99:
Thanks for getting this working

PR #606, opened 2022-10-05
Logical Reborrow / Unnesting

xldenis:
Adds logical reborrowing / unnesting, so that we can project out mutable borrows to fields of mutable borrows in pearlite. This enables significantly more ergonomic specifications when dealing with mutable borrows.

There remains a question on which behavior to give unnesting: `&mut ** x` should it be:
- ~~`(**x, *^x)`~~
- `(**x, ^*x)`

The first is what happens in programs and is good for coherence between Pearlite and Rust, the second is perhaps more 'intuitive', you're effectively getting the borrow that's just one `*` deep. 
I'm personally leaning towards the first option, I like keeping the coherence between programs and specs.

--- 

After much agonizing I've come to accept the conclusion that the second choice is the only viable choice at least in Pearlite as it is. The short of it is that we should be stable under substitution and the second option isn't.

### DIFF
--- a/creusot/src/backend/term.rs
+++ b/creusot/src/backend/term.rs
@@ -316,6 +316,12 @@ impl<'tcx> Lower<'_, 'tcx> {
                 Exp::Abs(binders, box body)
             }
             TermKind::Absurd => Exp::Absurd,
+            TermKind::Reborrow { cur, fin } => Exp::Record {
+                fields: vec![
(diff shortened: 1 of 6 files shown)


xldenis:
The current version also does not support `ref mut` patterns in `match`, as I have no idea how those work or what they mean. It would be nice to eventually support them so that we could express `as_ref` / `as_mut_ref` in Pearlite.

Issue #605, opened 2022-10-05
Let binding pattern in ensures clause causes Creusot to panic


dewert99:
```rust
use creusot_contracts::*;

struct S(u32);

#[ensures({let S(y) = x; result == y})]
fn test(x: S) -> u32 {
    x.0
}
```
causes panic
```
thread 'rustc' panicked at 'no key for DefId(0:6 ~ creusot_experiments[ce4b]::fail::S)', creusot\src\ctx.rs:323:64
```

I found this while working on #594, but it seems like it would need an independent fix

xldenis:
`pattern_term` needs to trigger `ctx.translate` for the type `DefId` for the `Variant` and `Leaf` cases.

xldenis:
Fixed.

PR #607, opened 2022-10-05
Fixed #594

### LINKED ISSUE #594: Patterns in signatures

dewert99:
Fixes #594.

Note issue #605 leads to issue when pattern matching with a struct and needs to be fixed seperatly


creusot/src/translation/pearlite.rs
```diff
@@ -164,6 +164,35 @@ struct ThirTerm<'a, 'tcx> {
 // TODO: Ensure that types are correct during this translation, in particular
 // - Box, & and &mut
 impl<'a, 'tcx> ThirTerm<'a, 'tcx> {
+    fn body_term(&self, expr: ExprId) -> CreusotResult<Term<'tcx>> {
+        let body = self.expr_term(expr)?;
+        let owner_id = util::closure_owner(self.tcx, self.item_id.into());
```
xldenis, line 169: I'm a worried this won't play well with Pearlite closures, where we wouldn't want to look at the closure owner for params but the closure itself.

xldenis: Additionnally, what about when we are looking at the contracts of a program closure?

creusot/src/util.rs
```diff
@@ -678,3 +682,11 @@ pub(crate) fn closure_capture_subst<'tcx>(
     let TyKind::Closure(_, substs) = tcx.type_of(def_id).kind() else { unreachable!() };
     ClosureSubst { self_, map: subst, post: is_post, def_id, bound: Default::default(), substs }
 }
+
+pub(crate) struct AnonymousParamName(pub(crate) usize);
+
+impl Display for AnonymousParamName {
```
xldenis, line 688: :+1:

creusot/src/translation/pearlite.rs
xldenis: Perhaps, we could make `ensures` and `requires` duplicate the parent's signature in the generated closure so we don't need to look at the parents to extract the params?

dewert99: My first thought would be to get the parent if the def_id had the attribute `#[creusot::decl::spec]`, but looking at the macro expansion it seems like the specs on closures are siblings of the closure, not children.  Given an arbitrary spec_closure, it also looks like there is no good way to tell if it specifies the outer function or a different closure so I would know whether or not to use the parent.

dewert99: I'd prefer not to touch creusot_contracts_proc, but that doesn't seem possible.  I'm trying to decide if it would be better to duplicate the arguments as you suggested or go in the other direction and push spec_closures into the closures they specify and remove their arguments as well.  If we did that would it be possible to do away with IDs (`#[creusot::item]`) since all specifications would be directly contained in the thing they specify?

dewert99: I noticed you were using `dummy_call` and `closure_result` to help with type inference, if the spec_closure no longer repeats the arguments `dummy_call` won't be needed.  We can also modify the closure being specified to call `pub fn closure_result<R, F: FnOnce(R) -> bool>(_: F, _: R)` with it's ensures clause and return value instead of doing it in the spec closure.  Is there any other reason why this approach wouldn't work?

xldenis: >  go in the other direction and push spec_closures into the closures they specify and remove their arguments as well

I think the reason I did it that way is that we break type inference otherwise. We want to let real closure guide inference as much as possible. A possibly a little crazy idea I just had was: push the spec_closures in the program closure but put it at the *end* of the program closure, so hopefully type inference infers the arg types before getting to the specification.

dewert99: I is sort of what I was thinking for the `return` arg.  I was wondering if we could do away with the other args all together like we do with top level specs.

dewert99:
I tried modifying `creusot_contracts` so that specs on closures a specified within the closure, and everything in the test suite seems to type check, but the mlcfg output is different (which makes sense since things are encoded differently), do you think you could check this over?

xldenis:
> I tried modifying creusot_contracts so that specs on closures a specified within the closure, and everything in the test suite seems to type check, but the mlcfg output is different (which makes sense since things are encoded differently), do you think you could check this over?

I'll do that now.

xldenis:
It all looks good, a manual read of the MLCFG shows to me that only shuffling of instructions and temporary variables occured. Moreover, the only foreseeable side-effect of a bug would be that contracts are dropped (leading to verification errors) or inference is impacted (leading to failing to typecheck).

xldenis:
Go ahead and update the MLCFG files & the few affected sessions.

dewert99:
Now that I switched the closure spec nesting I was able to use the strategy to use the parent `defid` for `#[creusot::decl::spec]` and the original otherwise. I think this should work now so do you think you could take care of the merge conflicts?

xldenis:
Sure, otherwise I believe you can just overwrite them by running bless on the conflicting tests.

xldenis:
Thank you @dewert99!

PR #613, opened 2022-10-07
Cleanup std

### LINKED ISSUE #523: Vec::eq does not satisfy it's postcondition
`PartialEq::eq` currently has the postcondition `result == (@self == @rhs)` and `Vec::<T>::ModelTy` is `Seq<T>` which implies that `Vec::eq` should return true iff both vectors have exactly the same elements.  `Vec::eq` is implemented by calling `PartialEq::eq`  on all its elements so it returns true iff both vectors' elements have the same models. This can lead to panics in verified programs a shown below:
```
use creusot_contracts::*;
struct UnitModel(bool);

impl Model for UnitModel {
    type ModelTy = ();

    #[logic]
    fn model(self) -> Self::ModelTy {
        ()
    }
}

impl PartialEq for UnitModel {
    #[ensures(result == (@self == @other))]
    fn eq(&self, other: &Self) -> bool {
        true
    }
}

#[cfg_attr(not(feature = "contracts"), test)]
fn test() {
    let mut v1 = Vec::new();
    v1.push(UnitModel(true));
    let mut v2 = Vec::new();
    v2.push(UnitModel(false));
    if v1 == v2 {
        assert!(v1[0].0 == v2[0].0)
    }
}
```
There are a few possible ways to fix:
  * Change `PartialEq::eq`'s postcondition to `result == (self == rhs)`
    - This would make it challenging to have structs with various not identical versions that can often be treated as equivalent (eg. vectors with different capacities, btrees that are balanced differently, hashtables with elements in a different order)
  * Change `Vec::<T>::ModelTy` to `Seq<T::ModelTy>`
    - This would mean that only who's elements implement `Model` could be used (All methods are currently specified in terms of models)
    - We would lose information (eg. pushing and popping gives back an identical element not just one with the same model)
  * Change `Vec::<T>::ModelTy` to `Seq<T::ModelTy>` and intoduce a new logical (returning Seq<T> method to use instead of model)
    - The best way to do this would be to implement `Vec::model()` as `self.new_logic_method().map(|e| e.model())` but this requires logical closures
    - This could make specification for vectors more verbose

Note this also affects slices

jhjourdan:
Various cleanups to the standard library (see commit messages).

Fixes #523.


creusot-contracts/src/model.rs
```diff
@@ -0,0 +1,142 @@
...
+    #[logic]
+    fn shallow_model(self) -> Self::ShallowModelTy {
+        (*self).shallow_model()
+    }
+}
+
+impl<T, const N: usize> ShallowModel for [T; N] {
+    type ShallowModelTy = Seq<T>;
+
+    // We define this as trusted because builtins and ensures are incompatible
```
xldenis, line 82: Though we can't put the contract in Creusot, because builtins just use a why3 function, you can put the contract in why3 and it should work.

creusot-contracts/src/std/clone.rs
```diff
@@ -5,6 +5,10 @@ extern_spec! {
     mod std {
         mod clone {
             trait Clone {
+                // TODO:
+                // Requiring result == *self seems too strong since objects can
+                // contain information which is lost by the clone (e.g., capacity of
+                // vectors...).
```
xldenis, line 11: Is that the case?

creusot-contracts/src/std/default.rs
```diff
@@ -1,94 +1,121 @@
...
+        mod default {
+            trait Default where Self : Default {
+                #[ensures(result.is_default())]
+                fn default() -> Self;
+            }
+        }
+    }
+}
+
+// TODO: use a macro for these instances.
```
xldenis, line 21: What causes the group to appear, do you know?

xldenis: Can we have the macro generate a different type of group by inserting braces or parens?

creusot-contracts/src/std/ops.rs
```diff
@@ -76,15 +84,21 @@ impl<Args, F: Fn<Args>> FnSpec<Args> for F {
...
+    #[trusted]
+    #[ensures(self.postcondition_once(args, res) == self.resolve() && self.postcondition(args, res))]
+    fn fn_once(self, args: Args, res: Self::Output) {}
 }
 
 extern_spec! {
     mod std {
         mod ops {
+            // FIXME: we should not need the `Self :` bounds, because they are implied by the main trait bound.
+            // But if we remove them, some test fail.
```
xldenis, line 101: Is that still the case since #586 ?

creusot-contracts/src/std/default.rs
jhjourdan: I guess that as soon as you use a meta-variable in a macro, it inserts a group. Here, we would like to use a meta-varaible for the default value, and it needs to appear in the pearlite code.

creusot-contracts/src/std/clone.rs
jhjourdan: It's really not clear to me.

If we want things like the capacity of vectors to be available in the specs, then the spec of Clone breaks. If we want the addresses of Boxes, same thing (this might be needed for interior mutability).

creusot-contracts/src/std/ops.rs
jhjourdan: Yes.

creusot-contracts/src/std/clone.rs
xldenis: What do we do then? Compare the models?

xldenis: Seems like a reasonable compromise.

creusot-contracts/src/std/default.rs
xldenis: How do the other macros that generate ord instances get around this then?

creusot-contracts/src/model.rs
```diff
@@ -0,0 +1,135 @@
...
+
+impl<T: ShallowModel + ?Sized> ShallowModel for &mut T {
+    type ShallowModelTy = T::ShallowModelTy;
+    #[logic]
+    fn shallow_model(self) -> Self::ShallowModelTy {
+        (*self).shallow_model()
+    }
+}
+
+#[cfg(feature = "contracts")]
```
xldenis, line 61: I understand why this is only on the instances that require `Allocator` but perhaps we should just put it on all or none?

PR #626, opened 2022-10-13
Allow creusot_contracts::std to work with rustc

xldenis:


creusot-contracts/src/lib.rs
```diff
@@ -143,12 +144,9 @@ pub mod __stubs;
 
 pub mod logic;
 
-#[cfg(feature = "contracts")]
+#[allow(unused)]
```
jhjourdan, line 147: Can you only enable this if the feature is disabled? Otherwise we miss some relevant warnings.

xldenis: 👍

xldenis: it would be cool if long term even in passthrough mode the macros could somehow properly simulate a use of the imported names

PR #640, opened 2022-10-21
More iterators

jhjourdan:
@xldenis : there is a bug in extern specs which makes impossible to add a post-condition to `into_iter` (cf FIXME in iter.rs).


xldenis:
We are asking to resolve `Invariant` for `Self::IntoIter` in the following context:

```
ParamEnv { caller_bounds: [
Binder(TraitPredicate(<<I as std::iter::IntoIterator>::IntoIter as creusot_contracts::Iterator>, polarity:Positive), []), 
Binder(TraitPredicate(<<I as std::iter::IntoIterator>::IntoIter as creusot_contracts::invariant::Invariant>, polarity:Positive), []),
Binder(TraitPredicate(<I as creusot_contracts::IntoIterator>, polarity:Positive), []), 
Binder(TraitPredicate(<<I as std::iter::IntoIterator>::IntoIter as creusot_contracts::Iterator>, polarity:Positive), []), 
Binder(TraitPredicate(<<I as std::iter::IntoIterator>::IntoIter as creusot_contracts::invariant::Invariant>, polarity:Positive), []), 
Binder(TraitPredicate(<I as creusot_contracts::IntoIterator>, polarity:Positive), []), 
Binder(~const TraitPredicate(<I as std::iter::Iterator>, polarity:Positive), []), 
Binder(TraitPredicate(<I as std::marker::Sized>, polarity:Positive), [])], reveal: UserFacing, constness: NotConst }
```

xldenis:
the following code fails to typecheck:

```rust
#[predicate]
pub fn test_bounds<I>(i : I, res: I::IntoIter)
where <I as ::std::iter::IntoIterator>::IntoIter : Iterator + invariant::Invariant,

{
    res.invariant()
}
```

I feel like this is a MVCE for the error coming above... I can't figure out why that code would fail to typecheck

jhjourdan:
Note that there is still a bug somewhere, since removing the constraint `+Invariant` triggers a panic.

xldenis:
I wonder if `ParamEnv` needs to be fully normalized, (which would make sense), that is `T : Ord` is broken down into `T : PartialOrd, T : Ord, T : PartialEq, T : Eq`

PR #643, opened 2022-10-24
Normalize bounds while building param_env using extern specs

xldenis:
@jhjourdan this should solve the issue you encountered with `IntoIterator`. 

I don't think it's fully correct as it only uses the pre-existing `ParamEnv` for an item to normalize the additional bounds, but I don't know how we could do better...

### DIFF
--- a/creusot/src/clone_map.rs
+++ b/creusot/src/clone_map.rs
@@ -290,7 +290,7 @@ impl<'tcx> CloneMap<'tcx> {
 
             let base = Symbol::intern(&base_sym.as_str().to_camel_case());
             let count: usize = *self.name_counts.entry(base).and_modify(|c| *c += 1).or_insert(0);
-            debug!("inserting {:?} {:?} as {}{}", def_id, subst, base, count);
+            trace!("inserting {:?} {:?} as {}{}", def_id, subst, base, count);
 
             let info =
                 CloneInfo::from_name(Symbol::intern(&format!("{}{}", base, count)), self.public);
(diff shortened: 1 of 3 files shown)


jhjourdan:
Thanks.

Could you document in which case this might not be correct and leave a FIXME in the code about that?

Also, you are using `try_normalize_erasing_regions`. Are you expecting it to fail sometimes?

xldenis:
> Could you document in which case this might not be correct and leave a FIXME in the code about that?

Sure, my suspicion is that if we add bounds in an extern spec which are mutually dependent for normalization (can we even do this??) then it would break. 

> Also, you are using try_normalize_erasing_regions. Are you expecting it to fail sometimes?

No, I just have gotten in the habit of using that over the infallible `normalize_erasing_regions`.

jhjourdan:
> Sure, my suspicion is that if we add bounds in an extern spec which are mutually dependent for normalization (can we even do this??) then it would break.

In the following example:

```
trait T1<A> {
    type B;
}
trait T2<A> {
    type B;
}

fn f<U, V>(u: U, v: V)
    where U: T1<V::B>, V: T2<U::B>
{

}
```

Rust answers:

```
error[E0391]: cycle detected when computing the bounds for type parameter `V`
 --> test.rs:9:30
  |
9 |     where U: T1<V::B>, V: T2<U::B>
  |                              ^^^^
  |
note: ...which requires computing the bounds for type parameter `U`...
 --> test.rs:9:17
  |
9 |     where U: T1<V::B>, V: T2<U::B>
  |                 ^^^^
  = note: ...which again requires computing the bounds for type parameter `V`, completing the cycle
note: cycle used when computing explicit predicates of `f`
 --> test.rs:9:17
  |
9 |     where U: T1<V::B>, V: T2<U::B>
  |                 ^^^^

error: aborting due to previous error

For more information about this error, try `rustc --explain E0391`.
```

So it *seems* like there is some mechanism that prevents cycles and the case you are describing. It's not clear, though, that this is the same notion of mutual recursivity that you need here

BUUUT... in any case, I don't see how your code is handling additional constraints whose normalization would depend on another one non-recursively. Some kind of topological sorting is necessary, I think, and you need to update param_env incrementally somehow.

jhjourdan:
> No, I just have gotten in the habit of using that over the infallible normalize_erasing_regions.

Should'nt we use the faillible version to detect bugs?

jhjourdan:
What about `normalize_param_env_or_error`?

xldenis:
> Should'nt we use the faillible version to detect bugs?

The infaillibe one throws an error when it can't normalize, i call it infallible because the type doesn't let you react to the error.

xldenis:
> in any case, I don't see how your code is handling additional constraints whose normalization would depend on another one non-recursively.

Indeed, my code only allows a simple 'staging' of constraints.

PR #645, opened 2022-10-25
Add spec for Instant and refine spec for Duration

jhaye:
This PR does:
* implement `OrdLogic` for `Option<T>`
* tighten the upper bounds for `Duration` for accuracy
* define spec for simple addition and subtraction for `Duration`
  * updates the test accordingly
* add spec and test for `Instant`

While `Instant` is an opaque type, I've chosen the model type to be `Int` because it allows for easy relative reasoning between `Instant`s after doing arithmetic.

### DIFF
--- a/creusot-contracts/src/logic/ord.rs
+++ b/creusot-contracts/src/logic/ord.rs
@@ -232,3 +232,42 @@ impl<A: OrdLogic, B: OrdLogic> OrdLogic for (A, B) {
     #[logic]
     fn eq_cmp(_: Self, _: Self) {}
 }
+
+impl<T: OrdLogic> OrdLogic for Option<T> {
+    #[logic]
+    fn cmp_log(self, o: Self) -> Ordering {
+        match (self, o) {
+            (None, None) => Ordering::Equal,
+            (None, Some(_)) => Ordering::Less,
+            (Some(_), None) => Ordering::Greater,
+            (Some(x), Some(y)) => x.cmp_log(y),
+        }
+    }
+
+    #[logic]
+    fn cmp_le_log(_: Self, _: Self) {}
+
+    #[logic]
+    fn cmp_lt_log(_: Self, _: Self) {}
+
+    #[logic]
+    fn cmp_ge_log(_: Self, _: Self) {}
+
+    #[logic]
+    fn cmp_gt_log(_: Self, _: Self) {}
+
+    #[logic]
+    fn refl(_: Self) {}
+
+    #[logic]
+    fn trans(_: Self, _: Self, _: Self, _: Ordering) {}
+
+    #[logic]
+    fn antisym1(_: Self, _: Self) {}
+
+    #[logic]
+    fn antisym2(_: Self, _: Self) {}
+
+    #[logic]
+    fn eq_cmp(_: Self, _: Self) {}
+}
--- a/creusot-contracts/src/logic/ord.rs
+++ b/creusot-contracts/src/logic/ord.rs
@@ -244,30 +244,43 @@ impl<T: OrdLogic> OrdLogic for Option<T> {
         }
     }
 
-    #[logic]
-    fn cmp_le_log(_: Self, _: Self) {}
+    #[law]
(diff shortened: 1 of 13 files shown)


creusot-contracts/src/std/time.rs
```diff
@@ -1,12 +1,15 @@
...
+    time::*,
+};
 
 impl ShallowModel for Duration {
     type ShallowModelTy = Int;
 
     #[logic]
     #[trusted]
-    #[ensures(result >= 0 && result <= @u128::MAX)]
+    #[ensures(result >= 0 && result <= secs_to_nanos(@u64::MAX) + 999_999_999)]
```
xldenis, line 12: I kind of wonder if it wouldn't make sense to have a wrapper around `Int` that provides these methods... idk.

jhaye: Something like `struct DurationNano(Int)`?

xldenis: Yea exactly, idk if it's worth the hassle

jhaye: Yeah maybe a bit overkill, but I could factor the constant out into a function (a constant wouldn't work I assume)?

creusot-contracts/src/logic/ord.rs
```diff
@@ -232,3 +232,42 @@ impl<A: OrdLogic, B: OrdLogic> OrdLogic for (A, B) {
...
+    fn trans(_: Self, _: Self, _: Self, _: Ordering) {}
+
+    #[logic]
+    fn antisym1(_: Self, _: Self) {}
+
+    #[logic]
+    fn antisym2(_: Self, _: Self) {}
+
+    #[logic]
+    fn eq_cmp(_: Self, _: Self) {}
```
jhjourdan, line 272: These are `#[law]`s, and should duplicate the specifications (I know, this is a bit heavy). The reason why this does not trigger an error is that we don't run Why3 over

creusot-contracts/src/std/time.rs
```diff
@@ -107,6 +134,66 @@ extern_spec! {
...
+                #[ensures(@self <= @earlier ==> @result == 0)]
+                fn saturating_duration_since(&self, earlier: Instant) -> Duration;
+
+                #[ensures(@duration == 0 ==> result.deep_model() == Some(@self))]
+                #[ensures(@duration > 0 && result != None ==> Some(@self) < result.deep_model())]
+                fn checked_add(&self, duration: Duration) -> Option<Instant>;
+
+                #[ensures(@duration == 0 ==> result.deep_model() == Some(@self))]
+                #[ensures(@duration > 0 && result != None ==> Some(@self) > result.deep_model())]
+                fn checked_sub(&self, duration: Duration) -> Option<Instant>;
```
jhjourdan, line 163: These specifications seems very weak. Is this intended?

creusot-contracts/src/logic/ord.rs
jhaye: I see, does this also apply to the functions that I omitted since they were already specified in the `OrdLogic` trait?

jhjourdan: Nope: default implementation are like in Rust: they don't need to be repeated.

creusot-contracts/src/std/time.rs
jhaye: Instant is an opaque type with OS-specific behaviour which makes reasoning about its internal state close to impossible. But we can at least make some relative claims with these operations and that's what this specification captures.

creusot-contracts/src/logic/ord.rs
jhaye: 51ea82a37ddce74a4f84c3466e53911cf58f8aea should address this.

creusot-contracts/src/std/time.rs
xldenis: Just leave it for now, it's fine like this. A constant should work, we just don't support const functions or const generics

PR #650, opened 2022-11-10
Add spec for VecDeque

jhaye:
I added specs for the functions that I thought would be most useful for `VecDeque`. Most of them mirror the specs for `Vec`. Original specs are for `pop_front()` and `push_front()`.


xldenis:
@jhaye please rebase and this will merge

jhaye:
Rebase disables auto-merge unfortunately.

PR #658, opened 2022-11-30
Remove CloneMap

xldenis:
12030th time is the charm. 

Removes `CloneMap` in favor of pre-calculating all the dependencies in the crate. 
This also makes it simpler to not produce clones but instead monomorphize all the definitions.

The code is incomplete and not cleaned up, at the moment it can translate _basic_ logic functions and _basic_ program functions. I'll keep progressively upping the complexity (promoteds, closures, etc) until it covers everything before doing a major clean up pass. 

Major TODOs:
- [x] Closures
- [x] Constants
- [ ] External Dependencies
- [ ] Debug
- [ ] Add caching for dependency calculation
- [ ] Remove `CloneMap`


xldenis:
I have now managed to successfully compile a module with a basic closure in it. 
At this point the main work is going to be ironing out edge cases, simplifying the code and then progressively cleaning up the prior code.

PR #663, opened 2022-12-05
Implement DeepModel and OrdLogic for Reverse

jhaye:
Not sure if a test is necessary here.

### DIFF
--- a/creusot-contracts/src/std/cmp.rs
+++ b/creusot-contracts/src/std/cmp.rs
@@ -54,3 +54,63 @@ extern_spec! {
         }
     }
 }
+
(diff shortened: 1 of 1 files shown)


xldenis:
Once CI passes LGTM

jhaye:
It seems like the timeout of 1s isn't enough to fully replay `iterators/12_zip`. I'm not sure how to fix that.

jhjourdan:
Try another prover? Try to increase the timeout?

xldenis:
> In the CI I meant. Oh well, rerunning fixed it 🤷

Yea there is occasional flakiness, cloud smt etc.. I would say that so long as it's <1% no need to do anything at our current scale.

PR #669, opened 2022-12-31
Add capacity modifying methods to Vec

georgerennie:
This adds some basic models for methods that change the capacity of a Vec. As creusot doesn't consider the capacity of the vector in the model (as is already evident from `Vec::with_capacity`), these implementations are pretty trivial.

If there is a mistake with these or if it doesn't make sense to add them please let me know. In particular I was interested in adding support for `Vec::clear`, but thought I would add a few more while I was at it.


xldenis:
> If there is a mistake with these or if it doesn't make sense to add them please let me know. In particular I was interested in adding support for Vec::clear, but thought I would add a few more while I was at it.

The specs are pretty weak, but I don't see any real harm in adding them. In the future the model of `Vec` could always be refined to account for capacity if this turns out to be an important concept in proofs.

xldenis:
I went ahead and updated the sessions since they were obsolete for whatever reason.

xldenis:
Thanks for the contribution @georgerennie

PR #675, opened 2023-01-16
Introduce IR for trait refinement

xldenis:
More migrations.. introduces IR for trait refinement. The change *seems* correct, but avoids a few clones from before so I am little skeptical.


xldenis:
Turns out I was leaving out the refinement for program methods, which I've added back in. I was right to suspect an issue!

PR #700, opened 2023-02-28
Added default to vec.rs

cmester0:


xldenis:
Thanks, that seems good to me!

cmester0:
Seems `creusot/should_fail/bug/603.rs` does not fail any more, as there is actually an implementation for `Default` Should the file/test be removed, or is this something that needs to fail?

xldenis:
It's actually still failing, just with a slightly different error message. You can go ahead and update the test by running `cargo test --test ui "603" -- --bless`

PR #701, opened 2023-03-03
Simplify resolution & Fix #699

### LINKED ISSUE #699: Can't prove the resolve of an FnOnce in a branch
From discussion #698, I would expect this code to be provable:

```#![cfg_attr(not(feature = "contracts"), feature(stmt_expr_attributes, proc_macro_hygiene))]
use creusot_contracts::*;

#[requires(run.precondition(()))]
#[ensures(
    (b && run.postcondition_once((),()))
        || (!b && run.resolve()))]
fn test<F:FnOnce() -> ()>(b: bool, run: F) {
    if b { run() }; 
}

#[ensures((n > 7u32 && result == 2u32) || (n <= 7u32 && result == 0u32))]
fn example(n: u32) -> u32 {
    let mut r = 0u32;
    let run =
        #[ensures (r == 2u32)]
        || { r = 2 };
    test(n > 7, run);
    r
}
    
fn main(){
    example(8);
}
```
But, currently (e.g. from 0ffd39ad), the postcondition of `test` cannot be proved.

xldenis:
This PR simplifies the resolution relation by removing unnecessary information & changes the notion of liveness used in calculation.

Instead of using traditional liveness, we use a relation which ignores `drop` from consideration. 
Note that this is unsound if we have logically non-trivial drops, but today Creusot ignores `drop` terminators already (as we don't have a satisfying specification / understanding of them). 

To illustrate why we want this change, consider the following small program:

```rust

fn conditional<T>(b: bool, x: T) {
  if b { 
    f(x);
  }
  // implicit drop(x) here 
}
```

Today, the relation which determines which variables to resolve at a given location in the code is "variables which just died and are initialised". 
In this case this is a problem because since we use MIR pre-drop elaboration, which means that at the `drop` for `x`, we are in a state where we *might* be initialised and are *live* and thus we get no `resolve`. 
Since Creusot already ignores `drop`, the most direct solution to this is to just no longer consider drop in the calculation of liveless.
This allows us to insert a `resolve` statement for `x` in the (empty) else branch of the if-expression, and correctly let `f` consume `x` (potentially providing a `resolve` in its postcondition).  

r @jhjourdan, I would like you to take a look at this if you could.


creusot/src/resolve.rs
```diff
@@ -125,12 +77,12 @@ impl<'body, 'tcx> EagerResolver<'body, 'tcx> {
...
         }
 
-        live_at_start.union(&self.unactivated_borrows(two_phase_start).to_hybrid());
+        // live_at_start.union(&self.unactivated_borrows(two_phase_start).to_hybrid());
 
         end.seek_to(&mut self.local_live);
         let mut live_at_end: BitSet<_> = BitSet::new_empty(self.local_live.get().domain_size());
         live_at_end.union(self.local_live.get());
-        live_at_end.union(&self.unactivated_borrows(two_phase_end).to_hybrid());
+        // live_at_end.union(&self.unactivated_borrows(two_phase_end).to_hybrid());
```
jhjourdan, line 85: Idem.

creusot/src/resolve.rs
```diff
@@ -43,55 +34,23 @@ impl<'body, 'tcx> EagerResolver<'body, 'tcx> {
...
-        // TODO: Defensive check.
-        let local_live =
-            MaybeLiveLocals.into_engine(tcx, body).iterate_to_fixpoint().into_results_cursor(body);
-        let never_live = crate::analysis::NeverLive::for_body(body);
-        EagerResolver { local_live, local_init, local_uninit, never_live, body, borrows }
-    }
+        let _ = MaybeBorrowedLocals
+            .into_engine(tcx, body)
+            .iterate_to_fixpoint()
+            .into_results_cursor(body);
```
jhjourdan, line 40: Is this useful?

PR #705, opened 2023-03-08
Add cloned, copied and enumerate iterators

voidc:


creusot/tests/should_succeed/iterators/15_enumerate.rs
```diff
@@ -0,0 +1,83 @@
...
+        pearlite! { self.iter.completed() }
+    }
+
+    #[predicate]
+    fn produces(self, visited: Seq<Self::Item>, o: Self) -> bool {
+        pearlite! {
+            visited.len() == @o.count - @self.count
+            && exists<s: Seq<I::Item>> self.iter.produces(s, o.iter)
+                && visited.len() == s.len()
+                && forall<i: Int> 0 <= i && i < s.len() ==> @visited[i].0 == @self.count + i && visited[i].1 == s[i]
```
xldenis, line 41: We really need a `IteratorLogic` trait to generalize over these quantification patterns.

xldenis: Especially if like with program iterators we can guarantee that it desugars properly to just quantifiers at the why3 level.

creusot/tests/should_succeed/iterators/15_enumerate.rs
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+
+    #[predicate]
+    fn invariant(self) -> bool {
+        pearlite! {
+            self.invariant_aux()
+            && forall<reset : &mut Enumerate<I>> reset.completed() ==> (^reset).invariant_aux()
```
xldenis, line 49: I'm not sure that this is provable. `invariant_aux` requires `(^reset).invariant()` but I don't think you can prove that here.

xldenis: I think you'll really need to actually build an `enumerate` iterator to see whether this invariant is actually satisfiable.

xldenis:
I look a like at specifying `enumerate`, the challenge is as always dealing with _unfused_ iterators. 
I propose a simpler, more limited specification: require that if we have `self.completed` then that means we have an empty step from `self` to `^self`.

In short, change the invariant to:

```rust
#[predicate]
fn invariant(self) -> bool {
    pearlite! {
        self.iter.invariant() &&
        (forall<s: Seq<I::Item>, i: I> i.invariant() ==> self.iter.produces(s, i) ==> @self.count + s.len() < @usize::MAX) &&
        (forall<i: &mut I> i.invariant() ==> i.completed() ==> i.produces(Seq::EMPTY, ^i))
    }
}
```

and use these preconditions:

```rust
#[requires(forall<i: &mut I> i.invariant() ==> i.completed() ==> i.produces(Seq::EMPTY, ^i))]
#[requires(forall<s: Seq<I::Item>, i: I> i.invariant() ==> iter.produces(s, i) ==> s.len() < @usize::MAX)]
```

PR #706, opened 2023-03-08
Check that logic functions are not called from impure code

voidc:
This extends `PurityVistor` to check that logic functions are not called from impure code.


creusot/src/translation/specification.rs
```diff
@@ -330,22 +341,20 @@ impl<'a, 'tcx> thir::visit::Visitor<'a, 'tcx> for PurityVisitor<'a, 'tcx> {
...
-                        && !is_overloaded_item(self.tcx, func_did)
-                    {
+                    if self.in_pure_ctx != self.is_pure(fun, func_did) {
+                        let msg = if self.in_pure_ctx {
+                            "called impure program function in logical context"
+                        } else {
+                            "called logical function in impure context"
+                        };
+
                         self.tcx.sess.span_err_with_code(
```
xldenis, line 351: I think we could change `PurityVisitor` to take a `ctx` instead and use `ctx.error` to build the error.

creusot/src/translation/pearlite.rs
```diff
@@ -202,7 +202,7 @@ pub(crate) fn pearlite(tcx: TyCtxt, id: LocalDefId) -> CreusotResult<Term> {
         return Err(Error::new(tcx.def_span(id), "type checking failed"));
     };
 
-    visit::walk_expr(&mut PurityVisitor { tcx, thir: &thir }, &thir[expr]);
+    visit::walk_expr(&mut PurityVisitor { tcx, thir: &thir, in_pure_ctx: true }, &thir[expr]);
```
xldenis, line 205: Maybe we can move this check out of here and just do it at the same time for every function?

PR #710, opened 2023-03-11
Add `IndexLogic` trait

xldenis:
I'm declaring war on `@`.


jhjourdan:
I'm not quite sure I'm in favor. This is one more step towards giving a different semantics between pearlite and Rust, which is something we would really like to avoid...

xldenis:
> I'm not quite sure I'm in favor. This is one more step towards giving a different semantics between pearlite and Rust, which is something we would really like to avoid...

Well in that case we need to make `Vec::index` pure and allow specifying things directly in terms of sequences... My issue is our specifications are fairly illegible because of the constant presence of `Model`.

creusot-contracts/src/std/vec.rs
```diff
@@ -74,15 +74,15 @@ extern_spec! {
                 fn pop(&mut self) -> Option<T>;
 
                 #[requires(@ix < (@self).len())]
-                #[ensures(result == (@self)[@ix])]
+                #[ensures(result == self[@ix])]
```
xldenis, line 77: @jhjourdan my objective is fixing issues like this, and here we can also remove the `@` in front of the `ix`.

jhjourdan:
My objections are still relevant...

jhjourdan:
Another question: why don't we have instances of IndexLogic for borrows for things that are indexable ?

(I.e., how does logical reborrowing works with this?)

jhjourdan:
The point is that it should be true for every type that implement Index/IndexMut

xldenis:
> Another question: why don't we have instances of IndexLogic for borrows for things that are indexable ?
> 
> 
> 
> (I.e., how does logical reborrowing works with this?)

Right now all this does is provide a separate trait to index `Seq` and also provide the logical instance for vectors. 

Adding an instance for `&mut [T]` would be cool but that would introduce a potential divergence point from rust.

PR #715, opened 2023-03-14
Relax `still_specializable`

xldenis:
Slightly relaxes the condition on `still_specializable` so that we can see the bodies of default implementations which happened to be used in _final_ trait impls. This addresses the specific case where we have:

```
trait X { predicate foo() { true } } 

impl X for Bar<T> { } 
```

since the `impl` of `X` for `Bar` is not specializable, we should be allowed to open the definition of `foo` in methods where we are using `Bar<T>::foo()`.


jhjourdan:
If we merge this, are we exposed to evolutions of Rust that would allow more specializations.

xldenis:
> If we merge this, are we exposed to evolutions of Rust that would allow more specializations?

No, this uses the specialization query apis to make sure the definition is `final`. if we bump rust version in manner that would change the result then we will stop specializing as well.

jhjourdan:
Alright, but this means that some proofs will break if this notion of "final" is changed.

xldenis:
> Alright, but this means that some proofs will break if this notion of "final" is changed.

Yes, but if the language changes what it means for something to be specializable, that makes sense that we would (correctly) break affected proofs.

PR #716, opened 2023-03-15
Check if session is proved in why3 test

voidc:


xldenis:
if you can fix the ui tests this lgtm

PR #717, opened 2023-03-16
Add basic support for type invariants

voidc:


xldenis (commented):
some quick comments. I'll take a deeper look later

PR #723, opened 2023-03-20
Remove explicit impl Invariant for Once

voidc:
The explicit implementation is no longer needed thanks to #715.


xldenis:
Note: in the future we may also need / want an invariant on `Once` due to the inner `T` it holds (if only for consistency reasons).

xldenis:
Wait... shouldn't we be required to prove that the result upholds the invariant and thus we need a custom invariant on once?

xldenis:
I think this might be because `Option` doesn't yet have an invariant? Once it does I expect that we will have to give `Once` a real invariant.

voidc:
I think it makes sense to defer adding an invariant on Once until we have a blanket implementation of Invariant.

jhjourdan:
Can we merge this?

PR #725, opened 2023-03-20
Stop using `features = contracts` to toggle creusot.

xldenis:
This stops using `features=contracts` to toggle creusot code. I'm not sure how I can warn about this though...


andreykl:
Am I right that now it should be `$ cargo creusot` instead of `$ cargo creusot -- features=contracts`?

andreykl:
works like a charm!

PR #728, opened 2023-03-22
Translate Rvalue::Repeat and provide a custom vec macro

voidc:


creusot/tests/should_succeed/syntax/13_vec_macro.rs
```diff
@@ -0,0 +1,14 @@
+extern crate creusot_contracts;
+
+use creusot_contracts::{vec, *};
```
voidc, line 3: Sadly, rustc won't let us glob import the vec macro (as it shadows the builtin macro)

creusot-contracts/src/std/slice.rs
```diff
@@ -274,6 +277,9 @@ extern_spec! {
             forall<j:usize> i <= j && @j < (@self).len() ==> x.deep_model() < self.deep_model()[@j])]
         fn binary_search(&self, x : &T) -> Result<usize, usize>
             where T: Ord + DeepModel,  T::DeepModelTy: OrdLogic,;
+
+        #[ensures(@result == @self_)]
+        fn into_vec<A: Allocator>(self_: Box<Self, A>) -> Vec<T, A>;
```
xldenis, line 282: I think when we update to `syn 2.0` we'll be able to fix this glitch in the macros that forces us to write `self_`. previous versions of `syn` did not allow custom receiver types.

xldenis:
lgtm! merge when ci passes.

xldenis:
It seems like whenever we touch some specific part of `creusot-contracts` the rbt test breaks.. that sort of narrows things but not a ton..

PR #729, opened 2023-03-24
Smarter constructors for axioms

xldenis:
This makes our spec axiom generation smarter by exploiting that `A -> T <-> T`. Thus, if we have no post condition to a logical function we don't need to generate an axiom at all. 
Additionally, the smart constructor for `Exp::implies` was upgraded to leverage that same identity.

This clears a bunch of useless axioms from our context. @voidc, this might also help with some redundant circularities with type invariants.


xldenis:
I really hope that smt solvers weren't stupid enough to fall for these extra axioms but in any case, less quantifiers can only be good.

PR #731, opened 2023-03-24
 Simplify proofs using ADT support

jhjourdan:
E.g., stop using expression templates in RBT example.


xldenis:
We have a couple specs in `Creusot-contracts` that avoid matches, do you think we can refactor those now

jhjourdan:
This needs to be tried....

That being said, I don't know what's the status of ADT support for Altergo in Why3.

PR #736, opened 2023-04-01
Encode float operators with a dot

nilehmann:
I noticed some of the float operators were missing while working on [this](https://github.com/xldenis/creusot/discussions/735). I got the syntax for operators from here https://why3.lri.fr/stdlib/ieee_float.html. 

I added a test, but not sure if that's how it should be tested. I tried it locally and it worked.

### DIFF
--- a/creusot/src/backend/term.rs
+++ b/creusot/src/backend/term.rs
@@ -1,5 +1,5 @@
 use crate::{
-    backend::ty::{intty_to_ty, translate_ty, uintty_to_ty},
+    backend::ty::{floatty_to_ty, intty_to_ty, translate_ty, uintty_to_ty},
     ctx::*,
     pearlite::{self, Literal, Pattern, Term, TermKind},
     util,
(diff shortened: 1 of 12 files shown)


creusot/src/backend/term.rs
```diff
@@ -392,7 +392,10 @@ pub(crate) fn lower_literal<'tcx>(
             names.insert(id, subst);
             Exp::Tuple(Vec::new())
         }
-        Literal::Float(f) => Constant::Float(f).into(),
+        Literal::Float(f, fty) => {
+            let _why_ty = floatty_to_ty(names, &fty);
```
nilehmann, line 396: I needed this to import the module. Not sure if should be propagated down in the `Constant::Float`.

why3/src/exp.rs
```diff
@@ -50,11 +54,15 @@ impl BinOp {
             BinOp::Ne => Infix1,
             BinOp::Ge => Infix1,
             BinOp::Gt => Infix1,
-            BinOp::FloatAdd => Infix4,
-            BinOp::FloatSub => Infix4,
-            BinOp::FloatMul => Infix4,
-            BinOp::FloatDiv => Infix4,
+            BinOp::FloatAdd => Infix5,
```
nilehmann, line 57: I added a new precedence level to separate arithmetic from comparison for float operators. That felt sensible to me.

xldenis: Hmm the why manual lists the precedence levels of operators and how they should spread out throughout them. According to that all float operators should be Infix4

xldenis:
Hi, thank you for your contribution to Creusot! I just have one minor comment about precedence of the new operators but otherwise this looks good to me.

nilehmann:
Oh, I thought they had two levels. Removed it!

PR #738, opened 2023-04-03
Add a "rust-analyzer.rustc.source" setting for vscode's rust-analyzer.

krtab:
This setting enables rust-analyzer to discover rustc private crates. Cf https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/.E2.9C.94.20How.20to.20get.20rust-analyzer.20in.20vscode.20for.20OOT.20rustc.20plugins

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "rust-analyzer.rustc.source": "discover",
+}


xldenis:
I have this set as a global setting in vscode so I didn't consider adding it to Creusot, thanks for rectifying that!

PR #740, opened 2023-04-13
Point to ci/rust-toolchain in README.md

bclement-ocp:
The README directs users to use the same rust-toolchain file as Creusot itself, but that file in the Creusot repo is a symlink to ci/rust-toolchain which is not followed by the GitHub UI.

This patch links directly to ci/rust-toolchain whose content users can copy.


xldenis:
:+1: 

I think we can also probably just get rid of the symlink, I'm not sure we really need it anymore.

PR #745, opened 2023-04-18
Add postfix model operator

voidc:
Allows referring to the model of a value using the syntax `foo.@`.
For now, this still accepts the old prefix syntax to permit a gradual migration.

### DIFF
--- a/creusot-contracts-proc/src/extern_spec.rs
+++ b/creusot-contracts-proc/src/extern_spec.rs
@@ -358,7 +358,9 @@ fn escape_self_in_term(t: &mut Term) {
             escape_self_in_term(expr);
         }
         Term::Final(TermFinal { term, .. }) => escape_self_in_term(term),
-        Term::Model(TermModel { term, .. }) => escape_self_in_term(term),
+        Term::Model(TermModel { term, .. }) | Term::ModelPost(TermModelPost { term, .. }) => {
+            escape_self_in_term(term)
+        }
         Term::LogEq(TermLogEq { lhs, rhs, .. }) => {
             escape_self_in_term(lhs);
             escape_self_in_term(rhs)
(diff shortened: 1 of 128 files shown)


xldenis:
> Allows referring to the model of a value using the syntax `foo.@`.

Ah interesting, instead of `ident@` do `ident.@` to mimic `.async`. I like it!

xldenis:
Do you have some examples of how it improves specs (just for curiousity).

sarsko:
I prefer `x@` to `x.@`. I also think mixfix is super confusing (and would suggest not having a gradual migration).

I must admit that my immediate reaction was "No I hate this", but that may just be Stockholm syndrome after having written way too many parentheses. I'll have to try it to see how I like this. I assume that I will like it better, as it is less characters and also makes specs readable left to right without having to backtrack whenever we get a model.

xldenis:
Yea I agree with @sarsko and @jhjourdan  that we should not have mixfix and ideally do a one-shot migration (it can be over a few commits / prs but fast). 

also for completeness sake:

```rust
#[invariant(cuts_els, cuts@[0]@ == 0 && cuts@[cuts@.len()-1]@ == x@)]
#[invariant(cuts_bounds, forall<i: Int> 0 <= i && i < cuts@.len() ==> 0 <= cuts@[I]@ && cuts@[I]@ <= s@.len())]
#[invariant(cuts_mono, forall<i: Int> 0 < i && i < cuts@.len() ==> mono(s@, cuts@[i-1]@, cuts@[I]@))]
```

And if we had #710 

```rust
#[invariant(cuts_els, cuts[0]@ == 0 && cuts[cuts@.len()-1]@ == x@)]
#[invariant(cuts_bounds, forall<i: Int> 0 <= i && i < cuts@.len() ==> 0 <= cuts[I]@ && cuts[I]@ <= s@.len())]
#[invariant(cuts_mono, forall<i: Int> 0 < i && i < cuts@.len() ==> mono(s@, cuts[i-1]@, cuts[I]@))]
```

xldenis:
🎉

PR #756, opened 2023-04-27
Implemented GhostPtrToken

dewert99:


creusot-contracts/src/ghost_ptr.rs
```diff
@@ -0,0 +1,165 @@
...
+        }
+    }
+}
+
+// Coverts `val` into a [`GhostPtr`] and gives `t` permission to it
+// Safety this pointer was owned by a box so no other GhostToken could have permission to it
+#[trusted]
+#[ensures(!(*t)@.contains(result))]
+#[ensures((^t)@ == (*t)@.insert(result, *val))]
+pub fn ptr_from_box<T>(val: Box<T>, t: &mut GhostPtrToken<T>) -> *const T {
```
xldenis, line 106: I kind of feel like we should make these methods on the `GhostPtrToken` type, what do you think? We could also clarify the naming.

creusot-contracts/src/ghost_ptr.rs
```diff
@@ -0,0 +1,165 @@
...
+#[requires(t@.contains(ptr))]
+#[ensures(*result == t@.lookup(ptr))]
+pub fn ptr_as_ref<T>(ptr: *const T, t: &GhostPtrToken<T>) -> &T {
+    unsafe { &*ptr }
+}
+
+#[trusted]
+#[requires(new_model.subset((*t)@))]
+#[ensures((*result)@ == *new_model)]
+pub fn shrink_token_ref<T>(
```
xldenis, line 123: I don't quite understand how you would use this safely, can you explain? Shouldn't it be impossible to have two `GhostPtrToken` that are subsets?

creusot-contracts/src/ghost_ptr.rs
```diff
@@ -0,0 +1,165 @@
...
+// Safety no other token has permission to `self`
+// `t` can no longer be used to access `ptr`
+#[trusted]
+#[requires((**t)@.contains(ptr))]
+#[ensures(*result == (**t)@.lookup(ptr))]
+#[ensures((*^t)@ == (**t)@.remove(ptr))]
+#[ensures((^*t)@.remove(ptr) == (^^t)@)]
+#[ensures((^*t)@.get(ptr) == Some((^result).make_sized()))]
+#[ensures(!(^^t)@.contains(ptr))]
+pub fn ptr_as_mut2<'o, 'i, T>(ptr: *const T, t: &'o mut &'i mut GhostPtrToken<T>) -> &'i mut T {
```
xldenis, line 140: We have to come up with a better name. Something like `remove_mut_ptr`?

creusot-contracts/src/ghost_ptr.rs
```diff
@@ -0,0 +1,165 @@
...
+    new_model: Ghost<<GhostPtrToken<T> as ShallowModel>::ShallowModelTy>,
+) -> &GhostPtrToken<T> {
+    t
+}
+
+/// Mutably borrows `ptr` and shrinks `t` so that it can no longer be used to access `ptr`
+// Safety no other token has permission to `self`
+// `t` can no longer be used to access `ptr`
+#[trusted]
+#[requires((**t)@.contains(ptr))]
```
xldenis, line 134: I'm not a huge fan of having this mix of `**` `^*` `*^` and `^^`. I wonder if we can rephrase things to be more legible.

creusot-contracts/src/logic/fmap.rs
```diff
@@ -0,0 +1,124 @@
+use crate::{logic::Mapping, util::*, *};
+
+pub struct FMap<K, V: ?Sized>(pub Mapping<K, Option<SizedW<V>>>);
```
xldenis, line 3: As written, I think this is unsound, at the very least we should need the type to be opaque since otherwise it seems like there is a bug lurking akin to the `&[T]` and `Seq<T>` bug I discovered last week?

xldenis (commented):
I think this will be a terrific addition! 

I have a few questions about the soundness of the `GhostPtrToken` and `FMap` because it seems like they might need type invariants. 

Additionally, I think we can perhaps rename some of the `GhostPtrToken` methods to give them more descriptive names.

creusot-contracts/src/ghost_ptr.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    /// Leaks memory iff the precondition fails
+    #[requires(self@.is_empty())]
+    pub fn drop(self) {}
+}
+
+impl<T: ?Sized> GhostPtrExt<T> for GhostPtr<T> {
+    #[trusted]
+    #[logic]
+    #[ensures(forall<t: GhostPtrToken<T>> !t@.contains(result))]
```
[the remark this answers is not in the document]
dewert99: I now made `GhostPtrToken` opaque so this shouldn't apply to all `FMap`s which should hopefully make this sound. One nice thing about defining it this way is the lemma will only apply when using null.

creusot-contracts/src/ghost_ptr.rs
dewert99: I added a brief description, does it seem ok?

creusot-contracts/src/ghost_ptr.rs
dewert99: I'm not sure how that would work, I need to refer to the 4 parts of the `&mut &mut GhostPtrToken`

creusot-contracts/src/util.rs
```diff
@@ -0,0 +1,41 @@
+use crate::*;
+
+pub type SizedW<T> = Box<T>;
```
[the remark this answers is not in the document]
dewert99: "Size Wrapper".  It's necessary since `Option<T>` requires `T` to be sized, I'm using a box since Creusot erases them, and they don't have lifetimes.

creusot-contracts/src/ghost_ptr.rs
dewert99: I made these into methods, let if know if you have good suggestions for the names

creusot-contracts/src/ghost_ptr.rs
dewert99: I changed the name to `take_mut` (after https://doc.rust-lang.org/std/primitive.slice.html#method.take_mut), are you ok with that?

creusot-contracts/src/ghost_ptr.rs
xldenis: Yea, these kinds of code are just tricky, the comment makes the intent relatively clear

creusot-contracts/src/ghost_ptr.rs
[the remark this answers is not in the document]
xldenis: The alternative would be to define an `Invariant` instance, but this should work.

xldenis:
Just bless the RBT case (and bdd apparently), and I think we can merge this. The Red-Black Trees somehow have a diff when we modify `creusot-contracts` (need to investigate).

creusot-contracts/src/ghost_ptr.rs
```diff
@@ -0,0 +1,171 @@
...
+        absurd
+    }
+}
+
+pub trait GhostPtrExt<T: ?Sized>: Sized {
+    #[logic]
+    fn null_logic() -> Self;
+    #[logic]
+    fn addr_logic(self) -> Int;
+}
```
xldenis, line 145: Why do you use an extension trait here? Why not directly implement the methods on `GhostPtr`?

xldenis (commented):
two final comments and we can merge.

creusot-contracts/src/ghost_ptr.rs
dewert99: Right now `GhostPtr<T>` is just a type alias for `* const T`

creusot-contracts/src/util.rs
```diff
@@ -0,0 +1,41 @@
...
+#[requires(false)]
+#[ensures(false)]
+#[variant(0)]
+pub fn unreachable<T>() -> T {
+    unreachable()
+}
+
+#[logic]
+#[requires(b)]
+pub fn assert(b: bool) {}
```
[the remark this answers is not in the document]
dewert99: I think I was using it to debug logic functions.  Can `proof_assert!` be used in a logic function?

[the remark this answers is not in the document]
xldenis: Not yet but that's a bug. Can we remove this from the contracts? And we'll get assert working properly instead.

creusot-contracts/src/ghost_ptr.rs
xldenis: Of course, I forgot.

xldenis:
Thank you David!

PR #757, opened 2023-04-28
Lasso list reversal

jhjourdan:
- New example: lasso list reversal
- Upgrade to syn 2.0.15
- Fix the limitation where the pearlite proc macro could not be used in a declarative macro
- A bit of cleaning in the stdlib
- Get rid of invariant names


creusot-contracts/src/logic/seq.rs
```diff
@@ -120,18 +127,13 @@ impl<T> Seq<T> {
...
-    type Output = T;
+impl<T> IndexLogic<Int> for Seq<T> {
+    type Item = T;
 
-    #[trusted]
     #[logic]
+    #[trusted]
     #[creusot::builtins = "seq.Seq.get"]
-    fn index(&self, _: Int) -> &T {
+    fn index_logic(self, _: Int) -> Self::Item {
```
xldenis, line 136: This makes me happy

creusot/src/gather_spec_closures.rs
```diff
@@ -12,11 +12,10 @@ use rustc_middle::{
     mir::{visit::Visitor, AggregateKind, BasicBlock, Body, Location, Operand, Rvalue},
     ty::{TyCtxt, TyKind},
 };
-use rustc_span::Symbol;
 
 #[derive(Debug, Clone, Copy)]
 pub enum LoopSpecKind {
-    Invariant(Symbol),
```
xldenis, line 19: So I had left named invariants to give people time to upgrade their why3, I figured we would start by also deprecating it in Creusot.

Cargo.lock
```diff
@@ -550,7 +550,7 @@ dependencies = [
  "proc-macro-error-attr",
  "proc-macro2",
  "quote",
- "syn",
+ "syn 1.0.109",
```
xldenis, line 553: Hrmm, we should figure out which crate is inducing this dependency and get everything fully upgraded

xldenis: Ah it's `pearlite-syn`

creusot/src/gather_spec_closures.rs
jhjourdan: Isn't it already backward compatible wrt. why3 ? I thought Why3 supported unnamed invariants in the past.

Since when do we deprecate things in Creusot? That would really be the first time!

Anyways, it's really not a difficult change to do on the code. Simply remove invariant names.

Cargo.lock
jhjourdan: Well... If our dependencies use a different version of Syn, is this a problem?

creusot/src/gather_spec_closures.rs
xldenis: Mlcfg invariants needed to be named until yesterday, and only with the master branch of why3. The reason I've noticed is that we have external users will do an opam install why3 and it won't work. Another thing I was thinking is that cargo Creusot could try to detect the version of why3 and issue a warning if it's too old.

PR #759, opened 2023-05-02
WIP: Opaque predicates

xldenis:
Turns out opacity is pretty much already implemented?


xldenis:
The opacity system works, however it (obviously) breaks all of our proofs. This has led me to make a few remarks:

- We have (one) example of a `pub(crate)` trait which defines predicates which appear on public functions that we want to be transparent. This would violate the rule that transparency is a subset of visibility. While this pattern could be used to have 'sealed logic traits', I think we can just use one of the other techniques to seal the trait if needed.
- There are a few _private_ logic functions which (transitively) appear in public contracts with the same consequence. This is a little trickier to solve since it does seem like we could legitimately want to define a logic function but let it be fully transparent?
- The default policy of `open(self)` (akin to `pub(self)` ) is perhaps a bit annoying, remains to be seen, I've disabled then check for now. (TODO: Re-enable).
- We define builtins for why3 types in a two step process: first the raw why3 symbol is added as a private function and then a public function wraps it with the correct type signature. We **need** this public function to be transparent but it refers to a private symbol. The solution is probably to just not directly use why3 stdlib modules (a good idea for parametricity reasons anyways) and define our own `prelude.{FSet, Seq, ...}` modules with the correct signatures from the start.

xldenis:
oh wtf i never merged this?

PR #764, opened 2023-05-03
Add FSet::ext_eq for extensionality

bclement-ocp:
And bless red_black_trees.


xldenis:
seems like RBT needs to be blessed again? This bug still baffles me... I would very much like to know where / why it happens.

bclement-ocp:
again? I thought I blessed it already…

bclement-ocp:
Oh but I ran `rustfmt` after blessing :black_joker:

PR #765, opened 2023-05-03
Refactor handling of MIR bodies

voidc:
This enables translating const fns.

### DIFF
--- a/creusot/src/callbacks.rs
+++ b/creusot/src/callbacks.rs
@@ -95,6 +95,9 @@ impl Callbacks for ToWhy {
     }
 }
 
+/// Trys to retrieve the promoted MIR for a body from a thread local cache.
+/// The cache is populated when rustc runs the `mir_borrowck` query.
+/// After a body was retrieved, calling this function again for the same `def_id` will return `None`.
 pub fn get_body<'tcx>(tcx: TyCtxt<'tcx>, def_id: LocalDefId) -> Option<BodyAndPromoteds<'tcx>> {
     // trigger borrow checking
     let _ = tcx.mir_borrowck(def_id);
--- a/creusot/src/callbacks.rs
+++ b/creusot/src/callbacks.rs
@@ -1,5 +1,13 @@
 use rustc_driver::{Callbacks, Compilation};
+use rustc_hir::def_id::LocalDefId;
+use rustc_index::vec::IndexVec;
 use rustc_interface::{interface::Compiler, Config, Queries};
+use rustc_middle::{
+    mir::{Body, Promoted},
+    ty::{self, TyCtxt},
+};
+
+use std::{cell::RefCell, collections::HashMap, thread_local};
 
 use crate::{cleanup_spec_closures::*, options::Options};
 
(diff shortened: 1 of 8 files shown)


creusot/src/callbacks.rs
```diff
@@ -23,6 +43,27 @@ impl Callbacks for ToWhy {
...
+                // We use `mir_promoted` as it is the MIR required by borrowck
+                let (body, promoted) = tcx.mir_promoted(ty::WithOptConstParam::unknown(def_id));
+                let body_and_proms = BodyAndPromoteds {
+                    body: body.borrow().clone(),
+                    promoted: promoted.borrow().clone(),
+                };
+
+                // SAFETY: The reader casts the 'static lifetime to 'tcx before using it.
+                let body_and_proms: BodyAndPromoteds<'static> =
+                    unsafe { std::mem::transmute(body_and_proms) };
```
xldenis, line 57: uh. is there no better way to do this? (I can't think of any off the top of my head)

creusot/src/ctx.rs
```diff
@@ -22,26 +23,44 @@ use rustc_errors::{DiagnosticBuilder, DiagnosticId};
...
 use why3::exp::Exp;
 
 pub(crate) use crate::translated_item::*;
 
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+pub struct BodyId {
+    def_id: LocalDefId,
+    pub promoted: Option<Promoted>,
+}
+
```
xldenis, line 46: FINALLY, GOOD, THANKS. Incidentally, I've seen discussions about giving promoteds their own `DefId`s but this is good.

creusot/src/callbacks.rs
```diff
@@ -53,3 +94,16 @@ impl Callbacks for ToWhy {
         }
     }
 }
+
+pub fn get_body<'tcx>(tcx: TyCtxt<'tcx>, def_id: LocalDefId) -> Option<BodyAndPromoteds<'tcx>> {
```
xldenis, line 98: I think this is worth a small function level comment, explaining that it fetches from a thread local catch and will panic and will return `None` if called a second time.

creusot/src/translation/function/promoted.rs
```diff
@@ -24,13 +23,13 @@ pub(crate) fn promoted_signature<'tcx>(body: &Body<'tcx>) -> PreSignature<'tcx>
 
 pub(crate) fn translate_promoted<'tcx>(
     ctx: &mut TranslationCtx<'tcx>,
-    body: &Body<'tcx>,
-    parent: DefId,
+    body_id: BodyId,
 ) -> CreusotResult<(PreSignature<'tcx>, fmir::Body<'tcx>)> {
-    let func_translator = BodyTranslator::build_context(ctx.tcx, ctx, &body, parent);
+    let body = ctx.body(body_id).clone();
```
xldenis, line 28: I wish we didn't have to clone the body here but if that turns out to be a perf issue later, we can always either wrap it in an `Rc` or use interior mutability in `ctx` like rustc does.

xldenis:
lgtm

creusot/src/callbacks.rs
voidc: it's kind of stupid but I don't see any other way since storage in the thread local requires `'static`

PR #774, opened 2023-05-05
WIP: Add support for partial orders

bclement-ocp:
I do not know yet if the proofs still pass, as this seems to be breaking trait resolution somehow.  There are many instances (e.g. 08_haystack) where I get errors about Creusot's inability to find an instance for PartialOrdLogic but everything that used to have an OrdLogic instance should still have one, and OrdLogic implies PartialOrdLogic -- so I'm not sure what's going on there...

### DIFF
--- a/creusot-contracts/src/logic/ord.rs
+++ b/creusot-contracts/src/logic/ord.rs
@@ -1,157 +1,144 @@
-use crate::{std::cmp::Ordering, *};
+use crate::{std::cmp::Ordering, util::*, *};
 
 #[allow(unused)]
-pub trait OrdLogic {
+pub trait PartialOrdLogic {
+    #[predicate]
+    fn lt_log(self, _: Self) -> bool;
+
+    #[law]
+    #[ensures(!self.lt_log(self))]
+    fn lt_log_irrefl(self);
+
+    #[law]
+    #[requires(x.lt_log(y))]
+    #[ensures(!y.lt_log(x))]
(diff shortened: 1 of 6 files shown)


creusot-contracts/src/logic/ord.rs
```diff
@@ -1,157 +1,144 @@
...
+    #[ensures(result == (self.lt_log(other) || self == other))]
+    fn le_log(self, other: Self) -> bool {
+        self.lt_log(other) || self == other
     }
 
     #[law]
-    #[ensures(x.gt_log(y) == (x.cmp_log(y) == Ordering::Greater))]
-    fn cmp_gt_log(x: Self, y: Self);
+    #[ensures(self.le_log(self))]
+    fn le_log_refl(self) {}
```
jhjourdan, line 37: Partial orders in Rust are not reflexive because of floats.

jhjourdan:
We are sorry not to have handled this during one year an a half. Is this still something you want to add? What is your use case?

If you rebase and the issue still exists, I'll be happy to look at it.

bclement-ocp:
The use case was some code that was working on sets equipped with inclusion. IIRC I got around the issue by defining my own `PartialOrder` type. I can look at rebasing the PR but I'll probably not have the bandwidth before a couple of weeks or so.

jhjourdan:
Subsumed by #2174 .

PR #775, opened 2023-05-11
Fix `Span` encoding

xldenis:
The previous encoding of `Span` was totally wrong, I attempted to fix it here, however, it turns out this is trickier as the 'real' implementation in rustc uses non-public apis. 
I came up with a best-effort approximation.

### DIFF
--- a/creusot/tests/should_fail/bug/01_resolve_unsoundness.mlcfg
+++ b/creusot/tests/should_fail/bug/01_resolve_unsoundness.mlcfg
@@ -63,7 +63,7 @@ module CreusotContracts_Std1_Vec_Impl0_ShallowModel_Interface
   use Alloc_Vec_Vec_Type as Alloc_Vec_Vec_Type
   clone Core_Num_Impl11_Max_Stub as Max0
   function shallow_model (self : Alloc_Vec_Vec_Type.t_vec t a) : Seq.seq t
-  axiom shallow_model_spec : forall self : Alloc_Vec_Vec_Type.t_vec t a . Seq.length (shallow_model self) <= UIntSize.to_int Max0.mAX'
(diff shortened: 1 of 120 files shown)


creusot/tests/should_succeed/red_black_tree.mlcfg
```diff
@@ -420,7 +420,7 @@ module CreusotContracts_Logic_Ord_OrdLogic_LtLog
   clone CreusotContracts_Logic_Ord_OrdLogic_CmpLog_Stub as CmpLog0 with
     type self = self
   predicate lt_log (self : self) (o : self) =
-    [#"../red_black_tree.rs" 618 9 619 5] CmpLog0.cmp_log self o = Core_Cmp_Ordering_Type.C_Less
+    [#"../../../../creusot-contracts/src/logic/ord.rs" 19 20 19 53] CmpLog0.cmp_log self o = Core_Cmp_Ordering_Type.C_Less
```
xldenis, line 423: The problem spans are fixed!

creusot/tests/should_fail/bug/01_resolve_unsoundness.mlcfg
```diff
@@ -131,7 +131,7 @@ module CreusotContracts_Logic_Ops_Impl0_IndexLogic
     type self = s,
     type ShallowModelTy0.shallowModelTy = Seq.seq t
   function index_logic [@inline:trivial] (self : s) (ix : int) : t =
-    Seq.get (ShallowModel0.shallow_model self) ix
+    [#"../../../../../creusot-contracts/src/logic/ops.rs" 16 8 16 31] Seq.get (ShallowModel0.shallow_model self) ix
```
xldenis, line 134: A bunch of logic things are gaining their (appropriate) spans, since before they were just invalid offsets. Now that we encode spans more properly we are able to map them back onto their source files!

jhjourdan:
(Arg, I was currently working at this as well... but I am trying to handle the SyntaxContext correctly...).

xldenis:
I'm more worried that adding more spans to things will cause Why3 to perform even worse due to `split_vc`...

PR #778, opened 2023-05-12
Simplify resolution logic

voidc:


creusot/tests/should_fail/bug/specialize.mlcfg
```diff
@@ -277,6 +276,7 @@ module Specialize_F
     goto BB2
   }
   BB2 {
+    assume { Resolve0.resolve v_1 };
```
xldenis, line 279: Why are we resolving `v_1` here? it should be unitialized?

creusot/tests/should_succeed/swap_borrows.mlcfg
```diff
@@ -68,20 +68,20 @@ module SwapBorrows_Swap
...
-    assume { Resolve0.resolve _4 };
     _4 <- (let (a, _) = x_1 in a);
     x_1 <- (let (a, b) = x_1 in (any t, b));
-    assume { Resolve1.resolve x_1 };
+    assume { Resolve0.resolve x_1 };
     _0 <- (_3, _4);
     goto BB2
   }
   BB2 {
+    assume { Resolve1.resolve _4 };
```
xldenis, line 80: As far as I understand `_4` and `_3` should be unitialized, so why do we resolve them here?

xldenis (commented):
The code changes seem good, but I have some questions about the impact on MLCFG, in particular it seems like we are resolving unitialized variables?

creusot/tests/should_succeed/swap_borrows.mlcfg
voidc: After the assignment, `_3` and `_4` are uninitialized and dead but then we drop them which makes them become initialized again.
<details><summary>MIR</summary>

```rust
bb1
  _0 = (move _3, move _4) [_1] -> [_1]
    live=[_3, _4] -> [_0] init=[_1, _3, _4] -> [_0, _1] uninit=[_0, _2] -> [_2, _3, _4]
  drop(_4) -> [return: bb2, unwind: bb5] [_1] -> [_1, _4]
    live=[_0] -> [_0] init=[_0, _1] -> [_0, _1, _4] uninit=[_2, _3, _4] -> [_2, _3]
bb2
  StorageDead(_4) [_1, _4] -> [_1]
    live=[_0] -> [_0] init=[_0, _1, _4] -> [_0, _1] uninit=[_2, _3] -> [_2, _3, _4]
  drop(_3) -> [return: bb3, unwind: bb6] [_1] -> [_1, _3]
    live=[_0] -> [_0] init=[_0, _1] -> [_0, _1, _3] uninit=[_2, _3, _4] -> [_2, _4]
bb3
  StorageDead(_3) [_1, _3] -> [_1]
    live=[_0] -> [_0] init=[_0, _1, _3] -> [_0, _1] uninit=[_2, _4] -> [_2, _3, _4]
```

</details>

creusot/tests/should_fail/bug/specialize.mlcfg
voidc: Same here, we drop `_1` which makes it become initialized.

creusot/tests/should_succeed/swap_borrows.mlcfg
jhjourdan: Wait... On the contrary, dropping a variable makes it UNinitialized, right? We consume its content!

voidc: The operand of drop counts as mutating use (similar to the `&mut self` argument to the drop function).

jhjourdan: Who does this choice? To me, the operand is passed by-move, so the variable should be considered as consumed after the drop.

The drop method of the Drop trait indeed takes a mutable borrow, but this is different from the drop function, which takes value by full ownership.

xldenis: Do you mean it makes `_3` and `_4` *live*? Because that's true, but it should definitely de initialize them

creusot/src/analysis/init_locals.rs
```diff
@@ -93,6 +93,9 @@ where
                 | MutatingUseContext::Yield,
             ) => {}
 
+            // Ignore drops
+            PlaceContext::MutatingUse(MutatingUseContext::Drop) => {}
```
xldenis, line 97: this should be a `kill`. the local is not just ignored it is uninitialized after the drop.

xldenis: I checked what the original rustc pass did, https://github.com/rust-lang/rust/blob/80a96467ec5675e9f69683b5c075a8b15950c341/compiler/rustc_mir_dataflow/src/impls/init_locals.rs and it did seem like `Drop` was not a `kill`. I suppose this is a fine over approximation of unitialization, but normally when you combine this with the `maybeuninit` pass that should correctly rule out the cases that were causing problems above.

creusot/tests/should_succeed/swap_borrows.mlcfg
xldenis: Something is fishy here, according to the output you posted earlier in the thread, it shouldn't be the case we resolve `_3` and `_4`. they are in both maybe initialized and maybe uninitalized so they should be ruled out from consideration. since `init = maybe init - maybe uninit`.

xldenis: This mir code makes me so confused, why are we even calling `drop` on `_4`? is this something related to drop elaboration?

creusot/src/analysis/init_locals.rs
xldenis: if you're going to adjust `maybeinit` then you should also adjust `maybeuninit` in the same manner

xldenis: Also I do believe that it should be a `kill` here.

creusot/tests/should_succeed/iterators/08_collect_extend.mlcfg
```diff
@@ -747,11 +749,13 @@ module C08CollectExtend_Extend
     goto BB4
   }
   BB4 {
+    assume { Resolve1.resolve iter_old_9 };
```
xldenis, line 752: This is extraneous and caused by not calling `emit_resolve` at the correct place. That function should elide trivial resolves like for `Ghost<T>`.

PR #779, opened 2023-05-15
Add a DeepModel derive

xldenis:
This PR adds a derive macro for `DeepModel`. This macro currently will only work for monomorphic types as it needs to use associated types in its output. 

Nonetheless, this is of benefit for CreuSAT and CDSAT.

### DIFF
--- a/creusot-contracts-proc/src/derive.rs
+++ b/creusot-contracts-proc/src/derive.rs
@@ -1,7 +1,9 @@
 mod clone;
+mod deep_model;
 mod invariant;
 mod partial_eq;
 
 pub use clone::*;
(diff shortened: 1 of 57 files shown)


jhjourdan:
Actually, I think that what we don't know how to handle is trait constraints on type variables in type declarations. But I don't see why associated types could be problematic if they don't depend on a trait constraint coming from the type declaration itself: if there is no such constraint, we can always statically resolve the associated type.

xldenis:
Yes, but deep model will have those constraints for all generic types since we need the deep model of the parameters.

jhjourdan:
Ah, you're right!

PR #780, opened 2023-05-15
Better encoding of spans

jhjourdan:
Use already imported files instead of loading new files, and support SyntaxtContext serialization.

A lot of the encoding/decoding machinery is taken from rustc.

Also, removed a test that made imported never used for spans in generated mlcfg files. For span comming from the stdlib, if a relative path is required, then use a "virtual path" instead, because the stdlib is unlikely to be installed in a relatively stable directory.

### DIFF
--- a/creusot/tests/should_succeed/hillel.mlcfg
+++ b/creusot/tests/should_succeed/hillel.mlcfg
@@ -42,7 +42,7 @@ module Core_Num_Impl11_Max
   use prelude.Int
   use prelude.UIntSize
   let constant mAX'  : usize = [@vc:do_not_keep_trace] [@vc:sp]
-    (18446744073709551615 : usize)
+    [#"/rustc/1716932743a7b3705cbf0c34db0c4e070ed1930d/library/core/src/num/uint_macros.rs" 49 8 49 27] (18446744073709551615 : usize)
 end
 module CreusotContracts_Std1_Vec_Impl0_ShallowModel_Stub
   type t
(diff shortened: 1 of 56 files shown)


creusot/tests/should_succeed/hillel.mlcfg
```diff
@@ -42,7 +42,7 @@ module Core_Num_Impl11_Max
   use prelude.Int
   use prelude.UIntSize
   let constant mAX'  : usize = [@vc:do_not_keep_trace] [@vc:sp]
-    (18446744073709551615 : usize)
+    [#"/rustc/1716932743a7b3705cbf0c34db0c4e070ed1930d/library/core/src/num/uint_macros.rs" 49 8 49 27] (18446744073709551615 : usize)
```
xldenis, line 45: We might as well not include these paths as they are unusable in why3

jhjourdan: As you wish. I thought this would be better than nothing.

jhjourdan: I could make these paths relative to the homedir. But do you think that it would stable enough? I.e., is that path of the Rust stdlib sources always the same in the homedir?

xldenis: The path of rust sources is stable to the sysroot. In practice this is the same as being stable to the home directory

jhjourdan: Alright... since Why3 does not support the ~ syntax for paths, I chose to simply discard those spans...

PR #796, opened 2023-06-03
Check lower bound on index in Seq::get

gmorenz:
Documentation [here](https://why3.lri.fr/stdlib/seq.html) confirms that this should have a lower bound

>     (* FIXME requires { 0 <= i < length s } *)


xldenis:
Thanks!

PR #799, opened 2023-06-05
Fix low-hanging bug fruit

### LINKED ISSUE #789: Syntax error in mlcfg (appears to be a name collision with a meta keyword)
The following function results in a syntax error in the resulting mlcfg

```rust
    #[logic]
    #[requires(self.contains_ref(r))]
    pub fn tag(&self, r: ReferenceIdx) -> Tag {
        match self.references.get(r.0) {
            Some(meta) => meta.tag,
            None => Tag::Disabled,
        }
    }
```

Renaming `meta` to `rmeta` solves it, I assume the issue is just that `meta` is reserved in whyml, though a quick google search didn't confirm that and I haven't run further experiments confirming it or reducing the example.

### LINKED ISSUE #791: Why3 IDE mlcfg error on small floats.
Ie:
```rust
fn i_love_floats() {
    let _ = 0.000001;
}
```
Causes the following error in Why3 IDE:
`File "../Scratch.mlcfg", line 3363, characters 10-121: Invalid floating point literal: '0.0000009999999999999999547481118258862586856139387236908078193665'`

### LINKED ISSUE #797: Destructuring tuple in logical closure generates an invalid mlcfg
```rust
use creusot_contracts::{*, logic::Mapping};

#[logic]
pub fn make_mapping() -> Mapping<(Int, Int), bool> {
    |(x, y)| false
}
```

Results in

```
File "./target/debug/testbin-rlib.mlcfg", line 17, characters 114-117:
unbound function or predicate symbol '_2''
```

The relevant line of mlcfg is

```
    [#"../src/lib.rs" 5 4 5 18] Mapping.from_fn (fun (clos'0 : (int, int)) -> let (x, y) = clos'0 in let (x, y) = _2' in false)
```

Working around this is easy, `|tup| { let (x, y) = tup; false }` avoids the issue.

xldenis:
- Fix #789
- Fix #797
- Fix #791


sarsko:
Suggestion to fix this while you’re at it: https://github.com/xldenis/creusot/issues/415

why3/Cargo.toml
```diff
@@ -12,7 +12,7 @@ itertools = "*"
 pretty = "0.11"
 indexmap = "1.2.0"
 serde = { version = "1.0", optional = true, features = ["derive"] }
-
+num = "*"
```
xldenis, line 15: Not really elegant to pull this whole thing in just to get the bit patterns of a float's components, but /shrug.

why3/src/mlcfg/printer.rs
```diff
@@ -1030,7 +1031,20 @@ impl Print for Constant {
             }
             Constant::String(s) => alloc.text(format!("{s:?}")),
             Constant::Uint(i, None) => alloc.as_string(i),
-            Constant::Float(f) => alloc.text(format!("{f:.64}")),
+            Constant::Float(f) => {
+                assert!(!f.is_nan() && f.is_finite() && f.is_normal());
```
jhjourdan, line 1035: I understand why you are assuming that f is finite (and non-Nan, which is a corolarry), but there is really no reason it should be normal.

xldenis: The reason I need to is that the printing is false if that isn't the case, and again, is it even possible to write a non-normal float _literal_?

jhjourdan: You need to do a special case (I know, this is tedious). If the float is subnormal, then multiply it by 2^60, print the mantissa as you do, and shift the exponent by 60.

PR #802, opened 2023-06-06
Remove usage of seemingly useless synthetic variables

xldenis:
@voidc, I'm not sure if I'm missing something but I *think* the usage of synthetic variables was unneeded.
Your translation turned `l <- r` into:
```
?X <- r;
resolve l;
l <- ?X; 
```

this should be the same as just doing 

```
resolve l;
l <- r;
```

right?

cc @jhjourdan


jhjourdan:
Well, ok, that seems to work too.

PR #805, opened 2023-06-14
Add `open_inv` attribute to pass arguments with an open type invariant

voidc:
This replaces the `ignore_type_invariant` mechanism and is used to avoid cycles during translation.


creusot-contracts/src/std/slice.rs
```diff
@@ -400,7 +399,7 @@ impl<'a, T> ShallowModel for IterMut<'a, T> {
     #[open(self)]
     #[trusted]
     #[ensures((^result)@.len() == (*result)@.len())]
-    fn shallow_model(self) -> Self::ShallowModelTy {
+    fn shallow_model(#[creusot::open_inv] self) -> Self::ShallowModelTy {
```
voidc, line 402: This attribute does not have any effect during the translation of `03_std_iterators`, causing a cycle.

creusot-contracts/src/std/iter/map_inv.rs
```diff
@@ -131,7 +130,7 @@ impl<I: Iterator, B, F: FnMut(I::Item, Ghost<Seq<I::Item>>) -> B> MapInv<I, I::I
     #[open(self)]
     #[predicate]
     #[ensures(self.produced.inner() == Seq::EMPTY ==> result == Self::preservation(self.iter, self.func))]
-    pub fn preservation_inv(self) -> bool {
+    pub fn preservation_inv(#[creusot::open_inv] self) -> bool {
```
voidc, line 133: Same here with `06_knights_tour`

PR #808, opened 2023-06-18
Add support for associated types in declarations

xldenis:
This is implements support for associated types in declarations so that we could do somethign like verify `Peek`.
However, I have to say I'm not super happy with this solution, I find it to be leaky, in particular translating types now *has* to do trait resolution as the nested associated types we use could be unnormalized. I ended up pushing this into `CloneMap::projection` as the alternative was threading a `ParamEnv` everywhere which is quite ugly.

How this works: for each associated type in a declaration we add a generic parameter to the type. Then every usage of that type instantiates that generic correctly. 
If these types are nested in other types then we bubble the generics upwards. 

If we got rid of why3 level polymorphism it would make things signficantly nicer as we could "just" do like with functions etc...


xldenis:
@jhjourdan @voidc what do y'all think?

voidc:
looks good to me

PR #809, opened 2023-06-21
Use generated type invariants

voidc:
This does not yet include the blanket implementation of the `Invariant` trait.


creusot/tests/should_succeed/iterators/07_fuse.rs
```diff
@@ -73,8 +77,8 @@ impl<I: Iterator> Invariant for Fuse<I> {
     #[predicate]
     fn invariant(self) -> bool {
```
voidc, line 78: This shouldn't be required but we still need it in this case due to the lack of specialization

PR #810, opened 2023-06-21
Implement the `Invariant` trait for all types

voidc:
It is somewhat unfortunate that this requires us to enable the incomplete `specialization` feature everywhere. (`min_specialization` is insufficient)


xldenis:
> It is somewhat unfortunate that this requires us to enable the incomplete specialization feature everywhere. (min_specialization is insufficient)

:/

xldenis:
Why is that? Normally we don't need to depend on trait bounds anymore, right?

voidc:
> Why is that? Normally we don't need to depend on trait bounds anymore, right?

We still have bounds for some impls. For example, `impl<I: Iterator> Invariant for Enumerate<I>`.
If we had control over the `Enumerate` type, we could add the `Iterator` bound to the type, which would permit the impl under the `min_specialization` rules.

xldenis:
But why do you need that bound, shouldn't you be using the unconstrained `inv` function?

jhjourdan:
I guess this just means invariants should not rely on specialization: when generating the inv function, we have three cases:
- either we know the corresponding `Invariant` instance (i.e., it exists and it is not further specializable), then it is easy to generate.
- either we know there is no `Invariant` instance for this type (i.e., even with more constraints in the constext and more traits loaded), no instance will work, then it is again easy (the user invariant is true).
- or we don't know the user invariant, and it should be treated opaquely.

I guess this is painful to implement?

voidc:
> But why do you need that bound, shouldn't you be using the unconstrained `inv` function?

In the implementation of the `Invariant` trait, we call `self.iter.produces` which requires the `I: Iterator` bound.

> I guess this is painful to implement?

Specialization gives us exactly that: When we resolve to the default impl, we can check whether it is further specializable. In this case, we insert a call to the opaque trait method and in the other case, we know the invariant is trivial. The issue rather getting all tests to pass with the additional (opaque) invariants.

xldenis:
>In the implementation of the Invariant trait, we call self.iter.produces which requires the I: Iterator bound.

but wouldn't we remove that bound as _well_ since it would use `inv`? I guess my point is now that we have this unconstrained `inv` function we should *never* refer to `Invariant`, or am i missing something still?

voidc:
The `Invariant` trait represents the user invariant. While we generally do not want to call the `Invariant::invariant` function (but the opaque `inv` function, including the structural invariant), we still need to implement the trait to attach a user invariant to a type.

xldenis:
Yes we would need to refer to it to implement an invariant (and actually we can probably remove that limiation via macros), but other predicates should never refer to it afterwards, no?

voidc:
I don't see a reason why they should.

voidc:
Currently, we use trait resolution to determine whether a type needs an invariant. For example, when elaborating the invariants for a function with an argument of type `Option<NonZero>`, we walk the constituent types and try to resolve the `Invariant` trait for each. If at least one resolves to a non-trivial implementation, we consider the whole invariant non-trivial and thus add the corresponding precondition. Specialization comes into play for a type like `Option<T>` because we would resolve `Invariant for T` to the (further specializabe) blanket implementation. An alternative approach would be to treat parameter types as special. (which I tried out on the `tyinv-gen-param` branch) This approach does not rely on specialization and seems to be more akin to what @xldenis imagined? The only downside of this approach is that we cannot define user invariants for arbitrary types (i.e. non-ADT) as we only resolve the invariant trait for concrete ADT types. However, IMO this is a sensible limitation.

xldenis:
closing after discussion.

PR #813, opened 2023-06-23
Fix transparency issues

dewert99:
This just switches `FMap::get` to be transparent


dewert99:
I also found some of `iter::repeat`'s methods should be transparent

xldenis (approved):

xldenis:
A bunch of sessions are obsolete, too many in fact for it to be your changes (I think). It seems why3 did another big update recently

xldenis:
try rebasing.

dewert99:
You should be good to merge it now

Issue #436, opened 2022-06-22
Bugs about ghost code.


jhjourdan:
Consider this piece of code:

```rust
extern crate creusot_contracts;

use creusot_contracts::*;

struct S {
    g: Ghost<i32>
}

#[logic]
fn prophecy(x: &mut S) -> i32 {
    pearlite!{ *(^x).g }
}

pub fn f() {
    let b = &mut S{ g:ghost! { 1 } };
    b.g = ghost!{ (prophecy(b) + 1) };
    proof_assert! { false }
}
```

There are three issues here:
1- I need to put parentheses in ` ghost!{ (prophecy(b) + 1) };`, otherwise I get weird type errors. This sounds like an easy to fix bug in the `ghost` macro.

2- The `Ghost` in `struct S` is not translated into a ghost field. Similarly, a borrow of a ghost is not translated into anything Ghost, which is blocking since then we cannot make use of it.

In my view, the way we should solve this issue is by not using the `ghost` keyword in WhyML, and make sure that `Ghost` is indeed a monad that cannot be escaped, and make this check in Creusot.

3- Once this code compile correctly, it exhibits a causality loop. Using logic functions such as `prophecy` should be forbidden in ghost code.

dewert99:
Now that #488 has been merged this seems to have become a soundness issue. However, I'm not sure it would be okay to make logic functions forbidden in ghost code (eg. The desugaring of a for loop currently includes `produced = ghost!(produced.push(e))` which calls `Ghost::deref` and `Seq::push` both of which are logic functions).  Currently, there doesn't seem to be much difference between predicates and logic functions (that return bool), so I wonder if it would work to forbid calling predicates in logic functions and ghost code, and restrict `^`/`creusot_contracts::stubs::fin` to being used in predicates.

xldenis:
> Now that https://github.com/xldenis/creusot/pull/488 has been merged this seems to have become a soundness issue.

Yea, the example given above demonstrates that. Allowing access to prophetic information in ghost code is problematic. 

I'm leaving on vacation next week, but when I get back I think the most crucial concern will be overhauling ghost code and placing it on a firm theoretical foundation. There are several other questions that I want to answer simultaneously:

1. Should ghost code have ownership typing? 
2. Should ghost code have access to logical types and values?
3. Related to the previous questions: should there be a 'logic ghost code' and a 'linear ghost code'? 
[...]

PR #818, opened 2023-06-26
Disallow final value in logic functions and ghost

voidc:
This extends the purity check to consider three levels of purity: logic, ghost, and program.
Specifications (function pre- and postconditions and loop invariants) and predicate functions constitute a context of logic purity and can thus call predicate, logic, and builtin functions, and use pearlite operators.
Ghost blocks and logic functions constitute a context of ghost purity, which permits the same operations as the logic context except using the final value operator.
Lastly, every other function constitutes a context of program purity, in which only other program functions may be called.


creusot/src/translation/pearlite.rs
```diff
@@ -225,8 +225,9 @@ pub(crate) fn pearlite<'tcx>(
         return Err(Error::new(ctx.def_span(id), "type checking failed"));
     };
 
+    // let purity = Purity::of_def_id(ctx.tcx, id.to_def_id());
     visit::walk_expr(
```
voidc, line 229: Why do we have to check purity again here? Shouldn't `before_analysis` already check every HIR item?

voidc:
We also cannot allow calling predicates in ghost context as this would again permit depending on prophecies. However, we commonly call functions like `lt_log` in logic functions, which means changing them from predicates to logic functions. Ideally, such functions would be `pure` functions that can be called in any context.

creusot/src/translation/pearlite.rs
xldenis: hmm, i think its a relic of times long past.

xldenis:
This looks good to me! Perhaps we could also rename `logic` to `ghost` (or `ghost_fn` to make it easier for now) in a future PR, and `predicate` to `logic`?

xldenis:
Yea I think the idea would be to rename predicate to logic and lift the restriction on return types.

sarsko:
Hmm, I'm quite a fan of the `logic`/`predicate` distinction. They allow for easier grouping, which I have found valuable as a proof grows, and it is also useful when reading a proof written by others. Tbh, I would even advocate for the addition of `#[lemma]`, to enable further grouping and avoid the current scheme of:
```rust
#[logic]
pub fn lemma_bla_blabla() {}
```

I don't see the case for enabling predicates to return non-booleans. What's the reasoning behind this / why would this be useful?

dewert99:
Basically, the `ghost!` macro is currently unsound since it can call `logic` functions and they can use the final `^` operator (see #436).  One solution to the problem would be to disallow the final `^` operator in `logic` functions and only allow it in predicates, this could potentially lead to cases where someone wants to use `^` in a function and return a non-boolean.

Maybe instead of framing things in terms of renaming, we should just say we're adding a new `ghost` attribute that works like `logic` but disallows the final operator and then only allow `ghost` functions in `ghost!` macros.

PR #831, opened 2023-07-12
Fix session_stats test

voidc:


xldenis:
This is a step forward but I would really prefer using the git api to read the old version directly as if someone interrupts this command they won't realize that there is a stash which needs to be popped.

xldenis:
Also, please add a short (1 sentence is fine) PR body.

PR #832, opened 2023-07-12
Generate invariant modules for & and Box

voidc:
Add for `Seq` and slices.


creusot/src/backend/ty_inv.rs
```diff
@@ -267,6 +276,13 @@ fn build_inv_exp_struct<'tcx>(
                     inv.subst(&[("a".into(), inner)].into());
                     Some(inv)
                 }
+                "seq.Seq.seq" => build_inv_exp_seq(
```
xldenis, line 279: I don't like this pattern (don't fix it here). I think we should either use a rustc_diagnostic_item or a creusot::item here.

xldenis (approved):

PR #820, opened 2023-06-27
Add type invariants for generic parameters

xldenis:
Rather than taking the approach in #810, we are going to go with this.

PR #834, opened 2023-07-18
Various fixes around type invariants

voidc:
Factored out from #820.


voidc:
The `take_first_mut` test is broken.

PR #836, opened 2023-07-20
Allow logically reborrowing through index

xldenis:
This allows `&mut x[0]` to compile in pearlite (ok without the `&mut` syntax which must be added implicitly by rust) when `x : &mut &mut [T]`. 

cc @jhjourdan


jhjourdan:
Seems like there is a typing problem.

PR #838, opened 2023-07-21
Ghost refactor

jhjourdan:
Renamings : ghost! -> gh!, #[logic] -> #[ghost], #[spec_logic] -> #[logic].

The gh macro now takes pearlite syntax.


creusot-contracts-proc/src/lib.rs
```diff
@@ -406,14 +405,17 @@ pub fn proof_assert(assertion: TS1) -> TS1 {
...
     TS1::from(quote! {
         {
-            ::creusot_contracts::ghost::Ghost::from_fn(
+            ::creusot_contracts::__stubs::ghost_from_fn(
                 #[creusot::no_translate]
                 #[creusot::spec]
-                #[creusot::spec::ghost] || { ::creusot_contracts::ghost::Ghost::new (#assertion) }
+                #[creusot::spec::ghost]
+                || { ::creusot_contracts::ghost::Ghost::new (#assert_body) }
             )
```
xldenis, line 419: its worth doing another pass and seeing if we can't just do the same thing as `logic` though I suspect not due to interactions with initialization checking etc,..

jhjourdan: What do you want to do?

creusot/src/ctx.rs
```diff
@@ -309,7 +309,10 @@ impl<'tcx, 'sess> TranslationCtx<'tcx> {
     }
 
     fn mk_opacity(&self, item: DefId) -> Opacity {
-        if !matches!(util::item_type(self.tcx, item), ItemType::Predicate | ItemType::Logic) {
+        if !matches!(
+            util::item_type(self.tcx, item),
+            ItemType::Predicate | ItemType::Logic | ItemType::Ghost
```
xldenis, line 314: We should probably add an `is_ghost` or similar function on `item_type`, these kinds of checks are frequent.

jhjourdan: YEah... (Btw, this was an unsoundness where we could implement a logic method with a ghost method.

creusot/src/translation/function/terminator.rs
```diff
@@ -72,15 +71,14 @@ impl<'tcx> BodyTranslator<'_, 'tcx> {
                 }
 
                 let (fun_def_id, subst) = func_defid(func).expect("expected call with function");
-                if let Some(param) = subst.get(1) &&
-                    let GenericArgKind::Type(ty) = param.unpack() &&
-                    let Some(def_id) = is_ghost_closure(self.tcx, ty) {
-                    let mut assertion = self.assertions.remove(&def_id).unwrap();
+                if Some(fun_def_id) == self.tcx.get_diagnostic_item(Symbol::intern("ghost_from_fn"))
```
xldenis, line 74: this is a much needed extra check 👍

jhjourdan:
Something which is not clear to me: to we properly check ghostness of logical closures?

creusot-contracts-proc/src/lib.rs
xldenis: I'll take a stab at this later, I think I have the most expertise fighting `rustc` on this aspect.

jhjourdan:
(I assume not.)

xldenis:
>  to we properly check ghostness of logical closures

What do you mean by this? They are checked to be valid Pearlite syntax, and must be well-typed. Apart from that they receive no special verification.

jhjourdan:
Do we access propeties in logical closures used in a ghost context?

jhjourdan:
And are we allowed to do so in a logical context?

xldenis:
> Do we access propeties in logical closures used in a ghost context?

I don't understand this sentence... what properties?

jhjourdan:
Ok, I'll write a test.

xldenis:
A logical closure is just turned into a why3 `fun x -> ...`, there is nothing beyond this.

PR #841, opened 2023-07-21
Cleanup stubs, forbid macros in Pearlite.

jhjourdan:


creusot-contracts-proc/src/pretyping.rs
```diff
@@ -30,8 +30,11 @@ impl EncodeError {
 pub fn encode_term(term: &RT) -> Result<TokenStream, EncodeError> {
     let sp = term.span();
     match term {
-        // Macros could contain further pearlite expressions..
-        RT::Macro(m) => Ok(quote_spanned! {sp=> #m}),
```
xldenis, line 34: This should be undone. It's needed to do things like `proof_assert!` in logic functions.

jhjourdan: I'm not convinced. Anyway, if we want to have `proof_assert!` in logic functions, then we should give it a different interpretation than in program functions. Hence, we would probably interpret it directly (and not through a macro) in pearlite's parser (just like the pearlite parser does for avoiding nested `pearlite!` invocations).

In addition, allowing macros there really cause a problem, since it gives access to non-pearlite syntax in logical code.

xldenis: If you want to whitelist it sure, that can work. 

However, I am also not convinced by the argument that we could allow non-pearlite syntax. If that happens then Creusot will crash/error out.

PR #842, opened 2023-07-25
Clone `TyInv_Trivial` for param types and trivial ADTs

voidc:
When cloning the type invariant axiom for a type `t`, we now clone the trivial axiom (`forall x: t. inv x`) in more cases. In particular, we leverage parametricity to consider type invariants of parameter types trivial, as well as those of trivial ADTs.


jhjourdan:
Do you check that the type variable in question is never used in "open invariant" mode ?

I.e., before marking type parameter T as "trivial invariant", you should check that the function is never calling a logical or program function:
  - marked as open invariant, with the return or argument type mentioning T
  - with one parameter marked as open invariant, with the type of this parameter mentioning T

Is this clear?

voidc:
Hmm. it's not really clear to me. I think the issue is rather that if a function has an argument of generic type marked as having an open invariant it cannot assume that it has a trivial invariant. For example:
```rust
fn foo<T>(#[open_inv] x: T) {
    // resolve x
}

fn unsound() {
    let mut x = NonZero(3);
    let b = &mut x;
    *b = 0;
    foo(b);
}
```
That being said, usage of the `[open_inv]` attribute currently also has other issues, making its use "on your own risk".
Thus, I think fixing it it orthogonal to this change.

xldenis:
Ok so there is the remaining issue about open invariants, but does this PR fix type parameters to not be considered trivial in their clients?

voidc:
No this is only about clones; the client only sees the contract. Elaborating type invariants for param types will be handled by #820.

jhjourdan:
> That being said, usage of the [open_inv] attribute currently also has other issues

Which? Is it documented?

> Hmm. it's not really clear to me. I think the issue is rather that if a function has an argument of generic type marked as having an open invariant it cannot assume that it has a trivial invariant. For example:

Indeed, that's another issue. But the issue I am mentioning is:

```rust
trait Tr {
  #[open_inv]
  fn f() -> Self;
}

// Implicit for a user : #[ensure(inv(result))]
fn g<T: Tr>() -> T {
  return T::f(); // May return an element which is not valid.
}
```

voidc:
The `#[open_inv]` attribute at the function level not only affects the return value but completely "disables" type invariants. i.e. no guarded quantifiers, no assert on resolve. In its current state it's only meant to be used as an escape hatch for cases where would otherwise run into problems with translation cycles (e.g. the map tests). I agree that a proper implementation of this feature should have additional checks ensuring that it is used correctly.

jhjourdan:
Oh, wow... I did not know this was THAT broken... We should really either fix it or figure out a way to remove it.

jhjourdan:
Sure.

voidc:
I opened an issue to track fixing `#[open_inv]` and will merge this PR.

PR #846, opened 2023-07-27
Float quantifiers

voidc:
When elaborating type invariants for quantifiers, float inner quantifiers.
e.g. `forall<x> inv(x) ==> forall<y> P(y)` gets turned into `forall<x> forall<y> inv(x) ==> P(y)`


xldenis:
> doesn't seem to have a big overall impact on proof times
That's good!

Is this using a baseline from your machine? Even if it's just 11% that's a full ~45s!

> should_succeed/red_black_tree: t_old=403.29 t_new=356.78 (-11.5%) n_old=76467140 n_new=70683595 (-7.6%)

voidc:
should be, I last updated the proof in #820 
true, 45s is not bad

PR #847, opened 2023-07-28
Generate type invariants for arrays, closures, and the never type

voidc:
We only handle array types with non-generic lengths.
The invariant modules generated for closures still have issues with clones.

### DIFF
--- a/creusot/tests/should_succeed/type_invariants/generated.mlcfg
+++ b/creusot/tests/should_succeed/type_invariants/generated.mlcfg
@@ -131,9 +131,8 @@ module Generated_UseFoo_Interface
   use Generated_Foo_Type as Generated_Foo_Type
   clone CreusotContracts_Invariant_Inv_Stub as Inv0 with
     type t = Generated_Foo_Type.t_foo (Generated_Foo_Type.t_foo uint32, borrowed (Generated_Sum10_Type.t_sum10))
-  val use_foo [#"../generated.rs" 23 0 23 61] (x : Generated_Foo_Type.t_foo (Generated_Foo_Type.t_foo uint32, borrowed (Generated_Sum10_Type.t_sum10))) : ()
-    requires {[#"../generated.rs" 22 11 22 28] Inv0.inv x}
-    requires {[#"../generated.rs" 23 19 23 20] Inv0.inv x}
+  val use_foo [#"../generated.rs" 22 0 22 61] (x : Generated_Foo_Type.t_foo (Generated_Foo_Type.t_foo uint32, borrowed (Generated_Sum10_Type.t_sum10))) : ()
+    requires {[#"../generated.rs" 22 19 22 20] Inv0.inv x}
     
 end
 module Generated_UseFoo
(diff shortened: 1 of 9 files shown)


creusot/tests/should_succeed/type_invariants/generated.mlcfg
```diff
@@ -198,7 +196,367 @@ module Generated_UseFoo
...
+end
+module Generated_TestClosure_Closure0_Inv
+  use Generated_Sum10_Type as Generated_Sum10_Type
+  clone CreusotContracts_Invariant_Inv_Stub as Inv1 with
+    type t = Generated_Sum10_Type.t_sum10
+  clone CreusotContracts_Invariant_Inv_Stub as Inv0 with
+    type t = Closure00.generated_testclosure_closure0
+  clone Generated_TestClosure_Closure0_Interface as Closure00 with
+    predicate Inv0.inv = Inv0.inv,
+    predicate Inv1.inv = Inv1.inv
```
voidc, line 511: There is a cyclic dependency here: `Inv0` requires the closure type exported by `Closure00` which in turn requires the invariant symbol. One solution would be to first generate a `use Generated_TestClosure_Closure0_Type`.

jhjourdan:
> We only handle array types with non-generic lengths.

Is this because we don't support generic lengths in general?

What's happening if there is an array of generic length? Does Creusot emit an error? (It should.)

xldenis:
We ignore the length

xldenis:
> The invariant modules generated for closures still have issues with clones.

What's the issue? Do you know?

voidc:
Regarding arrays: in theory, we could ignore the length but in practice, we can end up with distinct `inv` symbols for array types with different lengths, so we need to generate axioms for all of them.

Regarding closure clones: the issue is that cloning the closure type also clones the closure interface module which in turn requires the type invariant (see the commented lines).

jhjourdan:
For arrays : I don't really care about what's happening with arrays of generic size. But I do care that we don't introduce soundness issue because of that. So either we treat them correctly, or we error.

creusot/tests/should_succeed/type_invariants/generated.mlcfg
```diff
@@ -198,7 +196,367 @@ module Generated_UseFoo
...
+    [#"../generated.rs" 1 0 1 0] true
+  predicate precondition [#"../generated.rs" 32 12 32 19] (self : generated_testclosure_closure0) (_ : ()) =
+    [#"../generated.rs" 1 0 1 0] true
+  predicate postcondition_once [#"../generated.rs" 32 12 32 19] (self : generated_testclosure_closure0) (_ : ()) (result : Generated_Sum10_Type.t_sum10)
+    
+   =
+    [#"../generated.rs" 1 0 1 0] true
+  val generated_TestClosure_Closure0 [#"../generated.rs" 32 12 32 19] (_1 : generated_testclosure_closure0) : Generated_Sum10_Type.t_sum10
+    requires {[#"../generated.rs" 1 0 1 0] Inv0.inv _1}
+    ensures { [#"../generated.rs" 32 12 32 19] Inv1.inv result }
```
voidc, line 457: Does it even make sense to have a contract on the `generated_TestClosure_Closure0` function? I would expect it to only ever be called through the `Fn*` traits which have independent contracts...

xldenis: > edit: I was wrong 

yep, direct calls to the closure go through directly

xldenis:
I'm going to close this as I think the PR is probably dead and would need to be rewritten against master anyways.

PR #852, opened 2023-08-09
Allow logically reborrowing through derefs of ghost and indexing things

xldenis:
This more correctly implements #836 but the code is suspicious enough that I don't want to merge this yet.

In particular, given `x : &mut Ghost<Vec<T>>` the expression `let y : &mut T = &mut x[0]` is expands to: `&mut index_logic (* deref (& * x))`. The real problem is the insertion of that extra `&` inside. 
My current implementation detects and correctly handles this case for obvious scenarios by 'skipping over it', but is that really what we want to do in general?


xldenis:
Rewritten since our discussion yesterday @jhjourdan, this turns `* Ghost::deref(& _)` into a single `Ghost::inner(_)`

PR #854, opened 2023-09-08
Remove ghost state from Fuse proof

xldenis:
I realized we don't need ghost state in `Fuse` at all. This also makes it possible to specify the stdlib fuse since we don't introduce a cardinality problem.


jhjourdan:
The new spec looks good to me... I don't know why we needed this ghost before. That's probably because we used to have a different spec for next in the None case.

jhjourdan:
Before merging, could you please include in your PR the extern_spec change ?

jhjourdan:
> Before merging, could you please include in your PR the extern_spec change ?

?

xldenis:
I had turned on auto-merge and forgot to turn it off, my bad.

PR #865, opened 2023-09-21
Fix type invariants for `Vec`

voidc:
This adds an `impl<T> Invariant for Vec<T>` which is marked as `structural_inv`. It is thus only considered non-trivial if `T` is non-trivial. I also fixed a bug where we incorrectly assumed that `Seq<T>` has a trivial invariant. This is the cause of most changes in the tests.


jhjourdan:
Could you please move the new trait to the module `std::vec` module? This is what we do shallow/deep model traits.

jhjourdan:
Another question: am I correct that we don't define an `Invariant` instance for `Seq`?  Isn't this a bug?

xldenis:
> Another question: am I correct that we don't define an Invariant instance for Seq? Isn't this a bug?

Looking at the diffs, this PR seems to be fixing that actually.

jhjourdan:
How so? Note that there is a similar problem with mappings.

BTW, we have an additional problem with mappings and extentionnality: the extensionnality property (coming from why3) states that for any two mappings m1 and m2:

(forall k : A, get m1 k = get m2 k)  -> m1 = m2

While we want actually:

(forall k : A, inv(k) -> get m1 k = get m2 k)  -> m1 = m2

I don't see how we can fix that easily without breaking soundness somewhere...

xldenis:
Oh I meant that `Seq` used to be getting the 'trivial invariant' treatment and now its adding a bunch of clones for invariants.

xldenis:
> I don't see how we can fix that easily without breaking soundness somewhere...

Simple, we have to stop using why3 container libraries. it's been a long time coming honestly.

jhjourdan:
> Simple, we have to stop using why3 container libraries. it's been a long time coming honestly.

I think it's orthogonal. If we use a custom definitions for mappings with a bizarre law of extentionnality and bizarre get-after-set rules, then we won't be able to use SMT's theories for arrays and sequences.

The more I think about it, the more I get convinced that we should use something closer to Why3's type invariants. What was the downside of this, again?

voidc:
> Another question: am I correct that we don't define an `Invariant` instance for `Seq`? Isn't this a bug?

The invariants for `Seq` and `Ghost` are hard coded at the moment. We could change them to use also use the `structural_inv` mechanism but this would introduce an extra level of indirection (inv s = forall<i> inv(s[i]) vs inv s = user_invariant s = forall<i> inv(s[i])).

creusot/src/backend/ty_inv.rs
```diff
@@ -265,8 +276,10 @@ fn build_inv_exp_struct<'tcx>(
             build_inv_exp(ctx, names, ident, adt_subst.type_at(0), param_env, mode)
         }
         TyKind::Adt(adt_def, adt_subst) if util::get_builtin(ctx.tcx, adt_def.did()).is_some() => {
+            // should these be structural user invs?
```
voidc, line 279: 👆

jhjourdan:
Let's merge this for now, since this is clearly an improvement, but there is really something that is fishy here, that I'd like to be improved.

xldenis:
> The more I think about it, the more I get convinced that we should use something closer to Why3's type invariants. What was the downside of this, again?

I think it was incompatible with cloning.

PR #867, opened 2023-09-21
Pearlite support for `Range`

xldenis:
Adds parser support for ranges like `0..10` and also adds an instance of `IndexLogic` for `Seq` that uses that to call `subsequence`. This should make sequence heavy specifications _much_ nicer to read. 

TODO:

- [ ] Add other instances for other forms of range
- [ ] Update tests to use new syntax


Lysxia:
subsumed by #1858

PR #880, opened 2023-10-04
Invariant cleanup

jhjourdan:
- Get rid of `open_inv`
- Get rid of the derive macro, which is no longer needed
- Get rid of uses of `inv(...)` here and there in pre/post-conditions
        (They are now automatically inserted."
- Get rid of trivial instances of the `Invariant` trait
- Get rid of `is_inhabited`
        The current version gave it a default implementation, so that
        the VC was never generated. It turns out that these VC are often
        difficult to prove, or even impossible (example: `Map` iterator).
        So it seems like we need to live with the eventuality of empty types,
        and find other fixes for the soundness issues this causes.

Cc @voidc


xldenis:
You updated several proofs to use Alt-Ergo 2.4.3 which CI doesn't yet have (we will probably just want to pass straight to 2.5.1). Could you edit the relevant sessions to use 2.4.2?

xldenis:
We need to be careful that there's no way to construct a borrow of an empty type or type with an 'empty' invariant.

For example, I believe the following is possible:

```
struct NotQuiteEmpty {};

impl Invariant for NotQuiteEmpty { #[predicate] fn invariant(self) -> bool  { false } } 

fn boom() {
  let x = &mut NotQuiteEmpty {};
  
  proof_assert! { false } ;
} 
```

voidc:
The changes look reasonable to me. I find the behavior that trait laws with default implementations do not generate goals in impls a bit surprising though. It seems like giving a default implementation acts like `#[trusted]`? Is this intended or a bug?

> We need to be careful that there's no way to construct a borrow of an empty type or type with an 'empty' invariant.
> 
> For example, I believe the following is possible:
> 
> ```
> struct NotQuiteEmpty {};
> 
> impl Invariant for NotQuiteEmpty { #[predicate] fn invariant(self) -> bool  { false } } 
> 
> fn boom() {
>   let x = &mut NotQuiteEmpty {};
>   
>   proof_assert! { false } ;
> } 
> ```
> 
> EDIT: Perhaps not as I imagine we will be required to prove the invariant when we initialize the constructor?

This example would indeed be unsound, without the `is_inhabited` law. We do not require proving the invariant when constructing values with a literal expression. What we could do instead is require proving `is_inhabited` when borrowing. This would allow types to be uninhabited as long as they are not borrowed.

xldenis:
> It seems like giving a default implementation acts like #[trusted]? Is this intended or a bug?

They do generate proof obliations _at the trait declaration_, the real problem is we don't build the proofs for `creusot-contracts` which I think is a fairly urgent thing to fix.

jhjourdan:
Anyway, I'm merging since this PR seems consensual.

PR #886, opened 2023-10-10
Use type invariants in bdd and sparse_arrays

voidc:
Both tests were already using the concept of type invariants but did not use the `Invariant` trait.
The sparse_array test required some adjustments: The `ShallowModel` implementation for `Sparse<T>` had a trusted contract to define the interpretation of the model axiomatically.
The generated axiom would have an ' inv ' condition since we add type invariants for logic functions with postconditions.
This led to a cyclic user invariant, which used the shallow model in its definition.
Furthermore, the refinement goal for the shallow model impl was unprovable.
Therefore, I added an extra axiom and include it in every function via `gh!`.
Ideally, with inherent lemma functions, the axiom would be included automatically.


jhjourdan:
Actually, instead of the trusted axiom, now that we have ghosts, what we want to use is a ghost field containing the model and the invariant stating that the model has the right property wrt. the non-ghost fields.

xldenis:
Agreed though this can be done as part of a follow up refactoring; it's rather independent of using the invariant trait or not.

jhjourdan:
Thanks to #888, there should be no problem with sparse array now. Could you please update the PR?

jhjourdan:
Thanks, merging !

PR #888, opened 2023-10-11
Sequence comprehensions.

jhjourdan:
- Add sequence comprehensions in standard library via the method new
- Logical closure are no longer annotated with creusot::decl::logic. This made them being checked for opacity.
- Use sequence comprehensions in sparse_array example.


xldenis:
oh, i like this

PR #892, opened 2023-10-13
Forbid the use of borrowed variables in ghost code

voidc:
This PR checks Pearlite terms in the `gh!` macro for uses of borrowed variables. Using borrowed variables in ghost code leads to soundness issues and is therefore rejected.


xldenis:
Does this also ban usage inside of `proof_assert`? I think it should

creusot/src/resolve.rs
```diff
@@ -160,6 +160,11 @@ impl<'body, 'tcx> EagerResolver<'body, 'tcx> {
         self.dead_locals()
     }
 
+    pub fn frozen_locals_before(&mut self, loc: Location) -> BitSet<Local> {
```
xldenis, line 163: I'm not sure I'm a huge fan of the resolver exposing this api used for something other than resolving.

voidc:
> Does this also ban usage inside of `proof_assert`? I think it should

It does not at the moment. Why do you think it should? There is no soundness issue AFAIK.

xldenis:
It's not unsound but the value of a borrowed place is counterintuitive and people attempt to make impossible assertions.

voidc:
As I see it, borrowed variables and the final value operator are two ways to refer to the same prophecy. Nothing stops users from using the latter before resolving, enabling impossible assertions. Therefore, we should either allow both or neither.

```rust
let mut x = 3;
let r = &mut x;
proof_assert! { ^r == 5 };
proof_assert! { x == 5 }; // equivalent to first assert
*r = 5;
```

xldenis:
Yea referring to `^` in asserts is almost always a mistake, but already banning the borrowed variable would be good and prevent confusing things like:

```
let mut x = 0;
let a = &mut x;
a += 1;
proof_assert! {x == 1};
a += 1;
proof_assert! { x == 2};
```

This example won't work in Creusot but can be confusing to users. If we can simply ban this code it would rpevent these issues entirely.

creusot/src/resolve.rs
xldenis: Why do you think this is the appropriate place for this API?

jhjourdan:
> Yea referring to ^ in asserts is almost always a mistake

I disagree. We have at least one relevant example in `mapping_test.rs`, and we often use prophecies in loop invariants.

BTW, @voidc, are borrowed variables forbidden in loop invariants too?

voidc:
> BTW, @voidc, are borrowed variables forbidden in loop invariants too?

I added that but it breaks the hashmap test. In `add` we reborrow `self.buckets` but the use self in the loop invariants.

jhjourdan:
That should be fixed by #896.

voidc:
> That should be fixed by #896.

thanks!

jhjourdan:
@xldenis : anything missing before merging?

xldenis:
I think this is fine

jhjourdan:
Thanks, @voidc !

creusot/src/resolve.rs
voidc: Ah, I forgot to reply, sorry. I don't think this necessarily is the right place for the API. It's just that this is where we compute the data-flow analysis, which I didn't want to duplicate.

PR #893, opened 2023-10-16
Add a Resolve derive

xldenis:
Adds a derive macro for `Resolve` based on a request by @dewert99.

Also _slightly_ generalizes the `DeepModel` derive though I think it should be changed further.

### DIFF
--- a/creusot-contracts-proc/src/derive/resolve.rs
+++ b/creusot-contracts-proc/src/derive/resolve.rs
@@ -1,0 +1,115 @@
+use proc_macro2::{Ident, TokenStream};
+use quote::{format_ident, quote, quote_spanned};
+use syn::{parse_macro_input, spanned::Spanned, Data, DeriveInput, Fields, Index};
+
+pub fn derive_resolve(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
+    let input = parse_macro_input!(input as DeriveInput);
+
+    let name = input.ident;
+
+    let (impl_generics, ty_generics, where_clause) = input.generics.split_for_impl();
+
+    let eq = resolve(&name, &input.data);
+
+    let expanded = quote! {
+        #[::creusot_contracts::trusted]
+        impl #impl_generics ::creusot_contracts::Resolve for #name #ty_generics #where_clause {
+            #[predicate]
+            #[open]
+            fn resolve(self) -> bool {
+                use ::creusot_contracts::Resolve;
+                #eq
+            }
+        }
+    };
+
+    proc_macro::TokenStream::from(expanded)
+}
+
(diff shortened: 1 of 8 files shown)


creusot-contracts-proc/src/derive/resolve.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+fn resolve(base_ident: &Ident, data: &Data) -> TokenStream {
+    match *data {
+        Data::Struct(ref data) => match data.fields {
+            Fields::Named(ref fields) => {
+                let recurse = fields.named.iter().map(|f| {
+                    let name = &f.ident;
+                    quote_spanned! {f.span()=>
+                        Resolve::resolve(&self.#name)
```
dewert99, line 36: I don't think this should create a shared reference since the resolve implementation of shared references is `true`. (Same on lines 47 and 104)

creusot-contracts-proc/src/derive/resolve.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    let (impl_generics, ty_generics, where_clause) = input.generics.split_for_impl();
+
+    let eq = resolve(&name, &input.data);
+
+    let expanded = quote! {
+        #[::creusot_contracts::trusted]
+        impl #impl_generics ::creusot_contracts::Resolve for #name #ty_generics #where_clause {
+            #[predicate]
+            #[open]
```
dewert99, line 18: It would probably be better to use:
```rust
#[::creusot_contracts::predicate]
#[::creusot_contracts::open]
```

creusot-contracts-proc/src/derive/resolve.rs
dewert99: Currently 
```rust
use creusot_contracts::{Resolve, open, predicate};

#[derive(Resolve)]
struct Test<'a>(&'a mut u32);

fn test() {
    let mut x = 0;
    let m = &mut x;
    let _ = Test(m);
    assert_eq!(x, 0);
}
```
Isn't able to be verified

dewert99:
Thanks for working on this

creusot-contracts-proc/src/derive/resolve.rs
xldenis: Yea ure right, I'll fix it .

jhjourdan:
How does this derive play with opacity? I guess in the case of private fields, it will generate a resolve that publicly exposes these fields, which is forbidden.

In this case (concrete example: red-black trees), we would like to le the user define the resolve predicate, but still generate a VC to make sure it is sound (so that we avoid the trusted attribute).

PR #898, opened 2023-10-20
simple_triggers option that adds triggers to def and spec axioms

dewert99:
I thought I would start to explore what it would look like for Creusot to manually pick triggers for `def` and `spec` axioms instead of having tools infer them, so I added a `cargo-creusot` option to do this. Even when the option is enabled specific functions can opt out of triggers by using `#[creusot::decl::no_trigger]`. Since `law`s are not meant to be called directly the `law` attribute macro automatically adds `#[creusot::decl::no_trigger]`. I also noticed that `law`s were previously considered `ghost` functions which seemed strange to me since they usually return `()` which isn't useful for building `Ghost`s so I changed them to `logic` functions.

I also tried to start laying some groundwork for better trigger support in general. I'm not sure if this is something you would want to add, but I thought check in case.


creusot-contracts/src/logic/ord.rs
```diff
@@ -202,41 +211,7 @@ impl OrdLogic for bool {
         }
     }
 
-    #[ghost]
```
xldenis, line 205: Reminds me we need to change the behavior of default methods so that we don't need to repeat them, they aren't very default today..

xldenis (approved):
I am very favorable to this kind of experimentation; Creusot is a research project after all. This seems like a highly unobtrusive change that won't affect other users so it causes no problem with me. 

However, I'd love to hear the results of your experiments especially when compared to our  no trigger strategy.

jhjourdan:
Sorry, but I'm quite sceptical about this change.

Basically, you are complaining that tools automatically infer triggers for logical function, and, to solve this issue, you create a new tool to infer triggers. Is there a reason why your approach infers better triggers than long-engineered SMT solvers, which know better than us their internal functioning?

I can find at least one concrete reason why this might be a bad idea: if you take the result value as the trigger, then this will prevent solvers to take other triggers. And sometimes, the result value is *not* a good trigger. This is typically the case for lemmas returning `()`. In these cases, we would definitely prefer SMT-inferred triggers.

(Besides, you excluded laws from this new trigger inference system, even though there is no real reason to think that a law necessarily returns `()`. Thus you use the fact of being a law in a rather arbitrary way, IMHO.)

If I put my worries of regressions apart, I wonder whether these new Creusot-infered triggers are not already inferred by SMT solvers. After all, the retrun value of a logical function meets the main criterion for a trigger: it is a subterm with all the quantified variables appearing free, and it is (one of) the smallest of such subterm, since it is simply the logical function symbol applied to these parameters. Do you have concrete problematic examples were this trigger is not infered by SMT solvers?

But don't take me wrong: I *do agree* that the treatment of triggers in Creusot is terrible. But I would prefer opt-in and/or manual mechanisms.

xldenis:
> But don't take me wrong: I do agree that the treatment of triggers in Creusot is terrible. But I would prefer opt-in and/or manual mechanisms.

This _is_ opt-in: if you don't pass a CLI flag nothing happens.

creusot-contracts/src/logic.rs
```diff
@@ -5,8 +5,9 @@ mod fset;
 mod int;
 mod mapping;
 mod ops;
-mod ord;
+pub mod ord;
 mod seq;
+mod seq2;
```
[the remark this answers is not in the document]
dewert99: Sorry, I was also experimenting with a version of `seq` implemented directly in Rust instead of using Why3 so that it would be monomorphized, and forgot to remove it from the commit

dewert99:
> I can find at least one concrete reason why this might be a bad idea: if you take the result value as the trigger, then this will prevent solvers to take other triggers. And sometimes, the result value is not a good trigger. This is typically the case for lemmas returning (). In these cases, we would definitely prefer SMT-inferred triggers.
> 
> (Besides, you excluded laws from this new trigger inference system, even though there is no real reason to think that a law necessarily returns (). Thus you use the fact of being a law in a rather arbitrary way, IMHO.)

Sorry, maybe I misunderstood the purpose of `law`s I thought they were used for "lemmas returning ()". Do they have another purpose?

> Basically, you are complaining that tools automatically infer triggers for logical function, and, to solve this issue, you create a new tool to infer triggers. Is there a reason why your approach infers better triggers than long-engineered SMT solvers, which know better than us their internal functioning?

By the time these axioms reach Why3 they have already lost some information eg:
```rust
fn p(x: Int) -> bool {...}

#[requires(p(x))
#[ensures(result == g(x))]
fn f(x: Int) -> Int {...}
```
and 
```rust
#[ensures(result ==> (f(x) == g(x)))]
fn p(x: Int) -> bool {...}

fn f(x: Int) -> Int {...}
```
will produce the same axiom so solvers wouldn't be able to distinguish between them when generating triggers

dewert99:
I noticed that before my changes most of most implementations of the `OrdLogic` laws were annotated with `#[ghost]` but for `Option<T>` they were annotated with `#[law]` and this seems to affect the generated Why3. Is one way preferred (and should be used everywhere) or do you think I go back to the way things were?

jhjourdan:
> I mean, I definitely agree that if we aren't investing time SMT triggers are better, but I'll note that every single other verification tool picks its own triggers and this is something I've been told to do by SMT developers so I think enabling David to experiment on this subject is very valuable.

Ok, then let's try adding these triggers for lemmas originating from logical functions.

But, could we please deactivate this for functions returning `()`? That's probably a bad idea for all of these functions.

> Sorry, maybe I misunderstood the purpose of laws I thought they were used for "lemmas returning ()". Do they have another purpose?

Yes. Laws are lemmas attached to traits which are automatically brought into the context when the trait is used. I don't think it should have anything to do with triggers.

dewert99:
> Yes. Laws are lemmas attached to traits which are automatically brought into the context when the trait is used. I don't think it should have anything to do with triggers.

That makes sense, but I can still see a connection between this and triggers. Since the triggers are always `result`, they rely on calling the logical function in order to instantiate the axiom. Having a function with `result` as the trigger in the context would only matter if it was being called automatically bringing them into the context wouldn't be necessary. If a user uses the `law` attribute to ask for automatic instantiation I think we can assume they don't want `result` as the trigger.

dewert99:
@xldenis, before this can merged I would like to know how to best handle the `OrdLaw` implementations. I noticed that before my changes most of most implementations of the OrdLogic laws were annotated with #[ghost] but for `Option<T>` they were annotated with #[law] and this seems to affect the generated Why3. Is one way preferred (and should be used everywhere) or do you think I go back to the way things were?

xldenis:
They should all be `#[law]`, but I'm surprised it leads to a difference in the generated code.

dewert99:
Looking at the diffs from the CI failure it looks as though using the `#[ghost]`/`[#logic]` attributes (not using the `#[law]` attribute) on the implementation of a trait law causes the law to not be automatically brought into the context when the concrete type is known.

I'm not sure if this was intended but I wonder if it could be potentially useful for avoiding redundant axioms

xldenis:
> Looking at the diffs from the CI failure it looks as though using the #[ghost]/[#logic] attributes (not using the #[law] attribute) on the implementation of a trait law causes the law to not be automatically brought into the context when the concrete type is known.

In retrospect I wonder if I did that to avoid loading axioms for things like `Int` which have builtin notions... I do think it would be better to have clear consistent rules and add optimizations on top of that rather than the other way around (what is done today).

why3/src/declaration.rs
```diff
@@ -143,6 +143,7 @@ pub enum Attribute {
 #[cfg_attr(feature = "serialize", derive(Serialize, Deserialize))]
 pub struct Signature {
     pub name: Ident,
+    pub trigger: Option<Trigger>, // None means Trigger::single(function_call(self))
```
xldenis, line 146: Is the comment really true? That seems like an awfully strong guarantee about the meaning of inferred triggers.

dewert99: `Option::None` means that the flag --simple-triggers was passed in, the signature is not a `law`, and it doesn't return `()` so we should use `result` as per the description of --simple-triggers (just an optimization generate this expression lazily).

`Some(Trigger::NONE)` means we won't add a trigger and let the tool infer it.

dewert99:
@xldenis, Sorry to keep bothering you but I don't think I'm set up to fix the obsolete proofs, do you think you could do that?

xldenis:
no worries! I've just done that. feel free to merge once the CI is green.

PR #899, opened 2023-10-24
Implemented limited function encoding

dewert99:
Builds on #898 by using the limited function encoding (which seems to come from https://dl.acm.org/doi/10.1145/1529282.1529411) for recursive specification functions when the `simple_triggers` option is enabled. This prevents recursive functions from being unfolded more than once which makes the definition axiom weaker but avoids having it cause a matching loop. This is useful since it can help the solve return "unknown" instead of relying on a timeout, and give it a chance to apply map extensionality axioms.


creusot/src/backend/logic.rs
```diff
@@ -172,12 +173,16 @@ fn body_module<'tcx>(ctx: &mut Why3Generator<'tcx>, def_id: DefId) -> (Module, C
             decls.push(val_decl(ctx, &mut names, def_id));
         } else if body.is_pure() {
             let def_sig = sig.clone();
-            let val = util::item_type(ctx.tcx, def_id).val(sig.clone());
+            let val = ity.val(sig.clone());
             decls.push(Decl::ValDecl(val));
             decls.push(val_decl(ctx, &mut names, def_id));
-            decls.push(Decl::Axiom(definition_axiom(&def_sig, body)));
+            if sig.trigger.is_none() {
```
xldenis, line 179: I'm pretty sure this means it is applied even if `simple_triggers` isn't turned on, no?

creusot/src/backend/logic.rs
```diff
@@ -199,6 +204,49 @@ fn body_module<'tcx>(ctx: &mut Why3Generator<'tcx>, def_id: DefId) -> (Module, C
...
+                Exp::QVar(qname, _) if qname.module.is_empty() && &qname.name == self.0 => {
+                    *exp = Exp::pure_var(self.1.clone())
+                }
+                _ => super_visit_mut(self, exp),
+            }
+        }
+    }
+
+    QNameSubst(name, lim_name).visit_mut(body)
+}
```
xldenis, line 222: This does seem a little sketch, I'm not sure what a better way to do this would be? Would substituting the Pearlite term work?

creusot/src/backend/logic.rs
```diff
@@ -199,6 +204,49 @@ fn body_module<'tcx>(ctx: &mut Why3Generator<'tcx>, def_id: DefId) -> (Module, C
...
+
+fn limited_function_encode(
+    decls: &mut Vec<Decl>,
+    sig: &Signature,
+    contract: &Contract,
+    mut body: Exp,
```
xldenis, line 228: Could you add a comment referencing that paper and mentioning its relation to `simple_triggers`?

creusot/src/backend/logic.rs
xldenis: I don't like this test, is there a way to make it more obvious what would cause this to apply? ie: perhaps `sig.uses_simple_triggers()`?

creusot/src/backend/logic.rs
dewert99: Do you mean that I should move this transformation to the `pearlite::Term` level instead of using `why3::Exp`? That would probably make things more complicated in other ways.

xldenis: I would be curious to know what the complexity is, I want to avoid doing work at the why3 level as much as possible, especially if this is meant to be portable across backends (which I'll start work on a second soon).

dewert99: Function calls in `pearlite::Term` uses `DefId`s for function calls I'm not sure how I could do this other than by generating the limited function at the macro level (having `logic` `ghost` and `predicate` generate an addition Rust function that is used as the limited function)

xldenis: Hmm fair. We'll need to figure out a better solution in the future

dewert99:
Is there anything else I need to do? I'm not sure why the why3 check is falling since the output files didn't change

creusot/src/backend/logic.rs
```diff
@@ -199,6 +204,51 @@ fn body_module<'tcx>(ctx: &mut Why3Generator<'tcx>, def_id: DefId) -> (Module, C
...
+                _ => super_visit_mut(self, exp),
+            }
+        }
+    }
+
+    QNameSubst(name, lim_name).visit_mut(body)
+}
+
+// Use the limited function encoding from https://pm.inf.ethz.ch/publications/HeuleKassiosMuellerSummers12.pdf
+// Originally introduced in https://dl.acm.org/doi/10.1145/1529282.1529411
```
xldenis, line 225: I think a sentence or two of motivation (without deferring to the papers) would be good. Linking the papers is also good.

xldenis (approved):
Looks good, I think 1,2 sentences of explanation of the function/purpose of limited functions would be good. 

Unfortunately, CI is broken atm due to a why3 update so we'll have to wait for my fix to merge.

xldenis:
Go ahead and rebase.

Issue #869, opened 2023-09-26
Causality loop with ghosts and equality


jhjourdan:
This is unsound:

```rust
extern crate creusot_contracts;

use creusot_contracts::*;

#[ensures(false)]
pub fn bad() {
    let mut x = gh! { true };
    let b = &mut x;
    let bg = gh! { b };
    proof_assert! { ***bg == true && *^*bg == true };
    let evil = &mut x;
    proof_assert! { (evil == *bg) == (*^evil == true) }; // Since **evil == true
    *evil = gh! { if evil == *bg { false } else { true } };
    proof_assert! { **evil == !*^evil };
    // Resolve evil
    proof_assert! { **evil == !**evil }; //BOOM !
}
```

xldenis:
The real solution is something we should have done a _long_ time ago. **Ghost could should be Rust code**.

xldenis:
In that world, none of these issues could occur and ghost code would have much clearer and safer semantics. Additional notions of specifications-in-your-code could be added separately from _ghost code_

jhjourdan:
> Additional notions of specifications-in-your-code could be added separately from ghost code

Well but then how do you call that? Won't you have the same soundness problems?

xldenis:
> Well but then how do you call that? 

I don't know

> Won't you have the same soundness problems?

Maybe, but we can worry about that when we can _clearly_ define this not-quite-ghost-specification-code

dewert99:
> The real solution is something we should have done a _long_ time ago. **Ghost could should be Rust code**.

I'm not exactly sure what you mean by this.

Do you mean handling ghosts with trusted specifications?  somewhat like:
```rust
impl<T> Ghost<T> {
  #[ensures(*result = t)]
  fn new(t: T) -> Self;

  #[requires(f.precondition((*self,)))]
  #[ensures(f.postcondition_once((*self,), *result))]
  fn map<U>(self, f: impl FnOnce(T) -> U) -> Ghost<U>;
}
```
This seems like it would be somewhat limiting since values must be consumed to create ghosts and this version is also unsound:

```rust
fn unsound() {
  let g = Ghost::new(());
  let mut x = false;
  let g2 = g.map(|()| *x = true);
  assert!(x); //This would be verified, but panic when actually run
}
```

Or do you mean limiting ghost functions to only all rust expressions inside them? This would involve not allowing equality, quantifiers, and potentially the "logical reborrow/unnesting" (#606) as I'm not totally sure how that works.

[...]

PR #907, opened 2023-11-10
Added MIR optimization to avoid unnecessary re-borrows

dewert99:
In trying to come up with a good way to handle #869 and #902, most of the trouble seems to come from unnecessary reborrowing in the MIR. If it wasn't for that adding a third field to mutable references might be enough to solve these issues while not breaking too many tests (`{ x with current = y }` would preserve the third field but `Borrow.borrow_mut y` would not).
Removing these reborrows also simplifies the Why3 a bit, it probably doesn't make a huge difference to provers, but it might be nice for people using the Why3IDE.

Since I didn't want to mess with the MIR to FMIR transformation I implemented this as a MIR optimization that runs before the analysis so the borrow-checker facts used to determine resolution are accurate. The "optimization" translates statements of the form `_x = &mut *_y` to `_x = move _y` when `_y: &mut _` and `_y` is dead after the statement. I think this should be sound but I'm just making this a draft to see what you think.

### DIFF
--- a/creusot/src/callbacks.rs
+++ b/creusot/src/callbacks.rs
@@ -6,7 +6,7 @@ use rustc_middle::ty::TyCtxt;
 
 use std::{cell::RefCell, collections::HashMap, thread_local};
 
-use crate::{cleanup_spec_closures::*, options::Options};
+use crate::{cleanup_spec_closures::*, options::Options, remove_mir_reborrows::*};
 
 pub struct ToWhy {
     opts: Options,
@@ -32,6 +32,7 @@ impl Callbacks for ToWhy {
                 let mir = (rustc_interface::DEFAULT_QUERY_PROVIDERS.mir_built)(tcx, def_id);
                 let mut mir = mir.steal();
                 cleanup_spec_closures(tcx, def_id.to_def_id(), &mut mir);
+                remove_mir_reborrows(tcx, &mut mir);
                 tcx.alloc_steal_mir(mir)
             };
 
--- a/creusot/src/lib.rs
+++ b/creusot/src/lib.rs
@@ -48,5 +48,6 @@ use translation::*;
 mod error;
 pub(crate) mod lints;
 pub(crate) mod metadata;
+mod remove_mir_reborrows;
 mod translated_item;
 mod validate;
--- a/creusot/src/remove_mir_reborrows.rs
+++ b/creusot/src/remove_mir_reborrows.rs
@@ -1,0 +1,59 @@
+use rustc_ast::Mutability;
+use rustc_middle::{
+    mir::{
+        Body, BorrowKind, Location, Operand, Place, PlaceElem, Rvalue, Statement, StatementKind,
+    },
+    ty::{List, TyCtxt},
+};
+use rustc_mir_dataflow::{
+    impls::MaybeLiveLocals, Analysis, AnalysisDomain, Results, ResultsVisitor,
+};
+
+struct Visitor<'mir, 'tcx> {
+    patch: Vec<(Location, Statement<'tcx>)>,
+    mir: &'mir Body<'tcx>,
+}
+
+impl<'mir, 'tcx> ResultsVisitor<'mir, 'tcx, Results<'tcx, MaybeLiveLocals>>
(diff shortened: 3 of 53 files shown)


xldenis:
> Since I didn't want to mess with the MIR to FMIR transformation I implemented this as a MIR optimization that runs before the analysis so the borrow-checker facts used to determine resolution are accurate. 

The alternative would be to do it as an FMIR -> FMIR pass. 

> In trying to come up with a good way to handle https://github.com/xldenis/creusot/issues/869 and https://github.com/xldenis/creusot/pull/902, most of the trouble seems to come from unnecessary reborrowing in the MIR. If it wasn't for that adding a third field to mutable references might be enough to solve these issues while not breaking too many tests ({ x with current = y } would preserve the third field but Borrow.borrow_mut y would not).

Indeed, but one thing we're trying to balance with this is ensuring that `x` is equivalent to `&mut * x`, I think the current design for non-extensionality has that property.

jhjourdan:
I think that indeed removing these unnecessary reborrows can help simplify the generated MLCFG code and thus the goal in Why3IDE.

That being said, I don't think this helps much for having non-extensional reborrows that would solve the unsoundness in #869. The reason for that is that we want to address simultaneously solve the following goals:


1. **Make borrows non-extensional in a way that fixes the unsoundness.** In particular, this implies that two borrows with two different prophecies should be considered different, even if the *value* of the prophecy is the same. So, if a variable is reborrowed in a way where the initial borrow can be reused, then the reborrow and the initial borrow should not be considered equal, even if their current value and their prophetic values turn out to be equal.

2. **Make reborrows in executable code equivalent to logical reborrowing, when the initial borrow will not be reused.** In particular, this should be true for field reborrows:

```rust
#[ensures(result == &mut x.0)]
fn f(x : &mut (i32, i32)) -> &mut i32 {
  &mut x.0
}
```

I don't think this last example can be handle by a MIR/FMIR optimization, it should really be handle by translating "dead reborrows" in a specific way.

A particularly tricky example is in `15_enumerate.rs`:

```rust
    #[open]
    #[predicate]
    fn completed(&mut self) -> bool {
        pearlite! { self.iter.completed() }
    }

    #[ensures(match result {
      None => self.completed(),
      Some(v) => ...
    })]
    fn next(&mut self) -> Option<Self::Item> {
        match self.iter.next() {
            None => None,
            Some(x) => {
                let n = self.count;
                self.count += 1;
                Some((n, x))
            }
        }
    }
```

Here, in the `None` case, we want the reborrow `&mut self.iter` to be equal to its logical equivalent so that we can use the `completed` logical function in the post-condition. However, the borrow `self` is still used in the `Some` branch.

We could even imagine an example where we reborrow `self.iter` in the `Some` branch. In this case, we would still want the reborrow to be equal to its logical counterpart in the `None` branch, even though *at the time of reborrowing* the parent borrow is not dead.

dewert99:
> The alternative would be to do it as an FMIR -> FMIR pass.

I think doing MIR->MIR was easier since the change is more local (it doesn't need to consider resolves), and it could use the existing MIR liveness analysis.

xldenis:
Perhaps not this exact code, but I think it would probably be worthwhile to validate whether we have improved performance with this, especially if we can whip up a simple PoC version. My biggest worry is that this optimization would fundamentally change which borrows are final, and thus be observable at verification time.

jhjourdan:
(My point being that these unecessaries reborrows that we are removing in this PR would have been considered final afterwards, thus we would essentially have the same behavior.)

xldenis:
> That's a good point that we should take care, although I'm pretty sure that in most of the cases the finality of borrows should not change.

Yes, but we would need to verify that this is true.

jhjourdan:
Is this still relevant given that we now have a third field that works?

jhjourdan:
I'm closing this, because, thanks to final reborrows, we have now fixed the expressiveness issue. The remaining justification is the improvement in performances, but we will do this in a simpler and more general fashion (i.e., final reborrows of an entire place can be replaced with a move).

I am not convinced that this would bring a major change in performances, though.

PR #916, opened 2024-01-03
Turn fmir::Expr into a struct with a separate ExprKind enum

xldenis:
Minor refactoring which enables further work and also opens some follow up clean ups


xldenis:
This causes a bunch of diff because we emit a lot more span information, we can attempt to reduce the useless spans in further PRs if that proves to be of concern but I doubt it.

PR #924, opened 2024-01-08
Prototype `HasDeps` trait

xldenis:
I'm working on removing clones totally. The actual code to replace the current `CloneElaborator` is rather simple and small, however, there is a remaining issue where the `CloneMap` simultaneously gathers dependencies and provides names during translation. This means that a non-clone elaborator cannot change what the format or content of the provided names are since the elaborator is run *after* all dependencies are gathered. By moving dependency calculation to a standalone trait we can split this functionality out of the `CloneMap` making it solely responsible for providing names. 

My current approach aims for this trait to initially have 0 difference with the original code, we can refactor things like the handling of associated types in types later.

### DIFF
--- a/creusot/src/backend/dependency.rs
+++ b/creusot/src/backend/dependency.rs
@@ -1,3 +1,4 @@
+use indexmap::IndexSet;
 use rustc_hir::{def::DefKind, def_id::DefId};
 use rustc_macros::{TypeFoldable, TypeVisitable};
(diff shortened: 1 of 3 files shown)


creusot/src/backend/dependency.rs
```diff
@@ -152,3 +157,130 @@ pub(crate) fn closure_hack<'tcx>(
 
     (def_id, subst)
 }
+
+pub(crate) struct Dependencies<'a, 'tcx>(
+    pub &'a mut Why3Generator<'tcx>,
```
xldenis, line 162: I'd like to ensure that we didn't need this but unfortunately...

PR #926, opened 2024-01-09
Eliminate `clone`s

xldenis:
This PR FINALLY ELIMINATES CLONES. 

Initially, it will exactly copy the behavior of current clones, that is: the generated code will be exactly identical to if Why3 had expanded the clones. Future work will start leveraging the freedom from clones to actually improve the translation (aka remove polymorphism, simplify associated types, support mutual recursion, etc...)

The current code is still very much prototyping the idea, but is promising enough that I think it will get merged. 

There are two observations I am pulling from the code currently:
- The type invariant code should be refactored to produce a `Term` rather than an axiom directly. This opens the door to further refactorings and allows us to easily apply type substitutions.
- We need better handling for 'hacked' identifiers like closure postconditions and type invariants etc... 

Current sketch supports:
- [x] Logical methods
   + [x] Logical methods with contracts
- [x] Type Invariants
  + [x] Trivial Invariants
  + [x] Seq Invariants
  + [x] Tuple Invariants
  + [x] ADT Invariants
  + [x] Borrow Invariants
  + [x] Box Invariants
  + [x] Slice Invariants
  + [x] Array Invariants
- [x] Program functions
- [x] Type Declarations
- [x] Associated Types
- [x] Traits
- [x] Closures


jhjourdan:
Hooray !!!

xldenis:
I now have an initial working version of creusot-without-clones!  Every test generates valid MLCFG which correctly type checks, in the process we've reduced the size of the files by over 2x! Which is honestly surprising. I still need to make sure that the proofs pass, I expect _some_ breakage but also expect certain proofs to go faster: the current code should actually generate tighter contexts (and we can further improve this in the future). 

Here are the remaining tasks:

- [ ] Verify that type invariants work properly
- [ ] Verify that all tests checkout
- [ ] Refactor `logic.rs`
- [ ] Refactor `elaborator.rs`
- [ ] Merge!

xldenis:
This works. I'm going to merge this now and perform simplifications, follow up refactorings and optimizations (of type invariants) in a further PR.

PR #927, opened 2024-01-09
Box borrow resolve

arnaudgolfouse:
The following specification:
```rust
#[ensures(result == *x)]
fn f(x: Box<&mut i32>) -> &mut i32 {
    *x
}
```
Was not provable, because the borrow inside `Box` wasn't resolved. I just added `impl<T: ?Sized> Resolve for Box<T>`, and some `?Sized` bounds to compatible `Resolve` impls.


xldenis:
lgtm

jhjourdan:
That seems good to me too, merging!

PR #933, opened 2024-01-26
Make Why3 CLI Option into subcommand

dewert99:
This PR implements most of the suggestions from #925.
Example usages:
`cargo creusot why3 prove " -P z3 -a split-vc"` (The space before the -P is unfortunately necessary but I'm not sure why)  
`cargo creusot why3 ide`
`cargo creusot why3 replay`

Structurally it required a bit of a hack to allow passing through rustc arguments whether or not the `why3` sub command was used since the sub-command and the rustc arguments need to be last.

I currently report an error whenever `ide` or `replay` is chosen since they can exit successfully without everything actually verifying, although in future I can try and parse `replay`s output and succeed if it reports `n/n`


cargo-creusot/src/main.rs
```diff
@@ -1,12 +1,9 @@
-use cargo_creusot::options::Args;
-use clap::*;
+use creusot_args::options::Args;
 use std::{
     env,
     process::{exit, Command},
 };
 
-mod options;
```
xldenis, line 8: yea thanks for the cleanup here.

xldenis:
The code looks good but i don't understand where the 2k line Cargo.lock diff comes from?

xldenis:
If you do a `git restore --source=master Cargo.lock` and rebuild the changes to `Cargo.lock` are effectively reverted. I've done that and pushed the changes.

dewert99:
> If you do a git restore --source=master Cargo.lock and rebuild the changes to Cargo.lock are effectively reverted. I've done that and pushed the changes.

Thanks, I've added path dependencies to my local rustc in `Cargo.toml` to trick my IDE into understanding rustc imports, but this seems to affect the `Cargo.lock` I've been trying to avoid committing it in this state but this must have slipped through.

dewert99:
You can merge this if you think it is ready

PR #934, opened 2024-01-26
GhostPtrToken Updates

dewert99:
Implement the changes described in my (thesis)[https://dx.doi.org/10.14288/1.0438326] to add ZST alternatives to references to GhostPtrToken, and add missing spec to `take_mut`


creusot-contracts/src/ghost_ptr.rs
```diff
@@ -1,12 +1,26 @@
...
 /// No [`GhostToken`] has permission to a dangling [`GhostPtr`]
 #[trusted]
-pub struct GhostPtrToken<T: ?Sized>(Ghost<FMap<GhostPtr<T>, T>>, PhantomData<T>);
+pub struct GhostPtrToken<T: ?Sized>(PhantomData<T>);
+
+/// ZST equivalent of [`&'a GhostPtrToken<T>`](GhostPtrToken)
+/// Can be created using [`GhostPtrToken::borrow`]
+#[trusted]
+#[derive(Copy, Clone)]
+pub struct GhostPtrTokenRef<'a, T: ?Sized>(PhantomData<&'a T>);
```
xldenis, line 18: clever!

creusot-contracts/src/ghost_ptr.rs
```diff
@@ -140,6 +146,100 @@ impl<T: ?Sized> GhostPtrExt<T> for GhostPtr<T> {
...
+    pub fn shrink_token_ref(self, _new_model: Ghost<FMap<*const T, T>>) -> Self {
+        self
+    }
+}
+
+impl<'a, T: ?Sized> GhostPtrTokenMut<'a, T> {
+    #[trusted]
+    #[ghost]
+    #[open(self)]
+    pub fn curr(self) -> FMap<GhostPtr<T>, T> {
```
xldenis, line 184: For consistency can we name this `cur`, (same as other references to this in Creusot code)

creusot-contracts/src/ghost_ptr.rs
```diff
@@ -140,6 +146,100 @@ impl<T: ?Sized> GhostPtrExt<T> for GhostPtr<T> {
...
+        absurd
+    }
+
+    #[ensures(self.fin() == self.curr())]
+    #[ensures(result@ == self.curr())]
+    pub fn shr(self) -> GhostPtrTokenRef<'a, T> {
+        GhostPtrTokenRef(PhantomData)
+    }
+
+    /// Mutably borrows `ptr` and shrinks `t` so that it can no longer be used to access `ptr`
```
xldenis, line 201: I `t` should be `self` here?

xldenis: Also, given that this is the most complex function in the API by far, do you think you could include a simple example usage? Each time I read this it takes me a few minutes to understand how it works again.

xldenis (approved):
I like the new `Ref` types which preserve ZST-ness even for borrows!

dewert99:
I implemented your suggestions and decided to turn my example into a test case. Unfortunately, it seems like Creusot currently leaves out necessary "use prelude.Opaque" imports when using raw pointers. I artificially added this import to the expected mlcfg which is why CI is failing, but doing this (along with adding some `inline-trival` hints to `FMap` there may be a separate issue relating to how transparent non-inlined functions are handled) allows why3 to verify the example.

Would it make sense to merge a separate PR to fix the "prelude.Opaque" issue first?

xldenis (approved):
Thanks, I really appreciate the added test case, that should make things much clearer. However, could you add a proof session so that CI checks that the obligations hold? Otherwise it will only typecheck the `mlcfg`.

dewert99:
Sorry I think I still don't understand the `why3` test, I tried generating a session locally using
 `why3 ide -C ci/why.conf -L prelude creusot/tests/should_succeed/ghost_ptr_token.mlcfg` after running the ci github installation script:
```
wget https://github.com/Z3Prover/z3/releases/download/z3-4.12.4/z3-4.12.4-x64-glibc-2.35.zip
unzip z3-4.12.4-x64-glibc-2.35.zip
sudo cp -rn z3-4.12.4-x64-glibc-2.35/bin /usr/local
sudo cp -rn z3-4.12.4-x64-glibc-2.35/include /usr/local
sudo chmod +x /usr/local/bin/z3
```
I also ran `why3 replay -C ci/why.conf -L prelude creusot/tests/should_succeed/ghost_ptr_token` which reported `1/1 (replay OK)`, so I'm not sure why3 the ci test is failing

xldenis:
you can normally rebuild the test locally by running `cargo test --test why3` and it will "do the thing".

dewert99:
I tried `WHY3_CONFIG=../ci/why.conf cargo test --test why3 -- "ghost_ptr_token"` which reported:
```
Testing ../creusot/tests/should_succeed/ghost_ptr_token.mlcfg ... replayed
Success!
```
and didn't change the `why3session.xml` or the `why3shapes.gz` files, so I'm not sure what I'm doing wrong.

xldenis:
huh, i'm not sure either, but locally it did update the session slightly (seems to only have changed proof times though?) I wonder if it's related to the 'shapes' file

dewert99:
Thanks

PR #935, opened 2024-02-02
Bump Why3, update sessions, harden some timings.

jhjourdan:


jhjourdan:
It seems like there is a problem with CVC5: most of the failing timings on my machine were from goals solved with CVC5, and running CVC5 from the IDE in parallel with other provers results in unreasonable timings. I hope this is not a bug somewhere in Why3...

PR #937, opened 2024-02-02
Update `rustc` to late october

xldenis:
This brings us up to late october, were it not for one remaining bug in the metadata (de)serializaiton logic. I took a look at this earlier today, but haven't cracked it yet. It's something related to the serialization of span data. idk if you have ideas @jhjourdan (relevant test file is vecdeque). Otherwise I'll look into this on monday.


xldenis:
Ok I took a look at this issue. As far as I can tell `rustc` became stricter with regards to trait bounds on functions. In particular the following bound which is a sort of 'eta expansion' crashes:
```
          Option<T> :   Fake<FakeTy = <Option<T> as Fake>::FakeTy>,
```
whereas before we were at least able to use contexts including this bound. 

This bound occurs from the external contracts we give to trait items like `eq`, so I think we might be a ble to eliminate it by being a little smarter when constructing the extended parameter environments for these items. If we attempt to normalize & discharge obligations before adding them to the environment this may solve the issue.

jhjourdan:
I was able to reproduce the issue, but I don't see why you think this is elated to (de)serialization of *spans*.

xldenis:
> This is a breaking change in rustc, right? Perhaps worth reporting?

No it's not as it is already rejected by `rustc`, it has to do with how we build `ParamEnv`s for methods with `extern_specs`.

xldenis:
> I was able to reproduce the issue, but I don't see why you think this is elated to (de)serialization of spans.

I thought it was originally, but i think that was fatigue on my part.

PR #939, opened 2024-02-08
Update rustc to February

xldenis:
Brings us up to date. 

Mostly some renamings, and a few methods were moved around in the metadata stuff again.


xldenis:
This was one of the more painful upgrades of recent memory but I think it broadly puts us in a better position in the future: I think the hooks / assumptions we make of rustc are less contingent. 

The two problem areas were:
1. Rustc tightened the checks around valid MIR causing our ghost closures to error out. The fix was to erase the closures from the MIR entirely at the end of the pipeline after Creusot has already done its analysis. This allows MIR to proceed unimpeded. I think this approach makes things less likely to break in the future. 
2. Metadata Encoding changes. There were two things which happened here: `rustc` stopped using specialization for this, which is overall a good and simple syntactic change. However, some other refactorings to `EncodedSourceFileId` broke a bunch of things. More generally, we based `creusot-metadata` of the implementation of `on_disk_cache` which is stable across invocations but not really meant to be "shared" across projects. I think if we refactored the code to be based of the `rmeta` encoder it would be better. I fixed the code but I think some more maintenance will be needed in the future.

PR #941, opened 2024-02-14
Add a chapter on type invariants to the guide

voidc:
The chapter is not super long but should cover the most important bits. For an in-depth explanation users can always refer to my thesis :smile: 

[Rendered](https://github.com/xldenis/creusot/blob/tyinv-doc/guide/src/type_invariants.md)


xldenis:
Thanks Dominik! I think you should add a link somewhere to your thesis, maybe at the start:

_This work was added as part of [..](thesis-link)_

Issue #944, opened 2024-02-20
relative locations/spans are broken when using `cargo creusot`


Armael:
When using `cargo creusot` the spans/locations are currently broken:
- spans using relative paths are the default (which indeed seems necessary to allow committing mlcfgs into git)
- spans use relative paths that assume that the `.rs` file and related `.mlcfg` file are next to each other
- this assumption holds for the creusot testsuite (AFAICT the test harness calls creusot-rustc directly, in a way where the .mlcfg ends up next to the .rs file)
- but it *does not* hold for `cargo creusot` invocations, which generate the .mlcfg files in `target/debug/`.

I'm not sure how to properly fix the issue:
- adding `../../` to the relative paths would fix it for `cargo creusot` but break the testsuite
- it's also not clear that `target/debug` where we want the .mlcfg to be; if we want to check them into git then they should probably be somewhere else (I assume `target/` is usually in .gitignore). 
- maybe the relative paths could assume that we're in a `cargo creusot` setup, and additionally add an option to creusot-rustc to explicitly specify the relative path of the .rs file wrt the .mlcfg file, that the testsuite harness would use.

Any ideas?
Provided we agree on the design for the solution I'm happy to try and implement it.

xldenis:
> maybe the relative paths could assume that we're in a cargo creusot setup, and additionally add an option to creusot-rustc to explicitly specify the relative path of the .rs file wrt the .mlcfg file, that the testsuite harness would use.

I think something like this is reasonable to provide a 'relative from' option.

PR #947, opened 2024-02-20
fix: emit correct relative paths in spans with 'cargo creusot'

Armael:
This is an attempt at fixing #944 , along the lines mentioned in the issue.

This PR adds an option to creusot-rustc to specify the relative location of the project's root with respect to creusot's output. This option is used by the test harness, which uses a different file layout than 'cargo creusot'. ('cargo creusot' then simply uses the default value of the option, which matches its usecase: the default value is `../..` to match the `target/debug` directories where `cargo creusot` stores generated mlcfg files.)

### DIFF
--- a/creusot-args/src/options.rs
+++ b/creusot-args/src/options.rs
@@ -1,6 +1,6 @@
 use clap::*;
 use serde::{Deserialize, Serialize};
-use std::{error::Error, ffi::OsString};
+use std::{error::Error, ffi::OsString, path::PathBuf};
 
 #[derive(Parser, Serialize, Deserialize)]
 pub struct CreusotArgs {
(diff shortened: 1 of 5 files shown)


jhjourdan:
> the default value is ../.. to match the target/debug directories where cargo creusot stores generated mlcfg files.)

Really? I would have expected something like `../../src` since by default the source files are in `$ROOT/src`.

Armael:
It looks like we already get the `src` part somehow, so things work out (maybe from the rest of the intricate logic in `Options::relative_to_output`, which I don't really understand?)

xldenis:
> It looks like we already get the src part somehow, so things work out (maybe from the rest of the intricate logic in Options::relative_to_output, which I don't really understand?)

Rust considers all paths relative to the crate root if i recall, which means that we do get the `src` bit 'for free'.

xldenis (approved):
go ahead and merge

PR #951, opened 2024-02-21
Demote `Expr::Call` to `Statement::Call`

xldenis:
This PR reifies a data invariant in the types, namely that function calls are always at the 'head' level. 
This will be useful when switching to coma as it simplifies the handling of control flow required. 

I also performed some very *minor* cleanups of `fmir` types, but much more work is required on that front. 

A nice side-effect is that we visibly fix some spans related to `extern_specs`.

### DIFF
--- a/creusot/src/translation.rs
+++ b/creusot/src/translation.rs
@@ -100,7 +100,7 @@ pub(crate) fn after_analysis(ctx: TranslationCtx) -> Result<(), Box<dyn Error>> 
                 let outputs = why3.output_filenames(());
                 let crate_name = why3.crate_name(LOCAL_CRATE);
 
-                let libname = format!("{}-{}.mlcfg", crate_name.as_str(), why3.crate_types()[0]);
+                let libname = format!("{}-{}.coma", crate_name.as_str(), why3.crate_types()[0]);
 
                 let directory = if why3.opts.in_cargo {
                     let mut dir = outputs.out_directory.clone();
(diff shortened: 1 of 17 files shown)


creusot/src/translation.rs
```diff
@@ -100,7 +100,7 @@ pub(crate) fn after_analysis(ctx: TranslationCtx) -> Result<(), Box<dyn Error>>
                 let outputs = why3.output_filenames(());
                 let crate_name = why3.crate_name(LOCAL_CRATE);
 
-                let libname = format!("{}-{}.mlcfg", crate_name.as_str(), why3.crate_types()[0]);
+                let libname = format!("{}-{}.coma", crate_name.as_str(), why3.crate_types()[0]);
```
jhjourdan, line 103: This has nothing to do with this PR.

why3/src/coma.rs
```diff
@@ -235,16 +235,6 @@ impl Print for Expr {
...
-fn braces<'a, A: pretty::DocAllocator<'a>>(
-    doc: pretty::DocBuilder<'a, A>,
-) -> pretty::DocBuilder<'a, A> {
-    if !matches!(&*doc.1, pretty::Doc::Nil) {
-        doc.braces()
-    } else {
-        doc
-    }
-}
-
```
jhjourdan, line 247: This has nothing to do with this PR.

creusot/src/translation.rs
xldenis: oh that's a good one, I extracted this from another branch.

why3/src/coma.rs
xldenis: Indeed but it got rid of a warning so I left it in.

PR #959, opened 2024-03-01
Merge `Expr` into `RValue`

xldenis:
This finally removes the `Expr` type by combining it with `RValue`. 

There is one remaining minor issue with the translation in `program.rs`: RValues translate to a single why3 `Exp` while borrows must be translated to several mlcfg statements. Either the type of `RValue::to_why` should be changed or maybe `Borrow` should be made a full statement instead?


xldenis:
This also removes some duplicate spans around each assignment.

PR #961, opened 2024-03-02
Add 'cargo creusot setup' to manage creusot installations

Armael:
The implementation follows the design outlined here: https://github.com/creusot-rs/creusot/discussions/946 .
Note that this the first time I'm writing a non-trivial amount of Rust code, so I'm happy to take any comments regarding code style, organization, etc.

`cargo creusot setup` comes with the following subcommands:
- `status`: displays the state of the current installation
- `install`: sets up creusot in "managed" mode, downloading solver binaries at known-good versions and configuring why3 to use them. (Can also be invoked on an existing creusot installation to upgrade out-of-date solvers.)
- `install-external`: minimal setup of creusot in "external" mode, that assumes that the user manages the external tools (why3, solvers, etc) themselves.

For now, even the "managed mode" setup assumes that why3 and alt-ergo have been installed separately (e.g. using opam), but the goal is to later have 'creusot setup' download binaries when we have them.


Armael:
Since this changes a bit the CLI workflow (now you *have* to go through the setup phase even if it's just for `install-external`) I probably need to update the CI script and the README. Will do later.
(Regarding the CI, I've used the solver versions that the CI currently uses, and the idea is that the much of the CI script could be replaced by a `cargo creusot setup install` invocation).

cargo-creusot/src/setup/config.rs
```diff
@@ -0,0 +1,112 @@
...
+
+impl Config {
+    pub fn read_from_file(p: &Path) -> Result<Self, Error> {
+        if !p.is_file() {
+            return Err(Error::NotFound);
+        };
+        let s = match fs::read_to_string(p) {
+            Err(e) => return Err(Error::Invalid(e.to_string())),
+            Ok(s) => s,
+        };
```
xldenis, line 55: If we want more ergonomic error handling here we could look at `eyre` or more generally return a `Box<dyn Error>` object as the error case. Of you could define an instance of `From` for the appropriate types. 

It doesn't matter much here, but in case you wanted comments for future changes.

xldenis: Ah I see you're using `anyhow` down below.

cargo-creusot/src/setup/mod.rs
```diff
@@ -0,0 +1,311 @@
...
+}
+
+pub struct CreusotFlags {
+    pub why3_path: PathBuf,
+    pub why3_config: Option<PathBuf>,
+}
+
+// compute the flags to pass to creusot-rustc.
+// fail if the installation is not in an acceptable state, which means we will
+// stop there and do not attempt launching creusot-rustc.
```
xldenis, line 151: If you want these to count as documentation you can use `///` to make them doc comments.

HACKING.md
```diff
@@ -0,0 +1,54 @@
...
+  configuration.
+
+## Running the testsuite
+
+- Test the output of creusot (mlcfg files) against reference files:
+```
+cargo test --test ui
+```
+
+TODO: how to update an out-of-date reference file?
```
xldenis, line 42: `cargo test --test ui -- "optional-string" --bless`

Armael:
I've updated the PR. There is quite a bit of new code: I've been working on the developer workflow (for people hacking on creusot, running the testsuite, etc) and tried to harmonize it with the "user" workflow introduced by `cargo creusot setup`.

The idea is that someone who has a working creusot setup (using `cargo creusot setup`) should be able to start hacking on creusot and use it as configuration also for the testsuite. And alternatively it should still be possible to use a custom why3 or solver when needed.
The solution I propose (and implement here) is documented in [HACKING.md](https://github.com/Armael/creusot/blob/setup-cmd/HACKING.md#setup): by default the testsuite will use the global config from `cargo creusot setup`, but if you want to do something different, you can have a `.creusot-config` directory at the root of the git repository, and that will be used instead (it is ignored by git). (A sample configuration is provided in `.creusot-config.sample`.)

I've also updated the user setup instructions in README.md to use the new `cargo creusot setup install` command.

All in all this is a bit more logic/code than I initially expected, but this is what I came up with trying to make things more robust.

xldenis:
> The solution I propose (and implement here) is documented in [HACKING.md](https://github.com/Armael/creusot/blob/setup-cmd/HACKING.md#setup): by default the testsuite will use the global config from cargo creusot setup, but if you want to do something different, you can have a .creusot-config directory at the root of the git repository, and that will be used instead (it is ignored by git). (A sample configuration is provided in .creusot-config.sample.)

I think this is a good workflow, most of the time we are not working against custom solver versions or custom why3 versions. I also think it drastically simplifies the new contributor experience which is valuable as well.

xldenis:
I've been reading over the code today, it all looks really good to me, and I can't wait to try it out.

xldenis (approved):

HACKING.md
```diff
@@ -0,0 +1,54 @@
...
+
+- Replay proofs:
+```
+cargo test --test why3
```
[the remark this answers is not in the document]
Armael: I just added those, thanks!

xldenis:
🎉

PR #963, opened 2024-03-05
Add a ghost! block for ghost code with ownership

arnaudgolfouse:


creusot-contracts-dummy/src/lib.rs
```diff
@@ -35,7 +35,12 @@ pub fn snapshot(_: TS1) -> TS1 {
...
 #[proc_macro]
 pub fn ghost(_: TS1) -> TS1 {
-    quote::quote! { creusot_contracts::ghost::Ghost::from_fn(|| std::process::abort()) }.into()
+    quote::quote! { () }.into()
+}
+
+#[proc_macro]
+pub fn gh(_: TS1) -> TS1 {
+    quote::quote! { creusot_contracts::ghost::GhostBox::from_fn(|| std::process::abort())}.into()
 }
```
xldenis, line 44: If `gh` is just an alias for ghost I think you can define it only once in the leaf crate by doing `pub use ghost as gh`.

creusot/src/backend/program.rs
```diff
@@ -145,62 +145,85 @@ pub(crate) fn translate_function<'tcx, 'sess>(
...
+            let body_id = BodyId { def_id: def_id.expect_local(), promoted: None };
+            let mut coma_defn = to_why(ctx, &mut names, body_id);
+
+            // hack: to avoid having a second copy of the closure
+            names.hidden.insert(Dependency::new(tcx, (def_id, subst)));
+            // get the right name for the generated function
+            coma_defn.name = names.insert(Dependency::new(tcx, (def_id, subst))).ident();
+            Decl::Coma(coma_defn)
+        })
+        .collect::<Vec<_>>();
```
xldenis, line 180: This code needs to be cleaned up.

arnaudgolfouse: That `hidden` hack seemed weird to me, does it really need to exists or is there another way to avoid duplication of the closure definition ?

creusot-contracts-dummy/src/lib.rs
arnaudgolfouse: It does a bit more, right now it's syntax sugar: `gh!(e) := ghost!(GhostBox::new(e))`

xldenis:
> So there are surely more things to review, but overall the functionalities are there, so I'm un-drafting this PR. Fell free to play with it/suggest improvements!

Are you asking for code reviews?

creusot-contracts-dummy/src/lib.rs
```diff
@@ -33,6 +33,17 @@ pub fn snapshot(_: TS1) -> TS1 {
         .into()
 }
 
+#[proc_macro]
+pub fn ghost(_: TS1) -> TS1 {
+    // FIXME: this does not accept returning the unit type...
```
xldenis, line 38: ? I don't understand what this means.

creusot-contracts/src/lib.rs
```diff
@@ -156,9 +174,38 @@ pub mod std;
...
+    }
+
+    impl<T: ?Sized> Clone for GhostBox<T> {
+        fn clone(&self) -> Self {
+            GhostBox(std::marker::PhantomData)
+        }
+    }
+
+    impl<T: ?Sized> Copy for GhostBox<T> {}
+}
```
xldenis, line 207: I think we should move this outside of `lib.rs`, either into the existing `ghost.rs` or into a `ghost_no_creusot.rs` or something.

creusot/src/backend/program.rs
```diff
@@ -148,56 +147,84 @@ pub(crate) fn translate_function<'tcx, 'sess>(
...
+        .map(|body_id| Decl::Coma(to_why(ctx, &mut names, *body_id)))
+        .collect::<Vec<_>>();
+
+    let tcx = ctx.tcx;
+    let ghost_closure_ids: IndexSet<_> = names.find_ghost_closures(tcx).collect();
+    let ghost_closures = ghost_closure_ids
+        .into_iter()
+        .map(|(def_id, subst)| {
+            // Compile the closure normally
+            closure_aux_defs(ctx, def_id);
```
xldenis, line 171: this only is required if ghost closures are not compiled ordinarily, is that still the case?

creusot/src/translation/function/terminator.rs
```diff
@@ -71,16 +75,104 @@ impl<'tcx> BodyTranslator<'_, 'tcx> {
...
+                                )
+                                .may_apply();
+                            let ghost_in_caller = self.is_ghost_box(caller_ty);
+                            let ghost_in_callee = self.is_ghost_box(ty);
+                            if deref_in_callee && ghost_in_caller && !ghost_in_callee {
+                                self.ctx.error(*fn_span, &format!("Cannot instantiate a generic type {ty} implementing `Deref` with the ghost type {caller_ty}")).emit();
+                            }
+                        }
+                    }
+                }
```
xldenis, line 174: Can we refactor  this into  helper function(s)?

xldenis (commented):
Overall it looks good, I do have some comments I would like to see addressed though:

1. I would like the hack in `program.rs` to be fully removed in favor of the changes to `elaborator.rs`
2. In `terminator.rs` the could should be refactored to use additional helpers rather than a 100 line if branch.
3. I would like to see a test case which actually uses the new ghost code in some useful / novel way.

creusot-contracts-dummy/src/lib.rs
arnaudgolfouse: A ghost block is supposed to return either `GhostBox<T>` or `()`. In creusot we handle it during translation to FMIR, but we also need this to be true for normal code ; but I did not find a way to do this properly. So in program code only `GhostBox<T>` can be returned...

creusot/src/translation/function.rs
```diff
@@ -59,44 +56,82 @@ pub struct BodyTranslator<'body, 'tcx> {
...
-    // Spec / Snapshot variables
+    /// Spec / Snapshot variables
     erased_locals: BitSet<Local>,
 
-    // Current block being generated
+    /// Current block being generated
     current_block: (Vec<fmir::Statement<'tcx>>, Option<fmir::Terminator<'tcx>>),
 
-    past_blocks: IndexMap<BasicBlock, fmir::Block<'tcx>>,
+    past_blocks: &'temp mut IndexMap<BasicBlock, fmir::Block<'tcx>>,
```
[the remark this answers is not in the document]
arnaudgolfouse: Ah, that's a leftover from a previous version, when ghost closures were fully inlined in FMIR. I removed all `&'temp mut` fields here, thanks!

creusot/src/backend/program.rs
```diff
@@ -148,56 +147,84 @@ pub(crate) fn translate_function<'tcx, 'sess>(
...
+            let mut coma_defn =
+                to_why(ctx, &mut names, BodyId { def_id: def_id.expect_local(), promoted: None });
+
+            names.insert_hidden_func(def_id, subst);
+            // get the right name for the generated function
+            coma_defn.name = names.insert(Dependency::new(tcx, (def_id, subst))).ident();
```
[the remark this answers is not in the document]
arnaudgolfouse: Indeed, this is now removed

creusot/src/backend/program.rs
arnaudgolfouse: Indeed this is not the case, this is now removed

creusot/src/translation/function/terminator.rs
arnaudgolfouse: Done

creusot-contracts-dummy/src/lib.rs
arnaudgolfouse: We talked a bit with @jhjourdan, and he convinced me that:
- There should only be one macro (I tend to prefer `ghost!`)
- This macro always automatically wraps its result in a `GhostBox`

This is a more minimalistic design, the main drawbacks being that
- a ghost block cannot return unit (it returns `GhostBox<()>` instead), so sometimes a `;` must be appended at the end of the block
- we need an additional dereference in the case where the returned value is _already_ a `GhostBox`:
  ```rust
  let g: GhostBox<i32> = ...;
  let g2: GhostBox<i32> = ghost! {
      *g
  };
  ```
Overall I think these are acceptable, so I'll implement this change

creusot/src/backend/clone_map.rs
```diff
@@ -469,7 +469,7 @@ pub enum Kind {
 }
 
 impl Kind {
-    fn ident(&self) -> Ident {
+    pub(crate) fn ident(&self) -> Ident {
```
xldenis, line 472: Does this still need to be `pub(crate)` now that you've removed the hacks in `program.rs`?

xldenis:
Ok this looks good to me now! If you want to rebase, I think we can merge.

arnaudgolfouse:
> Ok this looks good to me now! If you want to rebase, I think we can merge.

I think @jhjourdan wanted to have a look at this first (and I would like him to as well 😁)

creusot/src/backend/clone_map.rs
arnaudgolfouse: Right, thanks!

xldenis:
It would be a good idea to try and merge this asap, @jhjourdan @arnaudgolfouse before it becomes difficult to rebase again. However, I'm planning on release 0.2 without it anyways.

creusot-contracts/src/ghost.rs
```diff
@@ -0,0 +1,86 @@
...
+/// ```compile_fail
+/// let b: GhostBox<i32> = GhostBox::new(1);
+/// ghost! {
+///     let value: i32 = *b;
+///     // use value here
+/// }
+/// let value: i32 = *b; // compile error !
+/// ```
+#[rustc_diagnostic_item = "ghost_box"]
+pub struct GhostBox<T: ?Sized>(Box<T>);
```
jhjourdan, line 25: Perhaps we could make this `Clone` and `Copy` as soon as `T` is.

jhjourdan (approved):
A few nitpicks, but all in all I think this is ready to merge!

jhjourdan:
I trust @xldenis for the changes to the backend, which I lack expertise to detect defects.

xldenis:
The backend looks good to me.

creusot-contracts/src/ghost.rs
```diff
@@ -0,0 +1,86 @@
...
+        Self(Box::new(x))
+    }
+
+    #[trusted]
+    #[pure]
+    #[requires(true)]
+    #[ensures(true)]
+    pub fn uninit() -> Self {
+        loop {}
+    }
```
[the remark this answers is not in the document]
arnaudgolfouse: At one point I thought it would be useful to conjure ghost objects from thin air. Thinking more about this, it now seems like a very bad idea, not only because of `!`, but also because we might want to use some kinds of unique ghost tokens. I removed it.

creusot-contracts-dummy/src/lib.rs
```diff
@@ -33,6 +33,11 @@ pub fn snapshot(_: TS1) -> TS1 {
         .into()
 }
 
+#[proc_macro]
+pub fn ghost(_: TS1) -> TS1 {
+    quote::quote! { GhostBox::from_fn(|| { loop {} }) }.into()
+}
```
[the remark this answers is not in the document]
arnaudgolfouse: Yes I just implemented your second point. It does add a dependency on `proc_macro2` for `creusot-contracts-dummy` (which was already transitively pulled in anyways).

creusot-contracts/src/ghost.rs
arnaudgolfouse: Agreed

creusot-contracts/src/lib.rs
```diff
@@ -156,9 +173,38 @@ pub mod std;
...
+        }
+    }
+
+    impl<T: ?Sized> Clone for GhostBox<T> {
+        fn clone(&self) -> Self {
+            GhostBox(std::marker::PhantomData)
+        }
+    }
+
+    impl<T: ?Sized> Copy for GhostBox<T> {}
```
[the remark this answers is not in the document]
jhjourdan: ?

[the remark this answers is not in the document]
arnaudgolfouse: Ah yes, I thought that was conditioned on `T: Copy`, I'll put this bound to be coherent with Creusot mode.

PR #966, opened 2024-03-09
setup: slight refactoring of the why3 config file generation

Armael:
This passes the destination path of the file to generate to why3, instead of passing it a fake HOME environment variable.
I also renamed the file from `.why3.conf` to `why3.conf` to make it harder to miss when trying to debug an installation.
(@xldenis when updating to this PR you'll probably want to erase the previous `.why3.conf` and redo `setup install` just to be safe)

### DIFF
--- a/creusot-setup/src/lib.rs
+++ b/creusot-setup/src/lib.rs
@@ -13,9 +13,11 @@ use tools::*;
 use tools_versions_urls::*;
 use ToolsConfig::*;
 
+// CAUTION: on MacOS, [config_dir] and [data_dir] are in fact the same directory
 struct CfgPaths {
     config_dir: PathBuf,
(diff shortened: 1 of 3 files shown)


creusot-setup/src/lib.rs
```diff
@@ -32,6 +34,7 @@ fn get_config_paths(custom_config_dir: &Option<PathBuf>) -> anyhow::Result<CfgPa
     Ok(CfgPaths {
         config_dir: PathBuf::from(config_dir),
         config_file: config_dir.join("Config.toml"),
+        why3_config_file: config_dir.join("why3.conf"),
```
xldenis, line 37: I think that's a good idea, no need to hide it inside our directory

xldenis (approved):

PR #971, opened 2024-03-15
Non Prusti specific changes that are helpful for adding Prusi-ish syntax

dewert99:
In #932 you mentioned it would be useful to extract out some of the changes I made that are not Prusti specific into there own PR.

I reorganized `creusot_contracts` adding `base_macros` which is just a re-export of either `creusot_contracts_proc` or `creusot_contracts_dummy`, and `prelude` which re-exports the items from `creusot_contracts` but not all the modules so that you can `use creusot_contracts::prelude::*` instead of `use creusot_contracts::*`.

I modified the types stored in pearlite `Term`s so that they match the types from THIR, instead of stripping off boxes and shared references. I also added a `creusot_ty` method that returns the type with boxes and shared references removed to use in places creusot was relying on this stripping.

### DIFF
--- a/creusot-contracts/src/lib.rs
+++ b/creusot-contracts/src/lib.rs
@@ -13,121 +13,28 @@
 extern crate self as creusot_contracts;
 
 #[cfg(creusot)]
-mod macros {
-    /// A pre-condition of a function or trait item
-    pub use creusot_contracts_proc::requires;
-
-    /// A post-condition of a function or trait item
-    pub use creusot_contracts_proc::ensures;
-
-    pub use creusot_contracts_proc::snapshot;
-
-    /// A loop invariant
-    /// The first argument should be a name for the invariant
-    /// The second argument is the Pearlite expression for the loop invariant
-    pub use creusot_contracts_proc::invariant;
-
-    /// Declares a trait item as being a law which is autoloaded as soon another
-    /// trait item is used in a function
-    pub use creusot_contracts_proc::law;
-
-    /// Declare a function as being a logical function, this declaration must be pure and
-    /// total. It cannot be called from Rust programs, but in exchange it can use logical
-    /// operations and syntax with the help of the [`pearlite!`] macro.
-    ///
-    /// # `prophetic`
-    ///
-    /// If you wish to use the `^` operator on mutable borrows to get the final value, you need to
-    /// specify that the function is _prophetic_, like so:
-    /// ```ignore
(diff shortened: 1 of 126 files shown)


creusot-contracts/src/lib.rs
```diff
@@ -13,121 +13,28 @@
...
-    pub use creusot_contracts_proc::open;
-
-    pub use creusot_contracts_proc::DeepModel;
-
-    pub use creusot_contracts_proc::Resolve;
-}
+extern crate creusot_contracts_proc as base_macros;
 
 #[cfg(not(creusot))]
+extern crate creusot_contracts_dummy as base_macros;
```
xldenis, line 19: Big fan of this '

creusot/src/translation/pearlite.rs
```diff
@@ -156,6 +156,38 @@ pub enum TermKind<'tcx> {
...
+                let subst = GenericArgs::for_item(tcx, def_id, |x, _| {
+                    let s = subst[x.index as usize];
+                    let us = u_subst.args[x.index as usize];
+                    if us.has_escaping_bound_vars() {
+                        s
+                    } else {
+                        us
+                    }
+                });
+                Self::Item(def_id, subst)
```
xldenis, line 185: Can you explain why this is needed?

xldenis (approved):
> I modified the types stored in pearlite Terms so that they match the types from THIR, instead of stripping off boxes and shared references.

This is something I wanted to do anyways it makes the interpretation of terms well founded.

I think the `creusot_ty` method is good but longer term I'd like tens to be fully well typed and not elude those references.

creusot/src/translation/pearlite.rs
dewert99: The types in THIR have all of the lifetimes erased, this is normally fine since the Prusti translation/type-checker can get the argument and return types with lifetimes and infer the rest. The issue the translation can't see lifetimes parameters that were manually instantiated by the user. This change replaces the erased lifetimes with the ones manually instantiated by the use, or leaves them erased if they weren't manually instantiated.

The translation currently only uses this information to extract the lifetime from `at::<'x>` expressions, but eventually it could also error if the instantiated lifetime of a `logic` function doesn't match what it inferred.

dewert99:
Sorry, when you approve changes do you mean for me to merge them, or do you prefer to do that yourself?

xldenis:
No go ahead and merge yourself.

PR #973, opened 2024-03-20
Pin Why3's version to a specific hash & refactor CI

Armael:
This PR does a number of related things to improve the setup workflow for users:
- introduce a `creusot-deps.opam` file, to allow creating the required opam switch with one command
- in the opam file, pin why3 to a specific "known good" commit, so that users don't pull an unknown version on why3 at setup
- refactor the CI workflows to:
  - use the opam file with the "known good" commit on master/PRs
  - have an extra nightly workflow to test creusot against why3 master

The idea is then that upgrading the "known good" why3 commit is a manual operation (just editing the opam file), that I documented in HACKING.md.


xldenis:
Yep, this looks good to me.

PR #975, opened 2024-03-20
Guide with mdbook

arnaudgolfouse:
This moves the user documentation out of the README (which was starting to be a bit busy IMO) an into the already existing mdbook.
I also filled some sections, but a lot more remains to be done.

It would also be cool if we could check that the examples are indeed provable (or even compilable) in CI.


xldenis:
>  I also filled some sections, but a lot more remains to be done.

Thanks for starting the work!

I think we can merge this PR but here are two ideas for the future of README:

- Include a BibTex entry for easy citing, also probably put this lower down.
- Replace the "Examples of verification" with an inline simple example. This could be useful for new users to use a smoke test additionally.

PR #983, opened 2024-03-29
strip trailing whitespace from why3 version string

### LINKED ISSUE #982: creusot-setup: Why3 version checking needs to `trim` whitespace
On my host, within `tools::detect_why3_version`:

```
(rr) list
166	fn detect_why3_version(why3: &Path) -> Option<String> {
167	    let output = Command::new(&why3).arg("--version").output().ok()?;
168	    let version_full = String::from_utf8(output.stdout).ok()?;
169	    let version = version_full.strip_prefix("Why3 platform, version ");
170	    version.map(|ver| {
171	        let parts: Vec<_> = ver.split(|c| c == '.' || c == '+').collect();
172	        String::from(&parts[..3].join("."))
173	    })
174	}
(rr) p version_full
$26 = "Why3 platform, version 1.7.1\n"
(rr)
```

which ends up getting stripped down to `"1.7.1\n"`

The problem:

```
(rr) list
155	pub fn detect_binary_version(bin: &ExtBinary, path: &Path) -> DetectedVersion {
156	    let detect_version = bin.detect_version;
157	    match detect_version(&path) {
158	        None => DetectedVersion::Bad(None),
159	        Some(ver) if ver != bin.version => DetectedVersion::Bad(Some(ver)),
160	        Some(_) => DetectedVersion::Good,
161	    }
(rr) p bin
$29 = (*mut creusot_setup::tools::ExtBinary) 0x7fff3aa074b0
(rr) p bin.version
$30 = "1.7.1"
(rr)
```

and these end up not matching.

----

(It might be nice to offer some option to let users override the version-checking? I.e. as a human I can see these version strigs and *know* that is problem with the string comparison routine, and its not flagging a real risk...)

pnkfelix:
Calls trim_end to strip trailing whitespace (such as a newline) from the why3 version string we read in from the why3 tool.

Also simplify similar code in detect_altergo_version which [definitionally was equivalent to trim_end](https://doc.rust-lang.org/src/core/str/mod.rs.html#2010).

Fix #982

### DIFF
--- a/creusot-setup/src/tools.rs
+++ b/creusot-setup/src/tools.rs
@@ -168,7 +168,7 @@ fn detect_why3_version(why3: &Path) -> Option<String> {
     let version_full = String::from_utf8(output.stdout).ok()?;
     let version = version_full.strip_prefix("Why3 platform, version ");
     version.map(|ver| {
-        let parts: Vec<_> = ver.split(|c| c == '.' || c == '+').collect();
+        let parts: Vec<_> = ver.trim_end().split(|c| c == '.' || c == '+').collect();
         String::from(&parts[..3].join("."))
     })
 }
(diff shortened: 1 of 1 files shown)


Armael:
thanks!

PR #984, opened 2024-04-04
Print config/data/cache paths with `cargo creusot setup status`

### LINKED ISSUE #977: (easy improvement) `creusot setup status`: display location of the config/data/cache dirs
Easy improvement to ease debugging scenarios: have `cargo creusot setup status` display the paths of its config/data/cache directories.

arnaudgolfouse:
To fix #977.

### DIFF
--- a/creusot-setup/src/lib.rs
+++ b/creusot-setup/src/lib.rs
@@ -23,6 +23,14 @@ struct CfgPaths {
     cache_dir: PathBuf,
 }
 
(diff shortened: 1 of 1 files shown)


Armael:
thanks! Small nit: could you instead display the paths at the end of the output, since they are "less important"/more expert information compared to the rest?

arnaudgolfouse:
Like this ?

PR #986, opened 2024-04-05
Replace the 'dev-why3' binary by 'dev-env' which sets shell environment variables

Armael:
This is strictly more flexible: instead of requiring the use of a custom why3 wrapper, we provide a way to setup a shell environment where `why3` resolves to the correct binary, and WHY3CONFIG indicates the config file.

This should then "just work" for binaries such as why3_tools that link to the why3 api, since they will pick up WHY3CONFIG.

@xldenis it may be useful for you to test this to check that it works in presence of paths with spaces or other unexpected macos weirdness.
It should be enough to check that the output of `cargo run --bin dev-env` looks reasonable, and that `eval $(cargo run --bin dev-env)` updated `PATH` and `WHY3CONFIG` in a way that works. Alternatively you could also check that `./ide` still works...

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -444,6 +444,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "creusot-setup",
+ "which",
 ]
 
 [[package]]
--- a/HACKING.md
+++ b/HACKING.md
@@ -5,9 +5,9 @@ working on the Creusot codebase.
 
 ## Setup
 
-On top of the usual Rust/Cargo workflow, the "Creusot developer setup" requires
-a working Why3 in order to run the testsuite and update tests proofs. You have
-two choices:
+The "Creusot developer setup" sometimes requires more flexibility in how it
+looks up why3 and related solvers, compared to the standard "user" workflow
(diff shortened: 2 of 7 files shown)


Armael:
To be perhaps clearer: the motivation for this PR is that currently, there's no easy way to launch `why3_tools` so that it uses the why3 config produced by `creusot setup`. Instead there's a way to launch `why3 -C path/to/the/config` which is what the `dev-why3` wrapper does, but it doesn't generalizes to `why3_tools` since that's a separate binary.
So instead of this why3 wrapper I propose to have a binary to setup the PATH and WHY3CONFIG environment variables, which works both to launch why3 and why3_tools.

xldenis:
To be clear, this doesn't change the invocation of commands like `cargo creusot why3 ide`, right? it only offers an additional way to get a shell with the correct why3 available?

xldenis:
I confirmed that `./ide` works and that `env` was properly updated.

Armael:
correct, it's just an extra mechanism to launch the correct why3, it doesn't change `cargo creusot why3 ide`

PR #997, opened 2024-04-30
Issue 943

laurentder:


cargo-creusot/src/why3_launcher.rs
```diff
@@ -0,0 +1,124 @@
...
+        match self {
+            Why3Mode::Ide => f.write_str("ide"),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct Why3Launcher {
+    mode: Why3Mode,
+    why3_path: Option<PathBuf>,
```
xldenis, line 28: Given that we cannot launch why3 (according to line 47) without this, can we make this not optional? To me, the error should be launched in the builder instead.

cargo-creusot/src/why3_launcher.rs
```diff
@@ -0,0 +1,124 @@
...
+    ) -> Self {
+        Self { mode, why3_path, config_file, args, output_file }
+    }
+
+    pub fn make(&self, temp_dir: &Path) -> Result<Command> {
+        let Some(ref mlcfg_file) = self.output_file else {
+            return Err(anyhow!("can't launch why3, no mlcfg_file specify"));
+        };
+        let mode = self.mode.to_string();
+        PRELUDE.extract(temp_dir).expect("can't launch why3, could extract prelude into temp dir");
```
xldenis, line 50: Not that I have any objection with this, but why is it now necessary to copy the prelude to a temporary directory? Just hygiene? Or is there a more fundamental reason?

xldenis (commented):
Thanks for this work! I think it's a nice step forward. I think a follow up could involve removing the code in `creusot-rustc` which is made obsolete by this, though I think that would require also handling `prove` and `replay`. cc @dewert99.

cargo-creusot/src/why3_launcher.rs
laurentder: I took this code in: creusot/src/run_why3.rs::run_why3

cargo-creusot/src/why3_launcher.rs
laurentder: Ok, like you want

laurentder:
> Thanks for this work! I think it's a nice step forward. I think a follow up could involve removing the code in `creusot-rustc` which is made obsolete by this, though I think that would require also handling `prove` and `replay`. cc @dewert99.

yes the final goal is also to manage the "prove" and the "replay", however I will not have time this week to achieve this.

xldenis:
> yes the final goal is also to manage the "prove" and the "replay", however I will not have time this week to achieve this.

That's totally fine, but I am looking forward to the day when we can clean that up :)

PR #999, opened 2024-05-06
Detect recursion

arnaudgolfouse:
This adds a new mechanism for detecting (mutual) recursion. In particular this now correctly flags cases that use complicated traits instances, like this one:
```rust
trait Foo {
    fn foo() {}
}

impl Foo for i32 {
    fn foo() {
        bar::<std::iter::Once<i32>>(std::iter::once(1i32));
    }
}

fn bar<I>(i: I)
where
    I: Iterator,
    I::Item: Foo,
{
    I::Item::foo();
}
```

# Design

Two new attributes are added: `#[terminates]` and `#[pure]`:
- A `#[terminates]` function can only call other `#[terminates]` functions, and must not have a recursive cycle in its call graph.
  
  Note that only functions that can be called by specialized instances of the calling functions are _not_ considered. For example:

  ```rust
  trait Tr {
      fn f() {}
  }
  fn g<T: Tr> {
      Tr::f();
  }
  // Nothing happens yet: as far as `g` is concerned, `Tr::f` does not call any other functions.
  impl Tr for i32 {
      fn f() {
          g::<i32>();
      }
  }
  // NOW there is an error, because this specialization of `g` calls `<i32 as Tr>::f` (the current function).
  ```
- A `#[pure]` function is a function that is both terminating, and that cannot `panic!`. This will be useful for adding executable ghost code.

  Only rare functions are `#[terminates]` but not `#[pure]`. Examples include `Vec::push` (can panic because of the length overflow).

  However out-of-memory is not considered, because it cannot happen in ghost code.

Logical functions are automatically marked with `#[pure]`.

# Caveats

- This only detects recursion, not loops
- Structural recursion is not handled, so we must use variants to prove the termination of e.g. `List::len`.


xldenis:
> This only detects recursion, not loops
> Structural recursion is not handled, so we must use variants to prove the termination of e.g. List::len.

While I would be fine with an initial implementation excluding these aspects, I would hope that both could be added subsequently, as neither should be complicated to provide.

creusot-contracts/src/std/iter.rs
```diff
@@ -154,11 +160,14 @@ extern_spec! {
                     where T: IntoIterator<Item = A>, T::IntoIter: Iterator;
             }
 
+            #[pure]
             fn empty<T>() -> Empty<T>;
 
+            #[pure]
```
xldenis, line 166: Perhaps in the future we could make a lint which recommends adding `pure` to appropriate functions when you write code (not for extern specs of course).

creusot/tests/should_fail/recursion/complicated_traits.stderr
```diff
@@ -0,0 +1,19 @@
...
+11 |         bar::<std::iter::Once<i32>>(std::iter::once(1i32));
+   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+note: finally bar calls <i32 as Foo>::foo.
+  --> complicated_traits.rs:21:9
+   |
+21 |         I::Item::foo();
+   |         ^^^^^^^^^^^^^^
+
+error: aborting due to 1 previous error
+
```
xldenis, line 19: Beautiful!

claudemarche:
> > While I would be fine with an initial implementation excluding these aspects, I would hope that both could be added subsequently, as neither should be complicated to provide.
> 
> I agree that this would be a good feature to have, but I'm not sure this is so easy to implement. One has to either define the subterm relationship (as a WhyML relation ?), assert that it is well founded and use this as a variant (but then we rely on SMT solvers to prove a Why3 inductive relation, which they are not good at), or do a static analysis that tracks subterms in MIR code, which is yet another static analysis on MIR.

If the sum types of Rust are translated into Why3 sum types, I don't see why you could not rely on Why3 support for proving structural recursion. Live discussion needed I suppose. Could be in a future Why3 GT, it's been a lot of time without any such GT. Please suggest this kind of topic on the GT channel of the Why3 Zulip

creusot/src/validate_terminates.rs
```diff
@@ -196,28 +196,34 @@ pub(crate) fn validate_terminates(ctx: &mut TranslationCtx) {
...
+                            // else this crashes the trait selection.
+                            None
+                        } else {
+                            match selection_ctx.select(&obligation) {
+                                Ok(Some(source)) => match source {
+                                    rustc_infer::traits::ImplSource::UserDefined(source) => {
+                                        Some(source.impl_def_id)
+                                    }
+                                    rustc_infer::traits::ImplSource::Param(_) => {
+                                        // FIXME: we take the conservative approach here, but what does this case actually mean ?
```
xldenis, line 218: This case means that the impl comes from a bound on a parameter, I think we have to be very conservative here since a clever instantiation of that generic & bound is exactly how you might cheat the detection.

creusot/src/validate_terminates.rs
```diff
@@ -196,28 +196,34 @@ pub(crate) fn validate_terminates(ctx: &mut TranslationCtx) {
...
-                                    None
-                                }
-                                // Used for marker traits (no functions anyway) and trait object/unsized variables (we really don't know what they can call)
-                                rustc_infer::traits::ImplSource::Builtin(_, _) => None,
-                            },
-                            Ok(None) => None,
-                            Err(_) => None,
+                        let impl_def_id = if actual_clause.trait_ref.has_param() {
+                            // else this crashes the trait selection.
+                            None
```
xldenis, line 210: Can you explain when this occurs and what this prevents?

creusot/src/translation/specification.rs
```diff
@@ -103,6 +116,10 @@ impl ContractClauses {
             let term = ctx.term(var_id).unwrap().clone();
             out.variant = Some(term);
         };
+        log::trace!("no_panic: {}", self.no_panic);
```
xldenis, line 119: note: I would like to see us upgrade to `tracing` for our logs, it provides more contextual / structured information especially when dealing with deep callgraphs.

creusot/src/validate_terminates.rs
jhjourdan: Actually, I don't think there is any risk here, as long as we add a dependency at the instantiation site. Think of polymorphism in system F (which is terminating, as you know) : in system F, we add no dependency when using a type variable.

xldenis: I see what you mean, but I think at the very least his module would then benefit from more extensive description of the criteria, what is up there is quite good but this is fairly sophisticated reasoning.

creusot/src/validate_terminates.rs
```diff
@@ -196,28 +196,34 @@ pub(crate) fn validate_terminates(ctx: &mut TranslationCtx) {
...
+                                    }
+                                    rustc_infer::traits::ImplSource::Param(_) => {
+                                        // FIXME: we take the conservative approach here, but what does this case actually mean ?
+                                        None
+                                    }
+                                    // Used for marker traits (no functions anyway) and trait object/unsized variables (we really don't know what they can call)
+                                    rustc_infer::traits::ImplSource::Builtin(_, _) => None,
+                                },
+                                Ok(None) => None,
+                                Err(_) => None,
```
xldenis, line 225: I think that the trait resolution stuff could be moved out into a helper method in the same module, it would keep the code a bit clearer imo, ie: `resolve_source(tcx, obligation) -> ImplSource`

creusot/src/validate_terminates.rs
jhjourdan: I don't understand: do you simply complain about the missing doc?

creusot-contracts/src/lib.rs
```diff
@@ -27,6 +27,14 @@ mod macros {
 
     pub use base_macros::snapshot;
 
+    /// Indicate that the function terminates: fullfilling the `requires` clauses
+    /// ensures that this function will not loop indefinitively.
+    pub use base_macros::terminates;
+
+    /// Indicate that the function is pure: fullfilling the `requires` clauses ensures
+    /// that this function will terminate, and will not panic.
```
xldenis, line 35: I think this description is a bit unclear since Creusot forces to you to prove absence of panics. I think we need to clarify that for programs Creusot actually 'allows' some kinds of panics (ie overflow of `Vec::len`). Otherwise it looks like `terminates = pure`. 

I wonder if it could be interesting to have programs which _do not_ verify the absence of panics? where the contract then says "if the function terminates successfully, _then_ the result satisfies the postcondition? (Not for here / now at all)

creusot/src/validate_terminates.rs
```diff
@@ -424,68 +425,81 @@ impl<'thir, 'tcx> thir::visit::Visitor<'thir, 'tcx> for FunctionCalls<'thir, 'tc
...
+    /// The current path of called function, e.g. `stack[i]` calls `stack[i + 1]`.
+    ///
+    /// All nodes in the stack must also be in `visited`.
+    stack: Vec<FunctionInstance<'tcx>>,
+    /// Nodes in the `graph` that were already visited. Also contains an index to order
+    /// `FunctionInstance`s.
+    visited: IndexMap<FunctionInstance<'tcx>, usize>,
+    /// The resulting set of cycles found.
+    detected_cycles: IndexSet<Vec<FunctionInstance<'tcx>>>,
 }
```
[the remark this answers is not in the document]
arnaudgolfouse: Yep, that helped!

creusot/src/validate_terminates.rs
arnaudgolfouse: Ok, thanks for the explanation!
Anyway this code wasn't working with the latest rustc, and I've taken the approach of using `TyCtxt::resolve_instance` here, which seems more robust and simpler to read.

creusot/src/validate_terminates.rs
arnaudgolfouse: So a check appeared somewhere in the call of `SelectionContext::resolve`: https://github.com/rust-lang/rust/blob/34582118afaf00b0eb2d209a90b181c7156b501c/compiler/rustc_trait_selection/src/traits/coherence.rs#L802
This `if` was a temporary workaround, now I use `TyCtxt::resolve_instance` which seems simpler and more robust than `SelectionContext::resolve`.

creusot-contracts/src/lib.rs
arnaudgolfouse: Indeed, I added lines about this just below this doc

arnaudgolfouse:
I think at this point this PR is more or less ready. The remaining questions are:
- how do we allow structural recursion (e.g. List::len) ?
- Can we take advantage of why3 termination checks, especially for logical functions ?

As-is, the termination check required minimal changes to the test suite (mostly in `derive_macros.rs`), so while those issues should eventually have an answer, I think the current state of things is enough from a usability viewpoint.

Mostly I should document what is allowed/forbidden by this check :)

xldenis:
> As-is, the termination check required minimal changes to the test suite (mostly in derive_macros.rs), so while those issues should eventually have an answer, I think the current state of things is enough from a usability viewpoint.

Agreed, and this is also a reason why we had a release, we can point external users to the officially supported version while we (temporarily) regress on HEAD. 

> how do we allow structural recursion (e.g. List::len) ?

I _still_ don't understand why there is an issue at _all_ with structural recursion? 

> Can we take advantage of why3 termination checks, especially for logical functions ?

I hope so, the other question to ask though is: are there any functions we accept (correctly) which why3 rejects?

PR #1003, opened 2024-05-17
creusot setup v2: more flexible per-tool settings for managed/external status and version checking

Armael:
This is a refactor/overhaul of creusot setup.
The motivation is to provide a better workflow for upgrading tools from one version to the next (why3 and solvers).
(Incidentally this should fix the nightly CI job which currently fails because it detects that why3 has a new "unexpected" version and stops... instead we'd like to still use the bundled solvers, but relax the version checking for why3! this PR makes that possible.)

There are three commits:

Commit 92d9b16 contains the bulk of the work:
    Rework 'creusot setup' to allow toggling "external binary vs downloaded binary" for each tool separately, instead of
    having a global setting that applies to all tools like previously. (And similar for version checking.)
    The new 'creusot setup' cli makes it easy to chose (at the install step) for each tool:
    - whether the tool will be downloaded from a binary release or looked up in the current PATH.
    - whether 'creusot setup' will error and quit when the tool has a different version than expected.

Commit 1bfa0ba cleans up some codepaths left broken by the first commit. The implemented solution is up for discussion: I chose to be somewhat opinionated and cut down the complexity by removing some flexibility.
The commit makes it so the creusot config is always looked up from the "global" config file (removing the ability to use a custom config for running the testsuite), and the why3 config file we use (why3.conf) is always the one generated by creusot setup (removing the ability to use ~/.why3.conf).
Previously there were ways to have creusot use the ~/.why3.conf config file, conditionally on whether one is running the testsuite from the git repo. But I think this was somewhat clunky and I'm not sure anyone actually used these escape hatches...

Commit a6c8d56 unstucks the nightly CI by using the new option `cargo creusot setup install --no-check-version why3`.

### DIFF
--- a/.creusot-config.sample/Config.toml
+++ b/.creusot-config.sample/Config.toml
@@ -1,6 +1,0 @@
-version = 1
-
-[tools]
-mode = "external"
-why3_path = "why3"
-altergo_path = "alt-ergo"
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -34,6 +34,6 @@ jobs:
     - name: run cargo creusot setup install
       if: steps.cache-creusot-setup.outputs.cache-hit != 'true'
       run: |
-        cargo run --bin cargo-creusot creusot setup install
+        cargo run --bin cargo-creusot creusot setup install --no-check-version why3
     - run: cargo test --test why3 "" -- --replay=none --diff-from=origin/master
     - run: cargo test --test why3 "" -- --skip-unstable
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -45,7 +45,9 @@ jobs:
     - name: Build
       run: cargo build
     - name: dummy creusot setup
-      run: cp -r .creusot-config.sample .creusot-config
+      run: |
+        mkdir -p ~/.config/creusot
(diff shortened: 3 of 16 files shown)


Armael:
The main question I have I guess is whether we're happy (for now) to have a slightly more streamlined `creusot setup` workflow where:
- there is a single global creusot configuration of the solvers
- one needs to run `creusot setup install [flags...]` each time one wants to change this configuration (e.g. to use a newer solver, etc)

(NB: the downloaded binaries for the solvers are cached so repeating `creusot setup install` only needs to regenerate why3.conf and the Config.toml, which is fast.)

xldenis (approved):
> I chose to be somewhat opinionated and cut down the complexity by removing some flexibility.
The commit makes it so the creusot config is always looked up from the "global" config file (removing the ability to use a custom config for running the testsuite), and the why3 config file we use (why3.conf) is always the one generated by creusot setup (removing the ability to use ~/.why3.conf).
Previously there were ways to have creusot use the ~/.why3.conf config file, conditionally on whether one is running the testsuite from the git repo. But I think this was somewhat clunky and I'm not sure anyone actually used these escape hatches...

This seems reasonable to me, I never personally used that flexibility and I don't know of anyone else who was. 

The code itself looks good, I've got no remarks to make.

Armael:
Thanks for the review!

PR #1009, opened 2024-05-22
Fix the `mlcfg` script

arnaudgolfouse:
The `mlcfg` script at the root of the repo was failing with the error
```
error: the following required arguments were not provided:
  --why3-config-file <WHY3_CONFIG_FILE>

Usage: creusot-rustc --why3-config-file <WHY3_CONFIG_FILE> --output-file <OUTPUT_FILE> --span-mode <SPAN_MODE> -- <RUST_FLAGS>...

For more information, try '--help'.
```

@Armael is that the correct fix ?


Armael:
Perfect, thanks a lot for the fix!

PR #1010, opened 2024-05-22
upgrade why3 and cleanup the session for the red_black_tree test

### LINKED ISSUE #1007: testsuite red_black_tree: remove manual transformations from the session
Currently the proof of the red_black_tree example involves some manual transformations (`compute_hyp`) added in #1005 when upgrading alt-ergo from 1.4.3 to 1.5.3.
The reason is https://gitlab.inria.fr/why3/why3/-/issues/854 ; after the issue is addressed in why3 we should clean up the red_black_tree proof.

Armael:
Fixes #1007 now that the why3 issue has been fixed.


xldenis (approved):
Thanks. Interesting that fixing this issue had such a major impact on the size of sessions!

Issue #1013, opened 2024-05-29
Verification failure for rust code in recent commit


ruth561:
I've recently started exploring Creusot. When attempting to verify the following Rust code snippet, verification fails:

```rust
#[ensures((^x) == 0i32)]
pub fn set_zero(x: &mut i32) {
	*x = 0;
}
```

In the Why3 IDE, the screen appears as shown below:

![スクリーンショット 2024-05-29 184914](https://github.com/creusot-rs/creusot/assets/87105992/18a69597-8923-4f78-b7b1-da18cfecba87)

When I moved to a commit from about two months ago ([3fdf297](https://github.com/creusot-rs/creusot/tree/3fdf29720144171393804e8168141a0f93217db4)) and tried, the verification succeeded. The Why3 IDE screen at that time looked similar to the image below.

![スクリーンショット 2024-05-29 173808](https://github.com/creusot-rs/creusot/assets/87105992/d80b0a8a-ca2c-45e7-bda4-eba6d9fcd59f)

I'd like to report this as an issue.

xldenis:
Hi, thanks for reporting, I'll take a look when I get home later. Already, I can see that the proximal cause is that `resolve0` is opaque in your first screenshot, the proof isn't allowed to use the definition. In the interest of speeding up resolution, can you provide the _complete_ source code and process you used to generate your `mlcfg` file?

Notably, did you use `cargo creusot`, and is this a standalone crate? Did `cargo creusot` output any warnings or other log messages?

xldenis:
So I wasn't able to reproduce using the `./mlcfg` script in the creusot repo, but I was using `cargo creusot`! Not sure why this occurs yet though.

xldenis:
Ok, so for some reason, the opacity of `resolve` is different when using `cargo creusot` in that crate than when using `mlcfg`.

xldenis:
We are probably losing something with the way we encode "opacity witnesses" in a multi-crate context and the way that we compile `creusot-contracts` for tests is sufficiently different to not materialize this. I think a simple fix will probably be to store opacity in crate metadata.

ruth561:
Hi,

Thank you for your response. Here are the details regarding my environment and the contents of the MLCFG file.

First, this is a standalone crate with the following directory structure:

```
$ tree -a -I ".git|target|mir_dump"
.
├── .gitignore
├── Cargo.lock
├── Cargo.toml
├── rust-toolchain
└── src
    └── main.rs

[...]

PR #1014, opened 2024-05-30
Fix metadata load path

xldenis:
In the recent rustc update, some apis changed with regards to getting output file locations in rustc, as a result I accidentally broke the loading of metadata. We would store to `foo-bar.cmeta` but attempt to load from `libfoo-bar.cmeta`, I've fixed this now, solving #1013. 

Unfortunately this is hard to test since our `ui` tests don't have this issue because they explicitly spell out the path for `creusot-contracts` metadata as an extra parameter.

### DIFF
--- a/creusot/src/metadata.rs
+++ b/creusot/src/metadata.rs
@@ -6,6 +6,7 @@ use indexmap::IndexMap;
 use rustc_hir::def_id::{CrateNum, DefId};
 use rustc_macros::{TyDecodable, TyEncodable};
 use rustc_middle::ty::TyCtxt;
+use rustc_session::config::OutputType;
 use rustc_span::Symbol;
 use std::{
     collections::HashMap,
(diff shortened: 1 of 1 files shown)


jhjourdan:
Should'nt we detect that a dependancy has been compiled with Creusot but the cmeta file is not found?

jhjourdan:
What I am proposing is that when we compile with Creusot, then a special attribute is inserted in the crate (thanks to macros?), and then if we load a crate that has this attribute but we don't have the cmeta file, then we error.

xldenis:
> What I am proposing is that when we compile with Creusot, then a special attribute is inserted in the crate

That's not possible unless the user specifically called a `creusot_marker!()` macro or we generated a _ton_ of garbage.

jhjourdan:
Such a macro could actually be a good thing, because anyway we need to enable some features at the beginning of any Creusot crate.

xldenis:
> Such a macro could actually be a good thing, because anyway we need to enable some features at the beginning of any Creusot crate.

But hard to enforce, and frankly I'm not a fan of, I'd rather see the few features we need stabilized, which is something I talked to Felix about at ETAPS. 

An alternative we _can_ do however, is query the dependencies of a crate and if we see `creusot-contracts` in it we can error.

PR #1019, opened 2024-06-03
setup: allow configuring the number of cores used by why3

### LINKED ISSUE #1011: setup: allow configuring the number of cores used by why3
Currently the why3 config file generated by `creusot setup` makes why3 use its default number of parallel solver processes, which is 2. It would probably be nicer to:
- by default, have `creusot setup` generate a why3 config file with `running_provers_max` equal to the number of cores of the machine
- add a command-line option to `creusot setup` to allow tweaking this number if needed

Armael:
Fixes #1011


xldenis:
Seems good to me modulo CI passing.

Armael:
(I had to regenerate a number of broken sessions in the testsuite.)

PR #1024, opened 2024-06-21
Update exercises

nhusung:
Before turning to a larger project, I thought it might be better to solve a few small exercises. Unfortunately, two of the provided ones have little bugs such that the compilation/translation to Coma already fails. I am not confident about the `DeepModel` trait bound (still learning Creusot …) but it seemed to be required.

### DIFF
--- a/exercises/02_linear_search.rs
+++ b/exercises/02_linear_search.rs
@@ -4,11 +4,11 @@ use creusot_contracts::*;
 // Prove the following:
 // 1. If we return Some(i) it is the first index containing `tgt`
 // 2. If we return None, then there are no indices containing `tgt`
-fn search<T: Ord>(v: &Vec<T>, tgt: &T) -> Option<usize> {
+fn search<T: Ord + DeepModel>(v: &[T], tgt: &T) -> Option<usize> {
     let mut i = 0;
 
     while i < v.len() {
-        if v[i] == tgt {
+        if v[i] == *tgt {
             return Some(i);
         }
 
--- a/exercises/04_all_zero.rs
+++ b/exercises/04_all_zero.rs
@@ -6,9 +6,7 @@ use creusot_contracts::*;
 pub fn all_zero(v: &mut Vec<u32>) {
     let mut i = 0;
     let old_v = snapshot! { v };
-    // Until https://gitlab.inria.fr/why3/why3/-/merge_requests/667 is merged
-    // the following invariant is needed to allow Why3 to remember prophecies dont change
-    #[invariant(proph_const, ^v == ^old_v.inner())]
+
     while i < v.len() {
         v[i] = 0;
         i += 1;


exercises/02_linear_search.rs
```diff
@@ -4,11 +4,11 @@ use creusot_contracts::*;
 // Prove the following:
 // 1. If we return Some(i) it is the first index containing `tgt`
 // 2. If we return None, then there are no indices containing `tgt`
-fn search<T: Ord>(v: &Vec<T>, tgt: &T) -> Option<usize> {
+fn search<T: Ord + DeepModel>(v: &[T], tgt: &T) -> Option<usize> {
```
sarsko, line 7: Are you saying compilation/translation to Coma fails without this? Or did you have to add this to prove whichever spec you wanted to prove?

nhusung: Ok, this happened after adding the specification.

```rs
use creusot_contracts::*;

#[ensures(forall<i: usize> result == Some(i) ==> i@ < v@.len() && forall<j: Int> 0 <= j && j < i@ ==> v@[j] != *tgt)]
#[ensures(result == None ==> forall<i: Int> 0 <= i && i < v@.len() ==> v@[i] != *tgt)]
pub fn search<T: Ord>(v: &[T], tgt: &T) -> Option<usize> {
    let mut i = 0;

    #[invariant(forall<j: Int> 0 <= j && j < i@ ==> v@[j] != *tgt)]
    while i < v.len() {
        if v[i] == *tgt {
            return Some(i);
        }

        i += 1
    }

    return None;
}
```

For the code above, I get the following error message:

```
error[E0277]: the trait bound `T: creusot_contracts::DeepModel` is not satisfied
  --> src/lib.rs:10:12
   |
10 |         if v[i] == *tgt {
   |            ^^^^^^^^^^^^ the trait `creusot_contracts::DeepModel` is not implemented for `T`
```

I can also use `tgt@` instead of `*tgt` in the specifications, but then I get an additional message about `ShallowModel` not being implemented by `T`.

I’m on commit `917bcee1`, if this is of interest.

xldenis: Indeed, you need to add this bound. I think my original intent was to have that be part of the exercise, but its probably best to provide that for the user.

exercises/04_all_zero.rs
```diff
@@ -6,9 +6,7 @@ use creusot_contracts::*;
 pub fn all_zero(v: &mut Vec<u32>) {
     let mut i = 0;
     let old_v = snapshot! { v };
-    // Until https://gitlab.inria.fr/why3/why3/-/merge_requests/667 is merged
-    // the following invariant is needed to allow Why3 to remember prophecies dont change
-    #[invariant(proph_const, ^v == ^old_v.inner())]
```
xldenis, line 11: Keep in mind, this is actually now needed _again_ by the change to coma. At least until 0.2 comes out (soon ish).

xldenis:
Hi Nils,

Thanks for the PR, I think we'll merge it but there are a few remarks / questions I have for you:

> Unfortunately, two of the provided ones have little bugs such that the compilation/translation to Coma already fails.

Do you mean that Creusot panics? Returns an error? Or that it generates ill-formed Coma code? 

More generally, I would strongly recommend sticking to `v0.1`, which is our first "stable" version and still uses MLCFG. The current Coma branch on master has some regressions compared to 0.1 though by 0.2 it will supersede it.

exercises/04_all_zero.rs
nhusung: Interesting, so that’s why I needed `#[invariant(^*old_v == ^v)]`? The old syntax with `proph_const` didn’t seem to work …

exercises/02_linear_search.rs
nhusung: In fact, I started to wonder whether adding the `DeepModel` trait bound was intended to be part of the exercise once I read @sarsko’s comment :sweat_smile: I guess the actual issue is that `DeepModel` is not mentioned in the [guide](https://creusot-rs.github.io/creusot/). Otherwise you could use [`#[diagnostic::on_unimplemented]`](https://github.com/rust-lang/rust/pull/119888) and link the respective page :slightly_smiling_face:

nhusung:
> Do you mean that Creusot panics? Returns an error? Or that it generates ill-formed Coma code?

I just noticed ordinary error messages. So everything is fine :slightly_smiling_face:

exercises/02_linear_search.rs
```diff
@@ -4,11 +4,11 @@ use creusot_contracts::*;
...
     let mut i = 0;
 
     while i < v.len() {
-        if v[i] == tgt {
+        if v[i] == *tgt {
```
nhusung, line 11: The missing `*` (or `&` before `v[i]`) is already an ordinary Rust error

PR #1026, opened 2024-06-28
Rename `ShallowModel` and `DeepModel`

xldenis:
With @jhjourdan we talked for years at this point about how we should rename the confusingly named `Shallow` and `DeepModel` traits. Working on Gillian-Rust I was once again confronted by this confusion, and figured I might as well pull the trigger on fixing this.


jhjourdan:
Great!

During one minute, I wondered wether `EqModel` was great and wether it should be replaced by simply `Model`. I really don't have a strong opinion, but `EqModel` may seem a bit criptic for newcomers. On the other hand, it better describe the fact that this really is the logical function that describes what it means to be compatible with program equality. @Armael, @voidc, @arnaudgolfouse, @dewert99, any opinion?

arnaudgolfouse:
I do think `EqModel` is slightly weird, but I'm not sure `Model` would be that much better. But the doc comments should be updated anyway  😁 !

guide/src/basic_concepts/requires_ensures.md
```diff
@@ -25,7 +25,7 @@ fn head(v: &[i32]) -> i32 {
 
 Note the **shallow model** (`@`) operator: it is needed to convert the Rust type `&[i32]` to a logical type `Seq<i32>`. Else, we could not call `[T]::len`, which is a program function (and not a logical one).
```
arnaudgolfouse, line 26: Shouldn't we call it the 'view operator' then ?

Armael:
You didn't write in the PR message what `ShallowModel` and `DeepModel` get renamed into.
It looks like the current choice is `View` and `EqModel`. This sounds fine to me, but I originally thought you would go for `Model` and `EqModel`, which I thought has some symmetry to it (there's the default model and the model for equality tests). Did you have a specific reason to avoid having `EqModel` without `Model`, or `View` instead of `Model`? 
(I'm not particularly attached to any particular option, though I think the new name for `ShallowModel` should carry the intuition that it is the "default" or "most natural" of the two options:)

xldenis:
> I do think EqModel is slightly weird, but I'm not sure Model would be that much better. But the doc comments should be updated anyway 😁 !

I agree that its a weird name, I do think it should have `Eq` in it because otherwise it introduces a confusion as people take it to be a _canonical_ model in the style of RustHornBelt. 

> You didn't write in the PR message what ShallowModel and DeepModel get renamed into.

You're right, my PR description was rather sloppy, I was hoping to get @jhjourdan's opinion on actually pulling the trigger on this refactor. 

> It looks like the current choice is View and EqModel. This sounds fine to me, but I originally thought you would go for Model and EqModel, which I thought has some symmetry to it (there's the default model and the model for equality tests).
> Did you have a specific reason to avoid having EqModel without Model, or View instead of Model?
I actually do want to break the symmetry between them, I think the idea that `DeepModel` and `ShallowModel` are related is a problem which even I have struggled with for a long time. At one point I thought that we could define `ShallowModel` as being "one step of unfolding" `DeepModel`, but I think @jhjourdan had a counter-example to that. 

> (I'm not particularly attached to any particular option, though I think the new name for ShallowModel should carry the intuition that it is the "default" or "most natural" of the two options:)

I'm not attached to the particular names either. I thought that the way to convey that ex-`ShallowModel` is the default would be by giving it a prefix free name (ie: `View`, `Model`, etc...)

jhjourdan:
Can we make progress on this? Everybody agrees the current situation is not good. We simply have to make a choice, then it's just a matter of executing a `sed` script.

My favorite choice would be `Model` and `View`, but `EqModel` is ok if you really prefer this. My point is that `EqModel` really has nothing more to do with equality than e.g., orders or hash functions.

xldenis:
Yea my once en is that DeepModel is in no way "canonical" it exists to serve the trifecta of traits: Eq,Ord,Hash and that naming it "Model" implies a status I'm not sure we want to encourage.

Armael:
Thinking about it a bit more this morning: what about `View` and `DeepModel`?

jhjourdan:
Looks like a good compromise. @xldenis ?

xldenis:
Sure.

jhjourdan:
Then let's go for it when we'll have a bit of time.

jhjourdan:
Subsumed by #1165.

PR #1031, opened 2024-07-15
Ghostptrtoken fixes

dewert99:
* Makes `GhostPtrToken::fin` `logic(prophetic)`, instead of `logic`.
* Makes the `Copy` implementation of `GhostPtrTokenRef<T>` apply for all `T`.
* Adds `GhostPtrTokenRef::to_ref` to allow accessing references with the longer lifetime.
* Adds a runtime check to make sure no elements in a `GhostPtrToken` are zeros sized.
  - `GhostPtrToken::box_to_ptr` ensures that the returned pointer isn't in the token which might not be true for ZSTs.
  - For statically sized types this check is trivial and should get compiled away.

### DIFF
--- a/creusot-contracts/src/ghost_ptr.rs
+++ b/creusot-contracts/src/ghost_ptr.rs
@@ -14,7 +14,6 @@ pub struct GhostPtrToken<T: ?Sized>(PhantomData<T>);
 /// ZST equivalent of [`&'a GhostPtrToken<T>`](GhostPtrToken)
 /// Can be created using [`GhostPtrToken::borrow`]
 #[trusted]
-#[derive(Copy, Clone)]
 pub struct GhostPtrTokenRef<'a, T: ?Sized>(PhantomData<&'a T>);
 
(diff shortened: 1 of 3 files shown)


creusot-contracts/src/ghost_ptr.rs
```diff
@@ -257,7 +271,7 @@ impl<'a, T> DerefMut for GhostPtrTokenMut<'a, T> {
 
 #[trusted]
 impl<'a, T> Resolve for GhostPtrTokenMut<'a, T> {
-    #[predicate]
+    #[predicate(prophetic)]
```
xldenis, line 274: Curious... this wasn't causing an error in CI? I thought we ran `cargo creusot` on `creusot-contracts` in CI?

dewert99: The resolve didn't need to be prophetic when `self.fin()` wasn't prophetic, but when I made `GhostPtrToken::fin` prophetic, `cargo creusot` broken until I also fixed `resolve`.

PR #1032, opened 2024-07-16
Make constant body optional. Add smart constructors for decls

xldenis:
This PR contains fixes necessary for @atoiz's work. 

The Why3 crate is really in desparate need of a near total rewrite, but that will have to wait.


xldenis:
@dewert99: please note I've generalized your work on triggers, which notably flipped the rather confusing meaning of values. 

Now, a `None` means no trigger was set (and thus none should be generated), and instead `use_simple_triggers` explicitly sets a trigger where relevant. 

Additionally, the triggers can now have multiple terms and multiple triggers can be set per quantifier.

PR #1043, opened 2024-07-29
Bump rustc

jhjourdan:
This was mostly easy, nothing to declare.


creusot/src/translation/function/statement.rs
```diff
@@ -169,6 +170,7 @@ impl<'tcx> BodyTranslator<'_, 'tcx> {
                 Operand::Constant(crate::constant::from_ty_const(
                     self.ctx,
                     *len,
+                    Ty::new_usize(self.ctx.tcx),
```
xldenis, line 173: strictly speaking we should be able to do `self.ctx.tcx.types.usize` here, but not important

creusot/tests/should_fail/terminates/trait_def_and_impl_disagree.stderr
```diff
@@ -10,10 +10,10 @@ error: Expected `h` to be `#[pure]` as specified by the trait declaration
 22 |     fn h() {}
    |     ^^^^^^
 
-error: Expected `f` to be `#[terminates]` as specified by the trait declaration
-  --> trait_def_and_impl_disagree.rs:17:5
+error: Expected `g` to be `#[pure]` as specified by the trait declaration
+  --> trait_def_and_impl_disagree.rs:20:5
```
xldenis, line 14: Is this is an expected change in error?

xldenis (commented):
lgtm, one comment about a test whose output changed.

xldenis:
`vc_insert_unique.39.0` is frustratingly unstable...

creusot/tests/should_fail/terminates/trait_def_and_impl_disagree.stderr
jhjourdan: For some reason, iterators over syntax tree definitions changed the iteration order, which results in order changes in generated coma files and in printed errors.

PR #1044, opened 2024-07-29
non prefix cur fin

xldenis:
This PR ditches prefix operators for their ordinary field access definition. This makes creusot code work better with `compute_specified`, which makes goals much more readable. 

A follow up piece of work would be for us to automatically raise the `compute_max_steps` meta to an appropriate value (the default 1024 being too low), and to provide a custom strategy as part of the creusot distrubition which systematically applies it at the start of proofs.

For reference, taking the goal of `move_red_left` in RBT, the *whole* task is ~14kloc in the ide before `compute_specified` and ~800 after. This shouldn't affect provers, but tremendously helps human debugging and seems to have *some* interaction with other transformations. 

In the process of working on this I've discovered that the performance issues in RBT seem to be at least partially due to `Tree::internal_invariant`, I don't know *why* (ie: what the actual matching loop is) but there is clearly a problem, it causes a ~8s increase in prover time for certain tasks.


jhjourdan:
Thanks for this piece of work, this will surely help debugging proofs in Creusot!

> In the process of working on this I've discovered that the performance issues in RBT seem to be at least partially due to Tree::internal_invariant, I don't know why (ie: what the actual matching loop is) but there is clearly a problem, it causes a ~8s increase in prover time for certain tasks.

Interesting. We should try reformulating the invariant in order to see if there is a way to make it less prone to matching loops. Also, ditching type invariants from quantifiers translation may help when it will be ready.

PR #1047, opened 2024-07-30
New further specializable

jhjourdan:
This PR implements a much finer criterion to determine whether a trait method implementation is further specializable or not.

The new criterion is:
- If the trait solver finds an instance which is not marked as default, then it is not specializable.
- Otherwise, if the orphan rules do not prevent a cousin or downstream crate to implement, then it is potentially specializable.
- Otherwise, we look at the specialization graph and determine whether there exists a more specialized instance which could apply.


jhjourdan:
@xldenis I'd be happy if you could tell me what you think about this.

xldenis:
> Otherwise, if the orphan rules do not prevent a cousin or downstream crate to implement, then it is potentially specializable.

Eliminating the double negation you mean "If orphan rules allow a cousin or descendant crate to implement the trait, it is potentially specializable"?

jhjourdan:
Indeed.

xldenis:
LGTM. 

The new criterion is clear and seems correct to me. We do seem to generate a few type trivial type invariants in a slightly different way now but that shouldn't matter.

PR #1057, opened 2024-08-03
Fix bug in creation of spec axioms

dewert99:
Spec axioms used to be replaced with true when using `simple-triggers`


creusot/src/backend/logic.rs
```diff
@@ -185,9 +185,9 @@ pub(crate) fn lower_logical_defn<'tcx, N: Namer<'tcx>>(
     );
 
     if has_axioms {
-        if sig.uses_simple_triggers() {
+        if sig.uses_simple_triggers() && !sig_contract.contract.variant.is_empty() {
```
xldenis, line 188: why is this second condition needed? we don't seem to require it in the non-simpl-trigger case?

xldenis (commented):
One question about the if expression, but also do you think you could provide a few tests for the `simple-trigger` functionality? I can pretty easily extend the test suite to handle this.

creusot/src/backend/logic.rs
dewert99: The current implementation only uses `lim` functions when there is a `variant` since otherwise it uses and interpreted why3 `function`

dewert99:
I added a test but #1058 needs to be merged before it will work

xldenis:
Done!

dewert99:
Am I fine to merge this?

PR #1058, opened 2024-08-03
Add magic comment for custom arguments in ui tests

xldenis:
@dewert99 this should let you add a simple-triggers test case with CREUSOT_ARG=--use-simple-triggers


dewert99:
Are you going to merge this?

dewert99:
I also fixed the parsing strategy so that it would accept arguments with "="

PR #1059, opened 2024-08-03
Fix spec that `GhostPtrToken`'s don't contain null pointers to apply to `Ref` and `Mut` variants

dewert99:


dewert99:
Am I fine to merge this?

Issue #1067, opened 2024-08-20
Modular verification


mojeanmac:
Verifying existing codebases must be done one module at a time. The current strategy of trusting individual functions and using `--focus-on` for individual strings might not scale for larger projects.
It would be great to have a way to focus on multiple modules, have "trusted" modules, or even have an option for creusot to ignore contractless functions.

xldenis:
@jhjourdan @Armael what do you think? Molly is working on her undergraduate thesis and needs some way of integrating verified into a larger unverified project. 

I think having `#[trusted]` modules trust all descendents is a good solution but might still be too cumbersome. Perhaps the semantics of `--focus-on` could be adjusted? something like `--only=MODULE_PATH` (with repetition allowed) causing creusot to only lookup and generate proofs for functions in those modules?

jhjourdan:
I put my veto on ignoring contractless functions. This makes too easy to forget to handle a function.

Trusted modules look like a good idea: this makes it easy to have modules handling e.g., some I/O operations or other "special" functions.

I'm don't like either `--focus-on` or other variants. Essentially, this makes it possible to assert stuff from the command line which makes it more difficult to review/identify the TCB. I'm not strongly opposed, though.

xldenis:
> I'm don't like either --focus-on or other variants. Essentially, this makes it possible to assert stuff from the command line which makes it more difficult to review/identify the TCB. I'm not strongly opposed, though.

I understand and agree, I mostly see `--focus-on` as a development tool to make why3 ide faster (since it suffers from long files).

> Trusted modules look like a good idea: this makes it easy to have modules handling e.g., some I/O operations or other "special" functions.

I think this might be the best choice overall, and shouldn't be complicated to implement.

mojeanmac:
Yes, I think that makes the most sense.

Also, while I'm not doing an undergrad thesis, I still hope to get what I'm working on published!

PR #1068, opened 2024-08-21
Inherit trusted attribute from parents

xldenis:
Here's a rough prototype of what was proposed to solve #1067.

I need to actually add tests here, but it should cover the intended usecase.

There's a wrinkle in the form of `trusted` on traits and types also having different semantics: I don't think we want to inherit `trusted` on types from parent modules, it makes them opaque. Similarly we probably don't want to inherit `trusted` for trait declarations?


jhjourdan:
Right. Note that the trusted declaration for types should go away at some point. Privacy is the right way of making types opaque.

xldenis:
@mojeanmac can you try this branch experimentally? you should be able to put `#[trusted]` on modules to disable all vcgen on anything inside, but I'm not sure I caught every edge case.

mojeanmac:
Ah here's one case where trusted doesn't cut it: Creusot still expects some impl'd traits to have the `creusot_contracts` definition as well. For example, a custom data type using `Default` still needs `creusot_contracts::Default` regardless if the module it's in is trusted or not.

xldenis:
Ah yea good point. I can't think of any *great* solution, but:

1. Use `cfg_attr(creusot,...)` to conditionally disable them.
2. Put the impls in a creusot only module.

PR #1070, opened 2024-08-22
Add sytnax to specify triggers for quantifiers

dewert99:
The syntax looks like:
```rust
forall<i: int, j: int>
        #![trigger a[i], b[j]]
        #![trigger a[i], c[j]]
        0 <= i < j < a.len() ==> a[i] != b[j] && a[i] != c[j],
```
and it is based on the Verus syntax:
```rust
forall|i: int, j: int|
        #![trigger a[i], b[j]]
        #![trigger a[i], c[j]]
        0 <= i < j < a.len() ==> a[i] != b[j] && a[i] != c[j],
```


dewert99:
Hmm I'm not sure why so many of the why3 tests are being marked as obsolete, I tried redoing the proofs with `why3 ide` and most of them passed (I saved the results). The only one I wasn't able to get working was `should_succeed/selection_sort_generic.coma` but that might be because I don't have CVC4.

xldenis:
I think you might actually need to update your version of why3, looking at the xml you have a bunch of trivial coma goals lurking around.

dewert99:
Has coma been merged into Why3 yet or do I need to pin another specific version?

xldenis:
> Has coma been merged into Why3 yet or do I need to pin another specific version?

Its merged, if you install the why3 version in the `creusot-deps.opam` file it should be good.

jhjourdan:
Nice! Could we extend this syntax to quantifiers generated by logical functions?

dewert99:
> Nice! Could we extend this syntax to quantifiers generated by logical functions?

Yes, but could we make that a separate PR.

dewert99:
@xldenis are you good with this being merged?

xldenis:
Yep!

xldenis (approved):

PR #1071, opened 2024-08-26
Placewise resolve

jhjourdan:
This fixes several unsoundnesses where mutable borrows were resolved too early (before an assignment). This avoids asserting type invariants on (partially) uninitialized locals.

In addition, I implemented an optimization which avoids resolving types which are Copy or that have no free lifetime variables.


xldenis:
Is this ready for review? Or is the WIP tag still active?

jhjourdan:
This is ready for review. I still need to rebase and perhaps do some cosmetic changes, but this is ready to review if you have some time.

jhjourdan:
I need to do something for partial resolution of FnOnce closures.

creusot/src/translation/function.rs
```diff
@@ -324,12 +435,96 @@ impl<'body, 'tcx> BodyTranslator<'body, 'tcx> {
...
+        location: Location,
+        destination: Place<'tcx>,
+    ) {
+        // The assignement may, in theory, modify a variable that needs to be resolved.
+        // Hence we resolve before the assignment.
+        self.resolve_places(need, &resolved, true);
+
+        // We resolve the destination place, if necessary
+        match self.move_data().rev_lookup.find(destination.as_ref()) {
+            LookupResult::Parent(None) => unreachable!(),
```
xldenis, line 455: Why is this unreachable?

creusot/src/translation/function.rs
```diff
@@ -339,30 +534,79 @@ impl<'body, 'tcx> BodyTranslator<'body, 'tcx> {
...
+                } else {
+                    continue;
+                };
+                let adt = discr_pl.ty(self.body, self.tcx).ty.ty_adt_def()?;
+                let targets =
+                    if let TerminatorKind::SwitchInt { targets, .. } = &bbd.terminator().kind {
+                        targets
+                    } else {
+                        unreachable!()
+                    };
```
xldenis, line 562: Why is this true?

creusot/src/translation/function.rs
```diff
@@ -339,30 +534,79 @@ impl<'body, 'tcx> BodyTranslator<'body, 'tcx> {
...
+                if targets.otherwise() == bb {
+                    continue;
+                }
+
+                let mut inactive_mps = self.empty_bitset();
+                on_all_children_bits(self.move_data(), discr_mp, |mpi| {
+                    inactive_mps.insert(mpi);
+                });
```
xldenis, line 570: Why is the discriminator `MovePlace` inactive? I don't really understand this bit of logic.

creusot/src/translation/function.rs
```diff
@@ -377,31 +621,141 @@ impl<'body, 'tcx> BodyTranslator<'body, 'tcx> {
...
     }
 
-    fn resolve_locals(&mut self, mut locals: BitSet<Local>) {
-        locals.subtract(&self.erased_locals);
+    fn resolve_places(
+        &mut self,
+        to_resolve: BitSet<MovePathIndex>,
+        resolved: &BitSet<MovePathIndex>,
+        assume_downcasts: bool,
+    ) {
```
xldenis, line 629: Just for future sanity, maybe add that this coalesces resolutions for places as a doc comment.

creusot/src/translation/function.rs
```diff
@@ -377,31 +621,141 @@ impl<'body, 'tcx> BodyTranslator<'body, 'tcx> {
...
+            };
+            match ty.ty.kind() {
+                TyKind::Adt(adt_def, subst) => {
+                    if adt_def.is_box() {
+                        insert(self.ctx.mk_place_deref(pl));
+                    } else if adt_def.is_enum() {
+                        if let Some(vid) = ty.variant_index {
+                            let var = adt_def.variant(vid);
+                            // TODO: honor access rights for these fields.
+                            // I.e., we should not resolve provate fields.
```
xldenis, line 683:
```suggestion
                            // I.e., we should not resolve private fields.

```

xldenis (commented):
Thank you for this massive chunk of work. It definitely fixes a bunch of issues in the `.coma` files, which is huge. 

Sorry to be such a stickler with my comments, but I figure that we should try to fix this code once and for all. Most of my comments are minor, documentation related or questions about the code, I didn't identify any visible bugs in logic, but there is at least one portion of the logic which is hard to follow.

creusot/src/translation/function.rs
```diff
@@ -107,29 +134,36 @@ impl<'body, 'tcx> BodyTranslator<'body, 'tcx> {
...
                     body,
-                    borrows.clone(),
+                    borrows,
                     with_facts.region_inference_context.clone(),
+                    mdpe,
                 );
-
+                // eprintln!("--------------- {:?}", body_id);
+                // eprintln!("{:?}", mdpe.move_data);
+                // resolver.debug();
```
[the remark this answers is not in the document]
jhjourdan: Actually, I'd prefer leave this here, as this makes it easy to see what's happening with the resolver.

creusot/src/resolve.rs
```diff
@@ -1,205 +1,283 @@
...
+            let place = self.borrow_set[bi].borrowed_place;
+            match self.move_data().rev_lookup.find(place.as_ref()) {
+                LookupResult::Exact(mp) => on_all_children_bits(&self.move_data(), mp, |mp| {
+                    frozen.insert(mp);
+                }),
+                LookupResult::Parent(mp) => drop(frozen.insert(mp.unwrap())),
+            };
         }
         frozen
     }
```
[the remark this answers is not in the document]
jhjourdan: You get a set with two elements, which correspond to the two borrowed move paths. Note that the parent move path will not be considered borrowed, and thus be resolved earlier (which will not be visible, because resolving the parent place without its children is a no-op).

Adding a comment.

creusot/src/translation/function.rs
```diff
@@ -267,16 +300,95 @@ impl<'body, 'tcx> BodyTranslator<'body, 'tcx> {
...
+                    } else {
+                        unreachable!()
+                    };
+                }
+                ProjectionElem::Downcast(_, variant) => {
+                    if first {
+                        let (adt, substs) = if let TyKind::Adt(adt, substs) = ty.ty.kind() {
+                            (adt, substs)
+                        } else {
+                            unreachable!()
```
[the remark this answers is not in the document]
jhjourdan: Well, I consider that we should use panic macros when entering this code path is actually a bug of the code (i.e., this case is thought actually impossible). And, in this particular case, I think that not having a string is fine, because the reason why this cannot panic is clear by reading the context: if we have a downcast, then the place must be an ADT.

creusot/src/translation/function.rs
jhjourdan: Well, it's not necessarilly. We initialized all the children of the discriminiator to true, and the remove whatever is active.

creusot/src/translation/function.rs
```diff
@@ -377,31 +621,141 @@ impl<'body, 'tcx> BodyTranslator<'body, 'tcx> {
...
+                TyKind::Adt(adt_def, subst) => {
+                    if adt_def.is_box() {
+                        insert(self.ctx.mk_place_deref(pl));
+                    } else if adt_def.is_enum() {
+                        if let Some(vid) = ty.variant_index {
+                            let var = adt_def.variant(vid);
+                            // TODO: honor access rights for these fields.
+                            // I.e., we should not resolve provate fields.
+                            // Problem: it's unclear what to do if we need to resolve a private
+                            // field, but not the whole struct/enum
```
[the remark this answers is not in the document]
jhjourdan: That's a possibility. But then it's not easy to describe what's allowed and not.

jhjourdan:
@xldenis CI is crashing, probably because `/target` contains obsolete files. I observed that on my machine, too, but the bug disappeared with a fresh `/target`.

This is probably a bug in the build system, and the CI cache needs to be cleared. I don't know if I'm allowed to do that.

xldenis:
You should have permissions, but I just wiped the cache and it still fails to build.

jhjourdan:
W. E. I. R. D.

xldenis:
doing `cargo clean && cargo test --test ui` replicates the issue on my machine.

xldenis (approved):

PR #1073, opened 2024-08-27
Ghost structures

arnaudgolfouse:
Some data structures usable in `ghost!` code.


jhjourdan:
I would have thought that we would reuse Mapping and Seq for this.

arnaudgolfouse:
Internally yes, but since we cannot use a method in both logic and program (here ghost), this seemed to yield a better API.
E.g. we do not need `map.remove` and `map.remove_program`

arnaudgolfouse:
I also added a minimal implementation of a `PCell` (a `Cell` that uses a ghost state to track its value), in `creusot/tests/should_succeed/ghost/pcell.rs`.

creusot/tests/should_succeed/ghost/pcell.rs
```diff
@@ -0,0 +1,158 @@
...
+    #[ensures(self.id == (^token.inner()).id)]
+    #[ensures(result == (*token.inner()).value)]
+    #[ensures(value == (^token.inner()).value)]
+    pub fn replace(&self, token: GhostBox<&mut Token<T>>, value: T) -> T {
+        self.inner.replace(value)
+    }
+
+    #[trusted]
+    #[requires(self.id == (*self_t.inner()).id)]
+    #[ensures(self.id == (^self_t.inner()).id)]
```
xldenis, line 101: Would doing a logical borrow help here? I don't know if we allow this specific syntax / case but `&mut self.id == &mut self_t.inner()`?

creusot/tests/should_succeed/ghost/pcell.rs
```diff
@@ -0,0 +1,158 @@
...
+    #[requires(other.id == (*other_t.inner()).id)]
+    #[ensures(other.id == (^other_t.inner()).id)]
+    #[ensures((^self_t.inner()).value == (*other_t.inner()).value)]
+    #[ensures((*self_t.inner()).value == (^other_t.inner()).value)]
```
xldenis, line 105: On the other hand we _should_ be able to express _this_ as a logical reborrow expression no?

creusot/tests/should_succeed/ghost/pcell.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    /// When creating two `PCell`s, we don't know that their ids are different.
+    ///
+    /// This function allows us to derive exactly that.
+    #[trusted]
+    #[pure]
+    #[ensures(*self == ^self)]
+    #[ensures(^self != *other)]
+    fn differentiate(&mut self, other: &Self) {}
+}
```
xldenis, line 25: There is no way this is sound. You definitely need the Id to be an infinitely sized type. I should be able to prove that `Id::fresh() == Id::fresh()`. You can put `#[trusted]` on the `Id` but I think it should be a wrapper around an `Int`.

creusot/tests/should_succeed/ghost/pcell.rs
```diff
@@ -0,0 +1,158 @@
...
+impl Id {
+    /// Generate a fresh id.
+    ///
+    /// To ensure that it is different from previously generated ids, see [`Self::differentiate`].
+    #[pure]
+    #[requires(true)]
+    #[ensures(true)]
+    fn fresh() -> Self {
+        Self(())
+    }
```
xldenis, line 15: What is our position on non-determinism in ghost code?

creusot/tests/should_succeed/ghost/pcell.rs
arnaudgolfouse: I'm not sure how, since this swaps `*` and `^`. Another way to write this I came up with is:
```rust
#[ensures({
    let self_val = &mut self_t.inner().value;
    let other_val = &mut other_t.inner().value;
    (*self_val, ^self_val) == (^other_val, *other_val)
})]
```
Which seems strictly worse.

creusot/tests/should_succeed/ghost/pcell.rs
arnaudgolfouse: Nope, this crashes Creusot with an obscure error... I think we can't create a mutable borrow "out of thin air" in pearlite, we need to reborrow something.

creusot/tests/should_succeed/ghost/pcell.rs
arnaudgolfouse: Indeed, although `Id` does not need to be infinitely sized, only to contain more than one value (e.g. `bool` is enough). I think @jhjourdan was unsure about `Int`, as it could give the solvers to (wrong) idea to use arithmetic ?

creusot/tests/should_succeed/ghost/pcell.rs
jhjourdan: I would indeed have thought that ghost code should be deterministic. At some point we have imagined that ghost functions could be turned into logical functions under appropriate assumptions.

creusot/tests/should_succeed/ghost/pcell.rs
jhjourdan: `()` and `bool` are not better than `Int`, they are all concrete types. I proposed to use an *abstract* type here (i.e., using `#[trusted]` on the type), so that solvers only have the theory of equality on this type.

jhjourdan:
> Internally yes, but since we cannot use a method in both logic and program (here ghost), this seemed to yield a better API.
> E.g. we do not need map.remove and map.remove_program

We need to discuss that. There is indeed the problem of the naming logical functions vs. ghost (program) functions. But this problem already exists for many types: for example, we have `cmp_log` and `cmp`, etc... Ideally, we should use two different name spaces for logical and program functions, but I'm not sure how to do that.

On the other hand, I'm not sure having three different types for the notion of sequence or mapping (logical, ghost and program) is a good idea. This is, first, going to be confusing, and second may lead to limitations: if we want to use ghost functions in logic, then we would need a way to convert from `GhostVec` to `Seq`. And then, what if we have sequances of sequences, etc ?

creusot/tests/should_succeed/ghost/pcell.rs
xldenis: Yea, I think that would be best, though we should ideally come up with a better way of doing that.

jhjourdan: Once we will honor `pub` annotations on fields of structures, that will be easy.

arnaudgolfouse: All right, am I right to put `#[trusted]` on `Id::fresh` then ?

xldenis: no on the `Id` type itself. This causes creusot to translate it to an uninterpreted type.

arnaudgolfouse:
As we discussed,  `Seq` and `FMap` are now used as the base type for ghost structures

arnaudgolfouse:
There are some potentially controversial changes here:
- For convenience, I added a hack to dereference GhostBox in pearlite
- I renamed `Seq::len` to `Seq::length` and `FMap::len` to `FMap::length`. This is very disruptive, maybe the ghost operations should be renamed instead.

Armael:
Regarding naming, I'm wondering if we should leave the logic functions as before with their intuitive names, and have `*_ghost` functions for the ones in the ghost code. (With the rationale that `Seq` is used more often in specifications than ghost code.)

Somewhat unrelated, but it would also be nice to have functions to push/pop (alt cons/uncons) at the beginning of the sequence.

arnaudgolfouse:
I think that would be better, yes. I tried it like this to see how it would work out with `length` et co., and I was really displeased with the result, so I'll revert it 😅

jhjourdan (approved):
Approved, provided we report `PCell` to another PR.

PR #1079, opened 2024-08-31
Cleanup in trait resolution.

jhjourdan:


creusot/tests/should_fail/bug/222.coma
```diff
@@ -125,10 +120,9 @@ module C222_UsesInvariant
       | s1 = -{inv'0 _5.final}- s2
       | s2 = take'0 {_5} (fun (_ret':Option'0.t_option t) ->  [ &_4 <- _ret' ] s3)
       | s3 = {[@expl:type invariant] inv'0 _4} s4
-      | s4 = -{resolve'0 _4}- s5
```
xldenis, line 128: This is weird no?

jhjourdan: This is expected. We use the new criterion in `traits.rs` to know whether the implementation of `Resolve` is trivial or not. Witht this new criterion, we know that the implementation is trivial here, so we do not emit the resolution.

PR #1084, opened 2024-09-06
mut and ref mut binders are forbidden in pearlite.

jhjourdan:


xldenis (approved):
This seems ok, but what context did this occur in?

jhjourdan:
At some point, we may use `mut ref` to do mutable reborrowing under enum types.

jhjourdan:
> This seems ok, but what context did this occur in?

I'm reading the code of logical reborrowing and final borrows, and I fix bugs or missing errors like this.

PR #1089, opened 2024-09-11
Final borrows fixes

jhjourdan:


creusot/tests/should_succeed/iterators/02_iter_mut/why3session.xml
```diff
@@ -42,17 +42,17 @@
...
   <goal name="vc_next.2" expl="postcondition" proved="true">
   <transf name="split_vc" proved="true" >
    <goal name="vc_next.2.0" expl="postcondition" proved="true">
-   <proof prover="3"><result status="valid" time="0.066377" steps="9131"/></proof>
+   <proof prover="3" timelimit="1"><result status="valid" time="0.066377" steps="9048"/></proof>
    </goal>
   </transf>
   </goal>
   <goal name="vc_next.3" expl="postcondition" proved="true">
-  <proof prover="0"><result status="valid" time="0.161973" steps="3021"/></proof>
```
xldenis, line 55: I wish we had a way to rigorously evaluate performance impact because I get the impression we significantly increased the time on this proof, no? or is that a machine / prover version artifact?

jhjourdan: This change in performance is caused (I beleive) by one change:

```
-    #[ensures(forall<i : _> 0 <= i && i < result.len() ==> *result[i] == self[i])]
-    #[ensures(forall<i : _> 0 <= i && i < result.len() ==> ^result[i] == (^self)[i])]
+    #[ensures(forall<i : _> 0 <= i && i < result.len() ==> result[i] == &mut self[i])]
+    // TODO: replace with a map function applied on a sequence
     fn to_mut_seq(&mut self) -> Seq<&mut T> {
```

This is only weakly related to final borrows, so I may do this in a separate PR if you prefer.

It apparently causes issues with triggers so that alt-ergo is no longer able to prove the goal, and we need to use CVC5 instead. I still think this change is worth it because:
   - The new spec is shorter and conveys more information
   - While the proof of `all_zero` in `02_iter_mut` indeed needs fixing, the one in `03_std_iterator`, which uses the slightly different version of `IterMut` present in creusot-contracts, hasn't changed.

Also note that the timings may be biased since I ran the tests on my heavily loaded desktop.

xldenis: What about with z3?

jhjourdan: Sorry, I was thinking about `vc_all_zero.7` bellow. Z3 finishes  on `vc_all_zero.7` in about the same time (0.5s).

As for `vc_next.3`, I'm not sure what to think. It's quite common to have one or two goals which are significantly slower like this.

jhjourdan: Then it's a matter of setting the right triggers for the spec of `to_mut_seq`. We don't support triggers for lemma functions currently.

arnaudgolfouse:
The `indexing` test in `should_fail/final_borrows.rs` should be removed, since this is now supported

creusot/src/analysis/not_final_places.rs
```diff
@@ -230,6 +224,11 @@ impl<'tcx> AnalysisDomain<'tcx> for NotFinalPlaces<'tcx> {
     fn initialize_start_block(&self, _: &mir::Body<'tcx>, _: &mut Self::Domain) {}
 }
 
+// The NotFinalPlaces analysis computes, for each location, places which either:
+// - do not contain a mutable borrow deref and may be moved or borrowed in the future
```
arnaudgolfouse, line 228: We should have a precise example of this, or even a test added to `should_fail/final_borrows.rs`
Something like 
```rust
// `bor` is a place that does not contain a mutable deref, but it is move after it has been reborrowed.
pub fn move_place_without_deref<T>(bor: &mut T) {
    let b1 = &mut *bor;
    proof_assert!(b1 == bor);
    f(bor);
}
#[requires(true)]
#[ensures(true)]
fn f<T>(x: &mut T) {}
```

arnaudgolfouse (approved):
The modifications look good to me

creusot/src/analysis/not_final_places.rs
jhjourdan: Too late, sorry.

Feel free to add a test in your future PR. But note that in your example, `f(bor)` is actually a reborrow of `bor`, and not a move. To force a move, the simplest is to wrap the borrow under a `Box`.

PR #1095, opened 2024-09-14
Various fixes (more or less) related to type invariants

jhjourdan:
- Use invariants written in creusot-contracts for builtin types. This is simpler than generating terms in creusot code.
- Lift Sized constraints in many places in creusot-contracts
- Force #[trusted] for any objects marked as builtin
- inv and invariants are prophetic predicates
- Make not loading creusot-contracts a hard error, and simplify code in Creusot accordingly (no need to consider cases where diag items are not found.)
- Add an type invariant for GhostPtrToken


jhjourdan:
I'm sorry, this is a bit of a "pot-pourri". I've these fixes on a branch for a long time, and they really need to be merged at some point.

xldenis:
> Make not loading creusot-contracts a hard error, and simplify code in Creusot accordingly (no need to consider cases where diag items are not found.)

I'm actually not a huge fan of this, though I've often had the same thought myself. The reason being if someone tries out creusot on a fresh rust project now the first thing they get is an error.

jhjourdan:
Even before this PR, the code was full of things like `get_diagnostic_item(...).unwrap()`. I have nothing against having a mode where we do not require `creusot-contracts`, but then we need to fix the code before in another PR.

jhjourdan:
If someone runs Creusot on random Rust code, I prefer if she has a clear error message telling him to do something simple (adding the crate), rather than a panic resulting from a failing `unwrap`.

PR #1096, opened 2024-09-16
Disable automatic insertion of type invariant assertions in relevant places.

jhjourdan:
- We no longer add invariant guards to quantifiers
- We no longer generate invariant pre/post-conditions for logical functions
- We have two new attributes #[creusot::open_inv] and #[open_inv_result] for manually disabling invariant insertion for parameters/return values.


jhjourdan:
We already discussed all that with @xldenis. Since this changes a lot of fils, this has a high probability of conflicts, so I'm merging quickly. Of course, you can still complain, I'll be happy to revert.

PR #1108, opened 2024-09-19
Make translation of names injective

### LINKED ISSUE #1107: snake_case to CamelCase conversion is not injective
Names of libraries, files, modules, and functions are converted from their conventional snake_case to CamelCase in the generated code (`module MyLibrary_MyFile_MyModule_MyFunction`). If the original names are unconventionally CamelCase, that creates collisions.

I ran into this while thinking about how to deal with names in a language server but this can likely also be exploited for unsoundness.

Lysxia:
Closes #1107

- There are some `snake_case` conversions remaining somewhere in `creusot/src/backend/dependency.rs`, are they related?
- The `why3session.xml` files need to be updated too. Is there a tool for that?

### DIFF
--- a/creusot/src/backend/dependency.rs
+++ b/creusot/src/backend/dependency.rs
@@ -1,4 +1,3 @@
-use heck::ToSnakeCase;
 use rustc_ast_ir::visit::VisitorResult;
 use rustc_hir::{def::DefKind, def_id::DefId};
 use rustc_macros::{TypeFoldable, TypeVisitable};
@@ -9,7 +8,7 @@ use rustc_type_ir::{fold::TypeFoldable, visit::TypeVisitable, AliasTyKind, Inter
 use crate::{
     ctx::TranslationCtx,
     translation::traits,
-    util::{self, item_symb, ItemType},
+    util::{self, item_symb, translate_accessor_name, type_name, value_name, ItemType},
 };
 
 use super::{
@@ -172,22 +171,19 @@ impl<'tcx> Dependency<'tcx> {
 
     pub(crate) fn base_ident(self, tcx: TyCtxt<'tcx>) -> Symbol {
         match self {
-            Dependency::Type(ty) => {
-                let nm = match ty.kind() {
-                    TyKind::Adt(def, _) => {
-                        item_symb(tcx, def.did(), rustc_hir::def::Namespace::TypeNS)
-                    }
-                    TyKind::Alias(_, aty) => tcx.item_name(aty.def_id),
-                    TyKind::Closure(def_id, _) => {
-                        item_symb(tcx, *def_id, rustc_hir::def::Namespace::TypeNS)
-                    }
-                    _ => Symbol::intern("debug_ty_name"),
-                };
-                Symbol::intern(&nm.as_str().to_snake_case())
-            }
+            Dependency::Type(ty) => match ty.kind() {
+                TyKind::Adt(def, _) => item_symb(tcx, def.did(), rustc_hir::def::Namespace::TypeNS),
+                TyKind::Alias(_, aty) => {
+                    Symbol::intern(&type_name(tcx.item_name(aty.def_id).as_str()))
+                }
+                TyKind::Closure(def_id, _) => {
+                    item_symb(tcx, *def_id, rustc_hir::def::Namespace::TypeNS)
+                }
+                _ => Symbol::intern("debug_ty_name"),
+            },
             Dependency::Item(_, _) => {
                 let did = self.did().unwrap().0;
-                let base = match util::item_type(tcx, did) {
+                match util::item_type(tcx, did) {
                     ItemType::Impl => tcx.item_name(tcx.trait_id_of_impl(did).unwrap()),
                     ItemType::Closure => Symbol::intern(&format!(
                         "closure{}",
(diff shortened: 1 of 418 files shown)


jhjourdan:
>     * There are some `snake_case` conversions remaining somewhere in `creusot/src/backend/dependency.rs`, are they related?

Yes, I think they should also use a new encoding. The difference here is that they have to start with a lower case letter.

> 
>     * The `why3session.xml` files need to be updated too. Is there a tool for that?

If you run `cargo test --test why3`, you tell Why3 to replay all the proofs. It should be able to recover the proofs even though the identifiers have changed. If the repay succeeds, then it should write "obsolete" in orange, and it overrides the session files on the fly.

In addition, please run the `fmt` script to normalize the formatting of Rust code. If you find this frustrating, please complain to @xldenis, we may perhaps make him change his mind ;)

xldenis:
> In addition, please run the fmt script to normalize the formatting of Rust code. If you find this frustrating, please complain to @xldenis, we may perhaps make him change his mind ;)

Nice try! but I would recommend just setting up format-on-save so you no longer need to think about it :)

xldenis:
> The why3session.xml files need to be updated too. Is there a tool for that?

Unfortunately, apart from `grep` no.

xldenis:
> There are some snake_case conversions remaining somewhere in creusot/src/backend/dependency.rs, are they related?

Yep!

creusot/src/util.rs
```diff
@@ -248,31 +248,75 @@ pub(crate) fn module_name(tcx: TyCtxt, def_id: DefId) -> Symbol {
...
+fn push_translate_name(n: String, dest: &mut String) -> (){
+    let mut chars = n.chars().peekable();
+    // Escape initial underscore
+    if chars.peek() == Some(&'_') {
+        let _ = chars.next();
+        dest.push('q');
+        dest.push('y');
+        dest.push('9');
+        dest.push('5');
+        dest.push('z');
```
xldenis, line 264: You should be able to use `.extend("qy95z")` here.

creusot/src/util.rs
```diff
@@ -248,31 +248,75 @@ pub(crate) fn module_name(tcx: TyCtxt, def_id: DefId) -> Symbol {
...
+    while let Some(c) = chars.next() {
+        let is_qy = c == 'q' && chars.peek() == Some(&'y');
+        if c == '_' {
+            match chars.peek() {
+                None | Some('_') => {
+                    dest.push('q');
+                    dest.push('y');
+                    dest.push('9');
+                    dest.push('5');
+                    dest.push('z');
```
xldenis, line 275: same,

creusot/src/util.rs
```diff
@@ -248,31 +248,75 @@ pub(crate) fn module_name(tcx: TyCtxt, def_id: DefId) -> Symbol {
...
+            }
+        } else if c.is_ascii_alphanumeric() && !is_qy {
+            dest.push(c);
+        } else {
+            dest.push('q');
+            dest.push('y');
+            // convert code point c to decimal
+            for d in format!("{}", c as u32).chars() {
+                dest.push(d);
+            }
```
xldenis, line 287: `.extend(format!("qy{}z", c as u32))`

creusot/src/util.rs
Lysxia: I found `.push_str()` which is similar. Do you have any preference vs `.extend()`?

creusot/src/backend/dependency.rs
```diff
@@ -172,22 +171,17 @@ impl<'tcx> Dependency<'tcx> {
...
-                Symbol::intern(&nm.as_str().to_snake_case())
-            }
+            Dependency::Type(ty) => match ty.kind() {
+                TyKind::Adt(def, _) => item_symb(tcx, def.did(), rustc_hir::def::Namespace::TypeNS),
+                TyKind::Alias(_, aty) => tcx.item_name(aty.def_id),
+                TyKind::Closure(def_id, _) => {
+                    item_symb(tcx, *def_id, rustc_hir::def::Namespace::TypeNS)
+                }
+                _ => Symbol::intern("debug_ty_name"),
+            },
```
Lysxia, line 181: Here the identifiers already seem to have been initial-lowercased (as far as the test suite can tell)

creusot/src/util.rs
xldenis: thats good. I forgot that `String` has a weird api there for whatever reason.

xldenis:
@Lysxia if you look at the `why3` test suite (`cargo test --test why3`
) you'll see some files which contain syntax errors. If you can fix those, someone else can help you update the obsolete proof sessions if needed. You can avoid running the provers to find syntax issues faster using `cargo test --test why3 -- --replay=none`

Lysxia:
I fixed the syntax errors. Because the names of the theories changed, why3 sees a lot of detached goals so the sessions are quite broken. We can probably generate a sed script to do the renaming in the session files, unless someone has a better idea.

I removed all uses of `to_camel_case` and `to_snake_case`, but there are still some `to_lowercase` or `make_ascii_lowercase` that make me a bit uneasy.

xldenis:
If you run `cargo test --test why3` it should fix most of the files automatically, but yea otherwise a sed may be the most efficient manner of fixing

Lysxia:
I patched the Why3 sessions and there are only 4 failing ones left on my machine:
- `list_reversal_lasso`, trying to `remove` no longer existing assumptions
- `red_black_tree`, timeouts
- `selection_sort_generic`, timeouts
- `sparse_array`, broken `pose`

jhjourdan:
Seems like the only problem is with `sparse_array`, with some detached goal that need to be cleaned. I'm hitting the "Enable auto-merge" button so that once you'll have fixed that, this will be merged.

PR #1109, opened 2024-09-20
[WIP] Type invariant fixes

jhjourdan:
See individual commits.


xldenis (approved):

jhjourdan:
The remaining TODO is
1- Do the static analysis that adds loop invariants for type invariants automatically.
2- Avoid making type parameters/type aliases trivial type invariant.

PR #1111, opened 2024-09-20
Fix typo in error message: 'Could _not_ determine value of constant'

Lysxia:


jhjourdan:
The CI failure is really weird. The goal in question finishes in 3.5s on my machine, but apparently takes forever on CI. This clearly looks like a bug somewhere.

In the meantime, @Lysxia you could perhaps use a much longer timeout (30s?) for this specific goal.

PR #1114, opened 2024-09-21
Basic Type Privacy

xldenis:
This PR respects the visibility of std types.

1. If a constructor is private then we dont generate the coma destructor.
2. If a type is from std, we may simplify its definition in the following cases:
	a. private fields are replaced by `Opaque.hidden_field`. This is safe since `std` fields can't be leaked through `open` predicates.
	b. if a constructor is private we skip generating the constructor entirely. Currently this is an all-or-nothing process in Rust: either all constructors are private or none are. 

Frustratingly, for `struct` types, even if all fields are private, you can do a pattern match on them so we cannot hide the constructors in that scenario. 

I expect this change to have no impact on performance, though a part of me secretly hopes it has some.


jhjourdan:
> private fields are replaced by Opaque.hidden_field. This is safe since std fields can't be leaked through open predicates.

Is this sound, in general? This seems to imply that we assume that the types of these fields always have the same cardinality.

IMHO, it would be better to use a fresh abstract type for each private field. If there are several private fields, then we can use only one abstrat type.

xldenis:
I think you're right, unfortunately that requires a notably more complex implementation so it'll have to wait for another time.

PR #1115, opened 2024-09-23
Try Bitwise

### LINKED ISSUE #743: Support for bitwise operations
I am trying to verify some code that contains bitwise operations like AND, OR, SHIFT, NEGATE that are currently not supported. How much effort will it take to support these types of operations?

### LINKED ISSUE #1124: Implement casts in pearlite
The pearlite code `1 as i32` should not panic, but instead do what is expected.

laurentder:
PR to have a first test with the current implementation of integer representations in BitsVector form.
Do not merge with the master for the moment ;-)

Fixes #743
Fixes #1124 as a side effect.


jhjourdan:
Some notes about this work:

- It would be probably necessary to have an option to opt-in this.
- Try to use mathematical bounded integers and convert them to BV instead of doing the convert
- Should we use type invariants to bound machine integers (then it is easy to have arithmetics on machine integeres in pearlite)

jhjourdan:
We currently use a different branch. Closing this PR.

PR #1117, opened 2024-09-23
Fix inferred invariants

jhjourdan:
When reading the code for inferred invariants prophecies in loops (in preparation to add an analysis to inferred type invariants as loop invariants), I found several issues in the code. Notably, it did not support nested loops nor nested borrows, and did not consider overriding the parent place of a borrow as an override of the borrow itself.


xldenis:
> Notably, it did not support nested loops nor nested borrows, and did not consider overriding the parent place of a borrow as an override of the borrow itself.

The lack of support for nested looks is somewhat surprising, I didn't expect that.

xldenis (approved):
Not much to say about the transformation/optimization. One thing I wonder about is `targets_mut`, this seems like a strange API which is rather difficult to use correctly in general. However, don't let that block merging.

PR #1119, opened 2024-09-24
Type invariants: type parameters and aliases are not trivial

jhjourdan:
The latest PR on type invariants. This should fix the remaining known unsoundness around type invariants.

The most important change is stopping assuming type invariants of type parameters and unconstrained aliases trivial. This is needed for soundness, but creates quite a few problems.

Most notably:
- Proofs of iterators are made much more complicated than previously. One of the reasons is that solvers have a hard time unfolding some type invariants. This is related to the encoding of polymorphism in Why3, to the way projections are defined by Creusot, to the fact that type invariants are using axioms instead of mutually recursive definitions. Improvements to any of these limitations can solve this specific problem.
- Specifications involving quantifiers over abstract types are made more complex, because they typically require guarding with type invariants.
- We need to add `#[invariant(inv(...))]` to many loops to make sure that type invariants are maintained during loop executions. These should be removed by a static analysis determining when places contain valid values. I don't have time to do that right now, but it would really help simplifying generated Coma code, by removing a lot of type invariant assertions, in particular at resolution points.

But, at least, the soundness hole is filled, and we can say that we have sound type invariant in Creusot, finally.

Also, I changed some examples to use type invariants instead of manually using pre/post-conditions.


jhjourdan:
@xldenis : As with previous PRs: you are welcome to review this if you want, otherwise I'll merge.

This code is not very satisfactory, because of the reasons explained above. However, it fixes a long-standing unsoundness, and I have some plans to improve parts of the implementation. And we have a deadline approaching.

xldenis:
> This code is not very satisfactory, because of the reasons explained above. However, it fixes a long-standing unsoundness, and I have some plans to improve parts of the implementation. And we have a deadline approaching.

I think you can merge, I don't have a lot of time for a detailed review right now. I would ask that you open issues for the 3 points you raise above though.

jhjourdan:
Ok, will do for points 1 and 3. Point 2 is rather a regression which I don't really have a plan to fix.

xldenis:
> Point 2 is rather a regression which I don't really have a plan to fix.

For that one I think a documentation adjustment is more appropriate.

PR #1120, opened 2024-09-25
Rework the final borrow analysis

arnaudgolfouse:


arnaudgolfouse:
This mostly works, but its a bit too overaccepting: the following borrow
```rust
pub fn unnest<'a, 'b: 'a, T>(x: &'a mut &'b mut T) -> &'a mut T {
    &mut **x
}
```
Is final, while it should not be (because the borrow it borrows from might not be resolved). I tried marking only "resolved" borrows as final in `statement.rs`, but it does not work the way I did it: @jhjourdan do you have an idea of how it should be done?

jhjourdan:
LGTM, thanks!

PR #1123, opened 2024-09-25
Use more stable naming of impls

Lysxia:
Rust internally numbers impls to disambiguate them, so it generates names like `{impl#42}::f` to refer to the function `f` in the 42-th `impl` block of some file. The problem is that those numbers appear in Why3 sessions, and whenever you insert a new `impl` in a source file it shifts all the numbers and breaks the Why3 sessions.

Instead we disambiguate impls by hashing the "impl subject" which is the type and/or trait that the impl is about: for `impl MyType<FooBar>` we hash `MyType<FooBar>` and use it as the name of the impl.

TODO:

- Right now, I'm somehow getting different hashes every time I re-run creusot on the same file
- Detect (improbable) hash collisions. Relatedly, here's Rust doing something similar: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/def_id/struct.DefPathHash.html; relevant quote:

    > There is a very small chance of hash collisions, which would mean that two different DefPaths map to the same DefPathHash. Proceeding compilation with such a hash collision would very probably lead to an ICE, and in the worst case lead to a silent mis-compilation. The compiler therefore actively and exhaustively checks for such hash collisions and aborts compilation if it finds one.


xldenis:
>Right now, I'm somehow getting different hashes every time I re-run creusot on the same file

You will probably want to use the `StableHash` mechanism present in rustc which exists to solve similar issues (ie hash stability across compilation sessions).

Lysxia:
- The "easy" way to do hash collision detection would be to create a global hashtable. How do you do that in Rust? What is a better solution? Or should we postpone handling this issue?
- I only did this hashing for `impl`. Closures are the other construct with a disambiguating counter that appears in tests. However the counter is reset in each function, so the issue is lesser. It's also harder to find features to hash that would distinguish closures without basically hashing the whole body of the function, making the result more fragile than with a dumb counter.

xldenis:
> The "easy" way to do hash collision detection would be to create a global hashtable. How do you do that in Rust? What is a better solution? Or should we postpone handling this issue?

Hmm I think realistically we can postpone this issue, alternatively we can try to use `DefPathHash` https://doc.rust-lang.org/beta/nightly-rustc/rustc_hir/def_id/struct.DefPathHash.html which the compiler builds using `HashStable` and already checks for collisions.

Lysxia:
If you mean using the `DefPathHash` of the impl, I don't think that's what we want because the hash depends on the numbering of impl blocks that we are trying to avoid.

xldenis:
Ah I didn't realize that but it makes sense that it would. 

I do think that upon a collision we have much less issues than rustc: it would lead to duplicate module definitions which why3 would reject outright. For this reason I don't think we need to worry too much about catching that, though we could detect when were done translating by scanning all module names I suppose.

Lysxia:
Good point. Then this PR is ready.

@jhjourdan Does that look good to you?

jhjourdan:
Yeah. You have a 64 bits hash, so in order to have a non-negligible probability of collision, you would have to have 2^32 modules in the coma file.

If a user tries to adversarially find a collision, well, it will make Why3 crash. She calls for problems, she gets problems.

xldenis (approved):
LGTM.

Only comment I can think of is that at a future date, the hashing of `DefId`s could turn out to be expensive, but it would be easy to throw up a thread-local cache around that. Again, no need to change anything for now.

PR #1126, opened 2024-09-26
Bump why3 dependency

jhjourdan:
Some proofs (including red_black_tree) are made simpler, thanks to `eval_match`.

Sadly, Why3 crashes on some tests, so we will need to wait for a patch.


Armael:
I wonder if why3 should include the coma files from creusot's testsuite in its own CI...

PR #1127, opened 2024-09-26
Fix typo in why3/README

Lysxia:
:smile:

### DIFF
--- a/why3/README.md
+++ b/why3/README.md
@@ -1,7 +1,7 @@
 # Why3 
 
 A Rust library furnishing the Why3 AST used by the Creusot verifier, and its associated pretty printer. 
-The library supports three input formats to why3: WhyML (`.mlw`), coma (`.coma`) and Coma (`.coma`).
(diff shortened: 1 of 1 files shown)


why3/README.md
```diff
@@ -1,7 +1,7 @@
 # Why3 
 
 A Rust library furnishing the Why3 AST used by the Creusot verifier, and its associated pretty printer. 
-The library supports three input formats to why3: WhyML (`.mlw`), coma (`.coma`) and Coma (`.coma`).
+The library supports three input formats to why3: WhyML (`.mlw`), Coma (`.coma`).
```
Armael, line 4: "three" -> "two"?

Lysxia: Good catch!

xldenis (approved):
I actually want to remove the mlcfg stuff since I could take advantage of that for some other serious cleanups and improvements.

why3/README.md
xldenis: technically it _is_ 3 since its `mlcfg`, `coma` and `mlw`.

PR #1128, opened 2024-09-26
Condense let-spans

Lysxia:
image


xldenis (approved):
Fair enough. Assuming you fix the ci issues lgtm.

PR #1129, opened 2024-09-26
Remove the default implementation of the Resolve trait.

jhjourdan:
Instead, we have a logic function resolve, which is equal to Resolve::resolve if an instance exists, and true if it does not exist.


creusot/src/backend.rs
```diff
@@ -202,10 +206,30 @@ impl<'tcx> Why3Generator<'tcx> {
...
+                        .unwrap();
+                    let substs = GenericArgs::identity_for_item(self.tcx, def_id);
+                    let arg = Term::var(pre_sig.inputs[0].0, pre_sig.inputs[0].2);
+
+                    lower_pure(self, &mut names, &Term::call(self.tcx, method, substs, vec![arg]));
+
+                    let deps = names.provide_deps(self, GraphDepth::Shallow).1;
+                    self.dependencies.insert(def_id.into(), deps);
+
+                    TranslatedItem::Logic { proof_modl: None }
```
xldenis, line 227: god i cant wait for us to kick out all these dependency calculations.

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -178,15 +179,40 @@ impl<'tcx> SymbolElaborator<'tcx> {
...
+        let mut sig = named_sig_to_why3(ctx, names, name, &pre_sig, def_id);
 
         if CloneLevel::Signature == level_of_item {
             return val(ctx, sig, kind);
         } else if CloneLevel::Contract == level_of_item {
             return val(ctx, sig, kind);
         };
 
-        if item.is_hacked() || ctx.is_logical(def_id) {
+        if let Dependency::Item(did, substs) = item
```
xldenis, line 190: you don't need the `if let` here on line 140 we extract the id and substs for you already.

xldenis (commented):
We should remove the `min_specialization` feature from `creusot-contracts` as well.

PR #1132, opened 2024-09-27
Simple proof of Filter

xldenis:
This will help provide more examples and test cases for the paper. It turns out that filter
is both *harder* and *easier* than I expected, and I think actually pushed the limits of what we know how to do well with creusot in interesting and useful ways. 

In no particular order here are some aspects of the proof I found notable:

1. Specifying the behavior of sequences of calls to closures is tricky, even when those closures are restricted to pure, pre-condition free contexts. If you write a loop invariant like `forall i. i \in produced ==> iter.postcondition(i, false)` this invariant won't actually be provable for two reasons.
	a. Logical reborrowing allows you to state this property, but the prophecy for `iter.func` will be taken from the prophecy of `iter`, that is the value of `iter` after the whole loop has finished. What you actually want is the prophecy for "iter up to the i-th iteration". This is really awkward and counter-intuitive, it *feels* like this spec should work. 
	b. Borrow Ids! since `postcondition_mut` is specified over a borrow, even if you know that the current and final states of the borrow are the same, if the ids are different you won't be able to prove the inductive step of your invariant. 
	c. A corrollary to this issue is that even if you know that the state of `f` doesn't change (ie: `* f == ^ f`), `let result = f.call_mut(foo) ; proof_assert!(f.postcondition_mut(foo, result));` is not generally provable. This seems undesriable...
2. Postconditions can be 'surprisingly' weak. We only allow reasoning about closures using `pre` and `post` predicates, but this can be very weak, for example, if you know that `f.post(i, true)` you *can't conclude `!f.post(i, false)`! This of course could occur if a closure had the spec: `ensures { true }`. This necessarily a weakness of creusot (unless we had some way to talk about non-determinism). But again a trap to stumble into
3. The proofs rely on fairly sophistcated existential instantiations, notably for transitivity. Its unfortunate because apart from these existentials, the proof is trivial. Once again, I wish we had some mechanism to provide 'tactics' here. A complication is that the existential i introduced depends on values obtained from existentials in the hypotheses. Could things be rewritten using ghost code?
4. Unlike what we had discussed years ago with JH, the scattered subsequence part of `Filter` actually proved to be the simplest part of the whole thing!
5. This might be niche, but two things occured to me working on iterators again:
     a. I want an `.invariant(logical_closure)` adapter, which can be used like a loop invariant or cut, and effectively "summarize" the imporant information of the iterator chain to this point. 
     b. I want an easier way to talk about the impact of iterator chains. Something like filter should be described as: `i ~{a}~> i' -> f(a) = true <-> result ~{a}~> result'` I guess I want some easier way to describe iterators in a "point-to-point" manner rather than talking about mappings between whole sequences.


xldenis:
Oh also, I made modules emit `meta "compute_max_steps" 1_000_000` since I wanted to use that transform to make reading tasks simpler. I can extract this to a separate PR if people care.

PR #1133, opened 2024-09-27
Structural resolve

xldenis:
This adds the long missing "VC" for `resolve` removing the need to mark this trait `#[trusted]`. It adds a new `structural_resolve` built-in function and adds a lemma to the `Resolve` trait. Unfortunately, this currently has the side-effect of raising the boilerplate to `Resolve`, but if we were to fix the long standing issues with default functions and traits, that would disappear. 

The idea of structural resolve is that it performs a shallow unfolding to the definition of `resolve` for a type based on its definition, that is:

```
structural_resolve<struct T(A,B)>(x) = resolve<B>(x.0) /\ resolve<B>(x.1)
structural_resolve<enum T { A, B}>(x) = match x { A => resolve<A>(..), B => resolve<B>(..) }
structural_resolve<&mut T>(x) = ^ x == *x
structural_resolve<Box<T>>(x) = resolve(*x)
structural_resolve<Snapshot<T>>(x) = True
structural_resolve<#[trusted] Ty>(x) = True
structural_resolve<_>(x) = True
```

Using this we add a lemma requiring that `shallow_resolve(x) ==> resolve(x)`. In practice these two are equal, but there are rare cases where they aren't. This VC ensures that the user provided definition cannot be stronger than the structurally implied one (modulo `#[trusted]` for types like `Vec<T>`).


creusot-contracts/src/resolve.rs
```diff
@@ -1,11 +1,17 @@
...
     #[predicate(prophetic)]
     #[rustc_diagnostic_item = "creusot_resolve_method"]
     fn resolve(self) -> bool;
+
+    #[logic(prophetic)]
+    #[requires(structural_resolve(&self))]
+    #[ensures(self.resolve())]
+    fn resolve_coherence(self)
+    where
+        Self: Sized;
```
jhjourdan, line 14: We should *not* require this. If this is a problem, then let's use a shared borrow.

creusot-contracts/src/ghost_ptr.rs
```diff
@@ -284,13 +287,22 @@ impl<'a, T: ?Sized> DerefMut for GhostPtrTokenMut<'a, T> {
...
 
-#[trusted]
 impl<'a, T: ?Sized> Resolve for GhostPtrTokenMut<'a, T> {
-    #[predicate(prophetic)]
     #[open]
+    #[predicate(prophetic)]
     fn resolve(self) -> bool {
         self.cur() == self.fin()
     }
+    #[trusted]
```
jhjourdan, line 296:
```suggestion
    }
    
    #[trusted]
```

creusot-contracts/src/resolve.rs
```diff
@@ -15,29 +21,86 @@ pub fn resolve<T: ?Sized>(_: &T) -> bool {
     true
 }
 
-#[trusted]
-impl<T1, T2: ?Sized> Resolve for (T1, T2) {
+impl<T1, T2> Resolve for (T1, T2) {
```
jhjourdan, line 24: Why do your remove the `?Sized` constraint?

creusot-contracts/src/std/deque.rs
```diff
@@ -53,13 +53,22 @@ impl<T, A: Allocator> IndexLogic<usize> for VecDeque<T, A> {
...
 
-#[trusted]
 impl<T> Resolve for VecDeque<T> {
-    #[predicate(prophetic)]
     #[open]
+    #[predicate(prophetic)]
     fn resolve(self) -> bool {
         pearlite! { forall<i : Int> 0 <= i && i < self@.len() ==> resolve(&self[i]) }
     }
+    #[trusted]
```
jhjourdan, line 62:
```suggestion
    }
    
    #[trusted]
```

creusot-contracts/src/std/iter/cloned.rs
```diff
@@ -15,14 +15,21 @@ impl<I> ClonedExt<I> for Cloned<I> {
...
 impl<I> Resolve for Cloned<I> {
     #[open]
     #[predicate(prophetic)]
     fn resolve(self) -> bool {
-        pearlite! {
-            resolve(&self.iter())
-        }
+        resolve(&self.iter())
+    }
+    #[trusted]
```
jhjourdan, line 24:
```suggestion
    }
    
    #[trusted]
```

creusot/src/backend/dependency.rs
```diff
@@ -23,7 +23,10 @@ use super::{ty_inv::tyinv_head_and_subst, PreludeModule, TransId};
 pub(crate) enum Dependency<'tcx> {
     Type(Ty<'tcx>),
     Item(DefId, GenericArgsRef<'tcx>),
+    // Type invariants and structual resolution expressions
+    // are identified by a "skeleton type" aka type with identity substs.
```
jhjourdan, line 27: In `Dependency`, we use the full type. It's in `TransId` that we use the skeletton type.

creusot/tests/should_succeed/red_black_tree.rs
```diff
@@ -233,7 +276,7 @@ fn cpn(c: Color, l: CP, r: CP) -> CP {
 
 impl CP {
     #[predicate]
-    fn match_t<K, V>(self, tree: Tree<K, V>) -> bool {
+    fn match_t<K: DeepModel, V>(self, tree: Tree<K, V>) -> bool {
```
jhjourdan, line 279: No need for this diff.

creusot/tests/should_succeed/red_black_tree.rs
```diff
@@ -178,6 +178,49 @@ impl<K: DeepModel, V> ShallowModel for Tree<K, V> {
     }
 }
 
+impl<K: DeepModel, V> Resolve for Tree<K, V> {
+    #[open(self)]
+    #[predicate(prophetic)]
+    fn resolve(self) -> bool {
+        pearlite! {
+            forall<k: K, v: V> self.has_mapping(k.deep_model(), v) ==> resolve(&v)
```
jhjourdan, line 186:
```suggestion
            forall<k: K::DeepModelTy, v: V> self.has_mapping(k, v) ==> resolve(&v)
```

creusot/tests/should_succeed/red_black_tree.rs
```diff
@@ -178,6 +178,49 @@ impl<K: DeepModel, V> ShallowModel for Tree<K, V> {
...
+    {
+    }
+}
+
+impl<K: DeepModel, V> Resolve for Node<K, V> {
+    #[open(self)]
+    #[predicate(prophetic)]
+    fn resolve(self) -> bool {
+        pearlite! {
+            forall<k: K, v: V> self.has_mapping(k.deep_model(), v) ==> resolve(&v)
```
jhjourdan, line 206: Idem

creusot/tests/should_succeed/red_black_tree.rs
```diff
@@ -178,6 +178,49 @@ impl<K: DeepModel, V> ShallowModel for Tree<K, V> {
...
+    #[logic(prophetic)]
+    #[requires(structural_resolve(&self))]
+    #[ensures(self.resolve())]
+    fn resolve_coherence(self)
+    where
+        Self: Sized,
+    {
+        let Node { left, right, key, val, .. } = self;
+
+        proof_assert!(forall<k : K, v: _> self.has_mapping(k.deep_model(), v) ==> k.deep_model() == key.deep_model() && v == val || left.has_mapping(k.deep_model(), v) || right.has_mapping(k.deep_model(), v));
```
jhjourdan, line 220:
```suggestion
        proof_assert!(forall<k : K::DeepModelTy, v: _> self.has_mapping(k, v) ==> k == key.deep_model() && v == val || left.has_mapping(k, v) || right.has_mapping(k, v));
```

creusot/tests/should_succeed/red_black_tree.rs
```diff
@@ -178,6 +178,49 @@ impl<K: DeepModel, V> ShallowModel for Tree<K, V> {
...
+    #[requires(structural_resolve(&self))]
+    #[ensures(self.resolve())]
+    fn resolve_coherence(self)
+    where
+        Self: Sized,
+    {
+    }
+}
+
+impl<K: DeepModel, V> Resolve for Node<K, V> {
```
jhjourdan, line 201: Ideally, we would also have an instance for the only public type `Map<K, V>` at the end of the file, and it should be based on the shallow model (`has_mapping` is a private symbol).

creusot-contracts/src/resolve.rs
xldenis: Because it's not actually usable as far as I know, I don't see where you would have an unsized second component to a tuple.

jhjourdan: That's valid Rust code:

```rust
fn f (x: &(i32, [i32])) -> &[i32]{
    &x.1
}
```

xldenis: In that case both should be unsized no?

jhjourdan: Nope, only the last one is allowed to be unsized, so that the compiler can compute offset.

xldenis: huh, i don't know why I didn't that would work on tuples.

creusot-contracts/src/resolve.rs
xldenis: this would mean changing `resolve(self) -> resolve(&self)` are you alright with that?

creusot/tests/should_succeed/red_black_tree.rs
xldenis: I just took a look at this, there is a complication introduced by the lack of invariants in the logic. We would need to add `inv` as a precondition to resolve, which... seems heavy weight and annoying. Additionally I worry that we're heading to a situation in which we are back to manually adding `inv` to every trait declaration :/

jhjourdan: You don't need to add `inv` as a precondition of `resolve`, you can simply guard `resolve` by it.

I.e., `resolve(map) = inv(map) -> forall k, resolve(map@(k))`

jhjourdan: Actually, you don't even need this here, because there is only one implication of `has_mapping_model` that actually needs the invariants, and we need the other implication to prove coherence!

jhjourdan: Thinking about it, I think it would not be too bad to add `inv` as a precondition:
- It's a preconditon of `resolve_coherence`, not of `resolve`
- It is not so annoying for the user, because when instantiating, we don't need to add the preconditon if we don't need it.
- At some point, we will change the handling of default implementations so that users will not have to repeat the full specification when instantiating.

xldenis: > It's a preconditon of resolve_coherence, not of resolve

Yea that's actually a good point which alleviates things a good amount imo.

creusot-contracts/src/resolve.rs
jhjourdan: We don't really have a choice. The current situation is unsound, because this means that resolve coherence is not required for unsized types.

xldenis: Yea good point. It should also be possible to only change the receiver on the coherence law, since sidedness shouldn't be enforced in pearlite

creusot/tests/should_succeed/red_black_tree.rs
jhjourdan: Done.

jhjourdan:
Alright, we should be good now. Merging.

PR #1135, opened 2024-09-29
More stable hash

### LINKED ISSUE #1134: Identifer names are unstable
Since @lysxia has moved us to hash based identifiers, they have become unstable, locally every proof regenerates at least one identifier causing proofs to be marked as obsolete / having detached goals. I don't know _why_ this occurs though, perhaps `StableHash` is not stable enough? Or we're too aggressively caching metadata?

Lysxia:
Closes #1134

@xldenis Do you get the same hashes with this?


xldenis:
```

**Comment by xldenis** [comment]
To help narrow things down, I added the following patch:

```
  diff --git a/creusot/src/util.rs b/creusot/src/util.rs
  index c348a0790..1d5b3ae4c 100644
  --- a/creusot/src/util.rs
  +++ b/creusot/src/util.rs
  @@ -322,6 +322,12 @@ fn push_translate_name(n: &str, dest: &mut String) -> () {
       }
   }

  +fn print_hash<'tcx, T : VeryStableHash<TyCtxt<'tcx>> + ?Sized>(tcx: TyCtxt<'tcx>, x : &T) -> Hash64 {
  +    let mut hcx = StableHasher::new();
  +    x.very_stable_hash(&tcx, &mut hcx);
  +    hcx.finish::<Hash64>()
  +}
  +
   pub fn translate_name(n: &str) -> String {
       let mut dest = String::new();
       push_translate_name(n, &mut dest);
  @@ -758,8 +764,10 @@ impl VeryStableHash<TyCtxt<'_>> for GenericArg<'_> {

   impl VeryStableHash<TyCtxt<'_>> for TraitRef<'_> {
       fn very_stable_hash(&self, tcx: &TyCtxt, hcx: &mut StableHasher) {
  +        eprintln!("{:?}", print_hash(*tcx, &self.def_id));
           self.def_id.very_stable_hash(tcx, hcx);
           self.args.very_stable_hash(tcx, hcx);
  +        eprintln!("{:?}", print_hash(*tcx, &*self.args));
       }
   }
```

which generated the following output:

```
Testing tests/should_fail/bug/specialize.rs ... failure
- 278 ┊ module M_specialize__qyi2765124771875051275
+ 278 ┊ module M_specialize__qyi478073546818809368
...- 340 ┊ module M_specialize__qyi14314522871948972163
...+ 340 ┊ module M_specialize__qyi2609712117916207421
+ 27 ┊ 12847387145618251460
+ 28 ┊ 2914297198926378469
+ 29 ┊ 12847387145618251460
+ 30 ┊ 5911455361386648334
thread 'main' panicked at creusot/tests/ui.rs:220:9:
```

**Comment by Lysxia** [comment]
Thanks! It's the `Ty`

```
+ 27 ┊ 12847387145618251460
+ 28 ┊ 14274724108898276587
+ 29 ┊ 12847387145618251460
+ 30 ┊ 2367775686090831460
```

xldenis:
> I managed to get it build on windows to I can reproduce the issue.

Let me know if you need any help or support.

Lysxia:
- I had to be more careful to not call `hash()` on `Symbol` directly (which is a pointer), and instead hash the string it points to.
- Another issue is that `DefId` identifies crates by numbers (`crate0`, `crate1`, etc.), and those numbers are different between machines (idk whether that's dependent on the platform or installation order or something else). My plan is to replace those numbers with the crate's name.

Lysxia:
@xldenis You can test again now :)

xldenis:
all good!

xldenis:
@Lysxia I've rebuilt all the obsolete proofs locally, if you don't mind I'll push and merge this PR.

Lysxia:
Uh oh, I have one more change to the hashing function which means the hashes are invalidated again. I've got a fix for the proofs though.

xldenis:
> Uh oh, I have one more change to the hashing function which means the hashes are invalidated again. I've got a fix for the proofs though.

Ah ok, that's fine I'll just leave this to you then.

Lysxia:
I probably forgot to update why3, but the names should be fixed so you just need to run why3 on them.

PR #1136, opened 2024-09-30
Specify most of `Option`'s methods

arnaudgolfouse:
A lot of methods involving closures are a bit awkward to use right now, this should get better with closure automatic spec inference.


creusot-contracts/src/std/option.rs
```diff
@@ -35,63 +35,276 @@ extern_spec! {
...
+                    F: FnOnce(&T);
+
+                #[requires(match self {
+                    None => true,
+                    Some(t) => f.precondition((t,)),
+                })]
+                #[ensures(match self {
+                    None => result == default,
+                    Some(t) => f.postcondition_once((t,), result)
+                })]
```
xldenis, line 143: These kinds of specs really make me wish for a predicate-transformer style specification, or some alternate mechanism to reduce this duplication.

xldenis (approved):
The specifications _look_ good when reading them but they are quite repetitive and its always hard to be 100% certain when adding axioms like this. 

Two thoughts occur to me: 
1. It would be nice to have better syntax than `postcondition_mut` & `precondition` to describe closure calls (even if it desugars into that). 
2. To alleviate the concerns of these axioms, and potentially allowing the "inference" of specifications for these simple functions, we could have an "external bodies" mechanism in `extern_specs!` where we actually provide the body and prove the spec (if provided) holds. Creusot would then treat this as the actual body for the relevant function.

arnaudgolfouse:
Hard agree on point 2: I actually did this by hand for some of those function where the spec wasn't entirely clear (e.g. `filter` and `take_if`)

arnaudgolfouse:
Also, I could not write a spec for `unwrap_unchecked` because this is an `unsafe` function.

xldenis:
> Also, I could not write a spec for unwrap_unchecked because this is an unsafe function.

Wait why is this relevant? We can definitely write the spec, its the same as `unwrap`. Or is it a limitation in the `extern_spec!` macro?

creusot-contracts/src/std/option.rs
```diff
@@ -35,63 +35,276 @@ extern_spec! {
...
+                        (predicate.postcondition_once((&t,), true) != predicate.postcondition_once((&t,), false)),
+                })]
+                #[ensures(match self {
+                    None => result == None,
+                    Some(t) => if predicate.postcondition_once((&t,), true) {
+                        result == Some(t)
+                    } else {
+                        result == None && resolve(&t)
+                    },
+                })]
```
jhjourdan, line 214: Shorter spec, which does not requires determinism of the predicate:


```suggestion
                #[ensures(match self {
                    None => result == None,
                    Some(t) =>
                           predicate.postcondition_once((&t,), true) && result == Some(t)
                        || predicate.postcondition_once((&t,), false) && result == None && resolve(&t)
                    },
                })]
```

jhjourdan: Or even:

```suggestion
                #[ensures(match self {
                    None => result == None,
                    Some(t) => match result {
                        None => predicate.postcondition_once((&t,), false) && resolve(&t),
                        Some(r) => predicate.postcondition_once((&t,), true) && r == t
                    },
                })]
```

creusot-contracts/src/std/option.rs
```diff
@@ -35,63 +35,276 @@ extern_spec! {
...
+                #[requires(match *self {
+                    None => true,
+                    Some(t) => forall<b:&mut T> *b == t ==> predicate.precondition((b,)),
+                })]
+                #[ensures(match *self {
+                    None => result == None && ^self == None,
+                    Some(cur) =>
+                        exists<b: &mut T, res: bool> cur == *b && predicate.postcondition_once((b,), res) &&
+                            if res {
+                                ^self == None && result == Some(^b)
```
jhjourdan, line 279: BUG:

```suggestion
                                ^self == None && result == Some(*b)
```

jhjourdan: Well, no, that's good, sorry!

creusot-contracts/src/std/option.rs
```diff
@@ -35,63 +35,276 @@ extern_spec! {
...
+                where
+                    F: FnOnce() -> T;
+
                 #[pure]
                 #[ensures(result == *self && ^self == None)]
                 fn take(&mut self) -> Option<T>;
 
+                #[requires(match *self {
+                    None => true,
+                    Some(t) => forall<b:&mut T> *b == t ==> predicate.precondition((b,)),
```
jhjourdan, line 272:
```suggestion
                    Some(t) => forall<b:&mut T> inv(b) && *b == t ==> predicate.precondition((b,)),
```

arnaudgolfouse:
The `unwrap_unchecked` method is now specified, and I added an `extern_spec` body to the "difficult" functions (the ones with closures)

PR #1139, opened 2024-09-30
Add ext specs for map and filter

xldenis:
This adds the missing `extern_spec!` declarations for `map` (we only had `map_inv`) and `filter` (which is new).


jhjourdan:
There are many remaining warnings visible in CI.

creusot-contracts/src/invariant.rs
```diff
@@ -41,3 +41,8 @@ impl<T: ?Sized> Invariant for &mut T {
 pub fn inv<T: ?Sized>(_: T) -> bool {
     true
 }
+
+#[cfg(not(creusot))]
+pub fn inv<T: ?Sized>(_: &T) -> bool {
+    true
+}
```
jhjourdan, line 48:
```suggestion
pub fn inv<T: ?Sized>(_: &T) -> bool {
    panic!()
}
```

creusot/tests/should_succeed/iterators/17_filter/why3session.xml
```diff
@@ -33,6 +34,25 @@
...
+    <proof prover="1"><result status="unknown" time="0.076436" steps="16772"/></proof>
+    <proof prover="2"><result status="timeout" time="1.000000" steps="21843"/></proof>
+    <proof prover="3"><result status="unknown" time="0.033070" steps="10638"/></proof>
+    <transf name="exists" arg1="(empty : seq t_Item&#39;0)">
+     <goal name="vc_produces_refl.0.0.1.0">
+     <proof prover="0"><result status="timeout" time="1.000000" steps="2110361"/></proof>
+     <proof prover="1"><result status="unknown" time="0.074295" steps="16825"/></proof>
+     <proof prover="2"><result status="timeout" time="1.000000" steps="21208"/></proof>
+     <proof prover="3"><result status="unknown" time="0.035550" steps="10695"/></proof>
+     </goal>
```
jhjourdan, line 52: You should clean up the proofs!

creusot-contracts/src/std/iter/filter.rs
```diff
@@ -0,0 +1,129 @@
...
+    #[predicate(prophetic)]
+    fn produces(self, visited: Seq<Self::Item>, succ: Self) -> bool {
+        pearlite! {
+            self.func().unnest(succ.func()) &&
+            // f here is a mapping from indices of `visited` to those of `s`, where `s` is the whole sequence produced by the underlying iterator
+            // Interestingly, Z3 guesses `f` quite readily but gives up *totally* on `s`. However, the addition of the final assertions on the correctness of the values
+            // blocks z3's guess for `f`.
+            exists<s : Seq<Self::Item>, f : Mapping<Int, Int>> self.iter().produces(s, succ.iter()) &&
+                // `f` is a monotone mapping
+                (forall<i : _, j :_ > 0 <= i && i <= j && j < visited.len() ==> 0 <= f.get(i) && f.get(i) <= f.get(j) && f.get(j) < s.len()) &&
```
jhjourdan, line 104:
```suggestion
                (forall<i : _, j :_ > 0 <= i && i < j && j < visited.len() ==> 0 <= f.get(i) && f.get(i) < f.get(j) && f.get(j) < s.len()) &&
```

THIS is injectivity.

creusot-contracts/src/std/iter/filter.rs
```diff
@@ -0,0 +1,129 @@
...
+                // `f` is injective
+                (forall<i : _, > 0 <= i && i < visited.len() ==> visited[i] == s[f.get(i)]) &&
```
jhjourdan, line 106: That is just saying `visited` is a subsequence of s. This is not injectivity.

creusot-contracts/src/std/iter/filter.rs
```diff
@@ -0,0 +1,129 @@
...
+where
+    I: Iterator,
+    F: FnMut(&I::Item) -> bool,
+{
+    #[open]
+    #[predicate(prophetic)]
+    fn completed(&mut self) -> bool {
+        pearlite! {
+        (exists<inner : &mut _> inv(inner) && *inner == self.iter() && ^inner == (^self).iter()
+            && inner.completed()) && (*self).func() == (^self).func() }
```
jhjourdan, line 91: I don't see why this is correct. When `Filter::next` returns, the underlying iterator may have performed several steps.

xldenis:
> There are many remaining warnings visible in CI.

I know, there are still some changes i need to make locally.

creusot-contracts/src/std/iter/filter.rs
xldenis: hmm, fair but I wouldn't say its a _subsequence_ here.

creusot-contracts/src/std/iter/filter.rs
xldenis: That's not forbidden?

creusot-contracts/src/std/iter/filter.rs
jhjourdan: That's the definition of the word "subsequence" in mathematics: https://en.wikipedia.org/wiki/Subsequence

The thing you think is a subsequence is a "factor".

creusot-contracts/src/std/iter/filter.rs
xldenis: lol nope, i can prove false trivially, yet no proof fails! unsoundness time!

creusot-contracts/src/std/iter/filter.rs
xldenis: yea but a subsequence in terms of smt sequences is definitely a contiguous series of elements, that's what our `Seq::subsequence` predicate describes.

creusot-contracts/src/std/iter/filter.rs
jhjourdan: Why would it be forbidden? If `Filter` does not find an item that satisfy the predicate, then it returns `None`, even if several items have been produced by the underlying iterator.

xldenis: agreed, i meant that it wasn't supposed to be forbidden, but now i notice that I never exercised that spec due to an unsoundness in our encoding.

PR #1140, opened 2024-09-30
Major unsoundness in pattern matching encoding

xldenis:
The encoding for pattern matching is busted! it doesn't properly assert `false` when the pattern isn't matched!


creusot/src/backend/ty.rs
```diff
@@ -514,12 +514,20 @@ pub(crate) fn destructor<'tcx>(
...
 
-    let fail = Expr::Assert(Box::new(Exp::mk_false()), Box::new(Expr::Any));
+    let fail = Expr::BlackBox(Box::new(Expr::Assert(Box::new(Exp::mk_false()), Box::new(Expr::Any))));
+
+    let existential_params : Vec<_> = fields.iter().cloned().collect();
+    let negative_assertion = if existential_params.is_empty() {
+        cons_test.neq(Exp::var("input"))
+    } else {
+         Exp::Exists(existential_params, vec![], Box::new(cons_test.eq(Exp::var("input")))).not()
+    };
```
jhjourdan, line 524: Not Exists x, P  => Forall x, Not P

```suggestion
    let forall_params : Vec<_> = fields.iter().cloned().collect();
    let negative_assertion = if forall_params_params.is_empty() {
        cons_test.neq(Exp::var("input"))
    } else {
         Exp::Forall(forall_params, vec![], Box::new(cons_test.neq(Exp::var("input"))))
    };
```

Also, we could add `cons_test` as a trigger, because this is the obvious thing to do here. The SMT solvers should be able to do that by themselves, but who knows what Why3 does with quantifiers and polymorphism...

xldenis: yea talking with @paulpatault he suggested doing the exists -> forall transformation. my incoming update includes that.

creusot/src/backend/ty.rs
```diff
@@ -514,12 +514,21 @@ pub(crate) fn destructor<'tcx>(
...
+    let fail =
+        Expr::BlackBox(Box::new(Expr::Assert(Box::new(Exp::mk_false()), Box::new(Expr::Any))));
+
+    let fields: Vec<_> = fields.iter().cloned().collect();
+    let negative_assertion = if fields.is_empty() {
+        cons_test.neq(Exp::var("input"))
+    } else {
+        Exp::Forall(fields, vec![], Box::new(cons_test.neq(Exp::var("input"))))
+    };
+
```
xldenis, line 526: @paulpatault can you confirm that the semantics of this encoding are correct? we want to match the designated constructor and reject other cases.

creusot/tests/should_succeed/bdd/why3session.xml
```diff
@@ -40,12 +40,35 @@
   <goal name="vc_hash.2" expl="postcondition" proved="true">
   <proof prover="6"><result status="valid" time="0.606420" steps="47937"/></proof>
   </goal>
+  <goal name="vc_hash.3" proved="true">
+  <transf name="destruct_term_subst" proved="true" arg1="childf">
```
xldenis, line 44: annoyingly this proof requires a manual destruct, I suspect triggers / data type encoding are to blame.

creusot/src/backend/program.rs
```diff
@@ -1081,23 +1087,29 @@ impl<'tcx> Statement<'tcx> {
...
-                    exp = Exp::Match(
+                let bound = lower.fresh_sym_from("inv_asrt");
+
+                let pat = pattern_of_place(lower.ctx.tcx, lower.locals, pl, bound);
+
+                let pat = lower_pat(lower.ctx, lower.names, &pat);
+                let exp = if let Pattern::VarP(_) = pat {
+                    rp.app_to(Exp::var(util::ident_of(loc)))
+                } else {
+                    Exp::Match(
```
xldenis, line 1105: I kept the optimization present before which avoided introducing a match on a wildcard (seems reasonable).

creusot/src/translation/pearlite.rs
```diff
@@ -250,13 +250,30 @@ pub enum Literal<'tcx> {
...
+    Constructor {
+        variant: DefId,
+        substs: GenericArgsRef<'tcx>,
+        fields: Vec<Pattern<'tcx>>,
+    },
+    /// Matches the pointed element of a pointer, so for `Box<T>` it matches `T`, for mutable borrows it matches the *current* value
+    Deref {
+        pointee: Box<Pattern<'tcx>>,
+        kind: PointerKind,
+    },
```
xldenis, line 262: This doesn't allow us to handle binding modes but it does give us potentially a little more flexibility in what we can do in source pearlite!

creusot/tests/should_succeed/iterators/06_map_precond.coma
```diff
@@ -40,9 +37,7 @@ module T_alloc__alloc__Global
   
   function any_l (_ : 'b) : 'a
   
-  let rec t_Global (input:t_Global) (ret  )= any
-    [ good -> {C_Global  = input} (! ret) | bad -> {C_Global  <> input} {false} any ]
-    
+  let rec t_Global (input:t_Global) (ret  )= any [ good -> {C_Global  = input} (! ret) ]
```
xldenis, line 40: @paulpatault additionally, if you can confirm its safe to remove the `bad` branch for product typesl ike this it would be useful.

creusot/tests/should_succeed/hashmap.coma
```diff
@@ -933,7 +921,11 @@ module M_hashmap__qyi7664122466964245986__add
     | bb15 = s0 [ s0 =  [ &_47 <- List'0.C_Cons _48 _52 ] s1 | s1 = bb16 ] 
     | bb16 = bb17
     | bb17 = s0
-      [ s0 = {[@expl:type invariant] inv'1 l.current} s1
+      [ s0 = {[@expl:type invariant] match l with
+          | {current = inv_asrt'0} -> inv'1 inv_asrt'0
+          | _ -> true
+          end}
```
xldenis, line 927: the matches generated here are worse than the prior code, but fixing this would be a lot more sophisticated and I would rather defer that.

xldenis:
Ok, I tried giving this a read over, but I'm really tired so I wasn't able to _read_ it. However, conceptually this fixes our soundness holes. Along the way I added a minor optimization to avoid generating "bad" branches in destructors when we only have on constructor in the type. It seems to have a surprising impact, it turns out that proving that proving that there isn't more than one constructor for those types is really challenging for the smt (something something polymorphism encoding). That along makes the soundness fix effectively invisible in the proofs.

creusot/src/backend/program.rs
```diff
@@ -1081,23 +1087,29 @@ impl<'tcx> Statement<'tcx> {
...
                 lower.ctx.translate(did);
                 let mut istmts = Vec::new();
 
                 let rp = Exp::qvar(lower.names.value(did, subst));
                 let loc = pl.local;
 
-                let mut exp = rp.app_to(pl.as_rplace(lower, &mut istmts));
-                if let Some(pat) = pat {
-                    exp = Exp::Match(
+                let bound = lower.fresh_sym_from("inv_asrt");
```
jhjourdan, line 1097: "inv_asrt" does not make any sense as an identifier. Even somthing like "x" would be better.

creusot/src/backend/program.rs
```diff
@@ -1121,22 +1133,28 @@ impl<'tcx> Statement<'tcx> {
...
+            Statement::AssertTyInv { pl } => {
                 let mut istmts = Vec::new();
 
                 let inv_fun = Exp::qvar(lower.names.ty_inv(pl.ty(lower.ctx.tcx, lower.locals)));
                 let loc = pl.local;
 
-                let mut exp = inv_fun.app_to(pl.as_rplace(lower, &mut istmts));
-                if let Some(pat) = pat {
-                    exp = Exp::Match(
+                let bound = lower.fresh_sym_from("inv_asrt");
```
jhjourdan, line 1142: Same here. The thing which is bound is *not* the invariant assertion.

creusot/tests/should_fail/bug/222.coma
```diff
@@ -40,7 +40,7 @@ module T_222__Once
   
   let rec t_Once < 't > (input:t_Once 't) (ret  (field_0:Option'0.t_Option 't))= any
     [ good (field_0:Option'0.t_Option 't)-> {C_Once field_0 = input} (! ret {field_0})
-    | bad (field_0:Option'0.t_Option 't)-> {C_Once field_0 <> input} {false} any ]
+    | bad -> {forall field_0 : Option'0.t_Option 't . C_Once field_0 <> input} (! {false} any) ]
```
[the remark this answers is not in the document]
paulpatault: This one is good.

creusot/tests/should_succeed/iterators/06_map_precond.coma
paulpatault: I don't see where is the product type, there is only one constructor in `type t_Global  = | C_Global` or I miss something ?

creusot/tests/should_fail/bug/222.coma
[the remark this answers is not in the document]
paulpatault: Notice that you may want to write 
```
let t_Once < 't >
           (input: t_Once 't)
           { exists x : Option'0.t_Option 't . C_Once x = input }
           (ret  (x: Option'0.t_Option 't) {C_Once x = input})
  = any
```
which is more readable in my opinion.

PR #1141, opened 2024-10-02
Add impl provenance to coma output

### LINKED ISSUE #1137: Add provenance information for impl hashes
Impls are now non-ambiguously encoded as hashes in coma identifiers, but then it's not really easy for an external tool or for humans to guess the origin of a given coma module, or conversely the translation of a given rust function, unless you reuse or replicate the hashing function that produces the hashes.

I'd like to make this information more easily accessible by showing the impl type in the output next to each module that contains a hashed impl.

```
impl MyType {
   fn f() { ... }
}
```

would generate

```
module M_mylib__qyi1234567890__f (* mylib::MyType *)
```

I've got a branch that does this, but before making a PR I'm wondering if there's a better solution, or any details to bikeshed.

For one thing, my implementation outputs the qualified types because that's what I have access to. It may be easier to read the types as written in the source instead, is there an easy way to get that from the rustc API?

Lysxia:
Closes #1137

### DIFF
--- a/creusot/tests/should_fail/bug/222.coma
+++ b/creusot/tests/should_fail/bug/222.coma
@@ -1,5 +1,5 @@
 
-module M_222__A__is_true__Impl
+module M_222__A__is_true__Impl [#"../222.rs" 14 4 14 16]
   type self
   
   let%span s2220 = "../222.rs" 13 14 13 34
@@ -30,7 +30,7 @@ module T_core__option__Option
     | bad -> {forall field_0 : 't [C_Some field_0 : t_Option 't] . C_Some field_0 <> input} (! {false} any) ]
     
 end
-module T_222__Once
+module T_222__Once [#"../222.rs" 34 0 34 18]
   use T_core__option__Option as Option'0
   
   type t_Once 't =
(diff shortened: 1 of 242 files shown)


why3/src/declaration.rs
```diff
@@ -16,6 +16,10 @@ use serde::{Deserialize, Serialize};
 pub struct Module {
     pub name: Ident,
     pub decls: Vec<Decl>,
+    pub attrs: Vec<Attribute>,
+    // Meta data stored in comments
+    // Stores a pretty representation of the impl
+    pub meta: Option<String>,
```
xldenis, line 22: If we start using this more often we may just want to add `Comment` nodes to our ast.

creusot/tests/should_fail/bug/222.coma
```diff
@@ -154,9 +154,9 @@ module M_222__uses_invariant
     | & _5 : borrowed (Option'0.t_Option t) = any_l () ]
      [ return' (result:())-> {[@expl:postcondition] [%#s2222] invariant'0 x.final} (! return' {result}) ] 
 end
-module M_222__qyi2574917115831112192
+module M_222__qyi2574917115831112192 [#"../222.rs" 19 0 19 13] (* <() as A> *)
```
xldenis, line 157: I wonder if Why3 has a feature that lets us remap the names in the ID, it would definitely be preferable to see "<() as A>" to `M_222__qyi2574917115831112192`

xldenis (approved):
lgtm. one thing that could be interesting in future work is to see if we can remap the name why3 displays in the ide, (I kind of doubt it).

PR #1150, opened 2024-10-11
Simplify test case 06_knights_tour

### LINKED ISSUE #1148: Simplifying 06_knights_tour
I'm looking at the [test cases that don't pass why3find](https://github.com/creusot-rs/creusot/issues/1002#issuecomment-2402610030) one by one to figure out what is missing.

The test `vector/06_knights_tour.rs` uses the transformation `use_th seq.FreeMonoid`, which imports a theory, like writing `use seq.FreeMonoid` in WhyML/Coma except that `use_th` is scoped to a specific goal, so it doesn't introduce unnecessary assumptions in other goals. For this test case, adding `use seq.FreeMonoid` before the definitions of `count_degree` and `knights_tour` makes `why3find prove` pass without any significant difference in other goals (all goals in the changed modules pass in < 1s). So the test case doesn't depend on the more fine-grained behavior of `use_th`.

1. Can we tell Creusot to insert `use` declarations from our contract annotations?
2. Do we need the more fine-grained behavior of `use_th` elsewhere? It seems that it should always be avoidable by introducing lemmas that are free to use `use` in their own proofs.

Lysxia:
Closes #1148. Contrary to what I naively thought, we can do this without `use`.

For the loop variable `m` which comes out of the iterable `result = moves()`, we have the following relation:

```
exists<a, b> result@ = a ++ singleton m@ ++ b
```

whereas the postcondition of `moves` gives a specification of `result` using indexing notation:

```
forall<i> 0 <= i < 8 ==> -2 <= result[i].0@ && ...
```

Solvers need a bit of help to connect those formulations. They need to see `m` as something of the form `result[i]` for some `i`.

This trick may be packaged better if it's going to be a common pattern. Maybe there is a way to improve the iterator specs, but I'm not familiar with those details yet.

---

From my experiments the lemma from `seq.FreeMonoid` that helped was

```
snoc_def : forall a, x. snoc a x = a ++ singleton x
```

although it doesn't work to just insert it in place of the assertion that's in this PR.


xldenis (approved):
It's good to remove usages of use_th, but it would be nice to find a proper solution to array reasoning issues.

PR #1161, opened 2024-10-16
Fix #1157

### LINKED ISSUE #1157: stop generating empty spans in .coma
```
# find creusot/tests -name \*.coma -exec grep -E -e 'let%span [a-z0-9]* = ""' {} \;
...
  let%span span41 = "" 0 0 0 0
  let%span span43 = "" 0 0 0 0
...
```

why3 and why3ide complain on those spans and then emit a lot of warnings, which are hard to then distinguish from actually incorrect spans.

xldenis:
r @armael

### DIFF
--- a/creusot/src/backend.rs
+++ b/creusot/src/backend.rs
@@ -211,7 +211,7 @@ impl<'tcx> Why3Generator<'tcx> {
                 debug!("translating {:?} as logical", def_id);
 
                 if util::is_resolve_function(self.tcx, def_id) {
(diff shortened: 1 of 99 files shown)


xldenis:
> This seems unavoidable for the testsuite .coma files, but it makes me wonder if we should make "absolute spans" the default mode for users?

Yeah, I agree. I think `cargo creusot` should be absolute by default (is it not already?). the relative spans are only useful in the test suite, until such a date as we can safely afford to avoid comitting those files.

Armael:
> I think cargo creusot should be absolute by default (is it not already?).

It's not. I'll make a PR!

PR #1167, opened 2024-10-18
Upgrade to alt-ergo 2.6.0; now as a downloadable binary

Armael:
With this change, `cargo creusot setup` now downloads binaries for alt-ergo.


xldenis (approved):
Should this be mentioned in the changelog? What's left to do the same with why3?

Armael:
wrt the changelog, it isn't really a user visible change so idk.
for why3 binaries, the hard part is building those binaries; I have a semi worked out plan, but I still need to execute it...

jhjourdan:
I think this is definitely a user visible change (it can break proofs), so it should be written to the changelog.

PR #1169, opened 2024-10-18
Specs for raw pointers & ghost ownership for pointers

Armael:
I'm extracting extern specs for basic raw pointer operations (mainly getting their address and comparing with null) from `ghost_ptr.rs` because I need them in another context. (cc @dewert99)

In the process I went and skimmed the docs (https://doc.rust-lang.org/std/ptr/index.html and https://doc.rust-lang.org/std/primitive.pointer.html) and I ended up with a more conservative model of raw pointers than `ghost_ptr.rs`, where pointers always have some hidden "metadata" associated, and one basically can never logically deduce equality of pointers except by calling `==`.

I have two questions:
- does this seem too conservative (ie not expressive enough)? Glancing at the docs I could tell if there could be a more precise yet reasonable model.
- the specs currently do not say anything about `==` (there is no `DeepModel` instance in particular). I would like to say that the DeepModel of a `*mut T` is "itself" (and we know nothing more), and that if `p == q` returns `true` then we know that `p = q` in the logic and also `p.addr_logic() = q.addr_logic()`. What do I need to write to express that in Creusot?

If the current approach seems OK then I'll have a go at updating `ghost_ptr.rs` to use the new model.


xldenis:
> . I would like to say that the DeepModel of a *mut T is "itself" (and we know nothing more), and that if p == q returns true then we know that p = q in the logic and also p.addr_logic() = q.addr_logic(). What do I need to write to express that in Creusot?

This sounds like a spec on `==` no? As it is, the specification of `==` says that `result ==> p.deep_model() == q.deep_model()`, you can then also add that the addresses must be the same. 

Is it true that `p.addr_logic() == q.addr_logic() ==> p = q`?

Armael:
> Is it true that `p.addr_logic() == q.addr_logic() ==> p = q`?

no! that's precisely what we want to avoid

xldenis:
> no! that's precisely what we want to avoid

Ah right, `addr_logic` doesn't include provenance.

jhjourdan:
>  I would like to say that the DeepModel of a *mut T is "itself" 

But then that would mean that program equality implies equality of provenance, which is wrong?

Armael:
Indeed... I'll push an updated version later.

creusot-contracts/src/std.rs
```diff
@@ -13,6 +13,7 @@ pub mod num;
 pub mod ops;
 pub mod option;
 pub mod panicking;
+pub mod pointer;
```
jhjourdan, line 16:
```suggestion
pub mod ptr;
```

In the `std` module, we try to follow Rust's stdlib structure.

jhjourdan: Also, we re-export the corresponding std module in each of these, so that the user does not have to import both (I know this has not been done everywhere, but this should be fixed.).

creusot-contracts/src/std/pointer.rs
```diff
@@ -0,0 +1,92 @@
...
+//
+// This is to account for provenance
+// (https://doc.rust-lang.org/std/ptr/index.html#using-strict-provenance) and
+// wide pointers. See e.g.
+// https://doc.rust-lang.org/std/primitive.pointer.html#method.is_null: "unsized
+// types have many possible null pointers, as only the raw data pointer is
+// considered, not their length, vtable, etc. Therefore, two pointers that are
+// null may still not compare equal to each other."
+
+pub trait PointerExt<T: ?Sized>: Sized {
```
jhjourdan, line 14: This trait should be exported in `base_prelude` so that it can be used without importing std::pointer.

creusot-contracts/src/std/pointer.rs
```diff
@@ -0,0 +1,92 @@
...
+        #[ensures(result@ == self.addr_logic())]
+        fn addr(self) -> usize;
+
+        #[ensures(result == (self.addr_logic() == 0))]
+        fn is_null(self) -> bool;
+    }
+
+    mod std {
+        mod ptr {
+            #[ensures(result.addr_logic() == 0)]
```
jhjourdan, line 76:
```suggestion
            #[ensures(result == Self::null_logic())]
```

Otherwise, `null_logic` should be removed.

creusot-contracts/src/std/pointer.rs
```diff
@@ -0,0 +1,92 @@
...
+            fn null<T>() -> *const T
+            where
+                T: std::ptr::Thin + ?Sized;
+
+            #[ensures(result.addr_logic() == 0)]
```
jhjourdan, line 81: Idem.

jhjourdan (approved):
As for the `DeepModel` impl for pointers, I don't know what to say: metadata should be taken into account, but we cannot say that the pointer is only the address and the metadata, because of provenance.

We could have an instance for `Sized` types only?

@xldenis : is it possible to have two pointer of equal address and underlying types, but without the same metadata.

creusot-contracts/src/std/pointer.rs
Armael: this is not sound wrt the existence of multiple nulls. We could remove `Self::null_logic()` if it's too tempting to incorrectly use to test nullity (instead of `.is_null_logic()`). In principle, `null_logic()` gives you the ability to talk about "a null pointer" in the logic, but I don't know how useful that is...

jhjourdan: What is not sound is to assume that all the null types are equal. This is not what I am proposing here.

What I am proposing is essentially to assume that the `null` function is deterministic (it always returns the same value). I am not completely sure, but this seems very reasonable.

Now, you may argue that this is useless, which I may agree on. But if we do not add this post-condition, then we should remove `null_logic`, because I think this is useless too.

arnaudgolfouse:
> @xldenis : is it possible to have two pointer of equal address and underlying types, but without the same metadata.

I think so: 
```rust
let s1: &[i32] = &[1];
let s2: &[i32] = &s1[..0];
let p1: *const [i32] = s1;
let p2: *const [i32] = s2;
```
p1 and p2 are the same pointer, and have the same type, but their metadata is different (1 vs 0).

creusot-contracts/src/std/pointer.rs
Armael: Indeed, I agree. To sum things up, there are two choices:
- keep `Self::null_logic()` and specify that it is equal to the result of the `null()` program function (thus specifying that `null()` is deterministic;
- remove `Self::null_logic` and only use `is_null()` / compare `addr_logic()` to 0. 

I'm leaning towards option 2, because having a logical NULL is not so useful and gives the wrong idea (that there is only one NULL pointer). Using `is_null` instead of comparing to NULL is the better option.

jhjourdan: Makes sense, and this does not prevent us from adding this in the future.

creusot-contracts/src/std.rs
Armael: done, thanks

creusot-contracts/src/std/pointer.rs
Armael: done

creusot-contracts/src/std/pointer.rs
Armael: I pushed changes implementing Option 2, with an additional helper function `is_null_logic()` which simply unfolds to `.addr_logic() == 0`. (I found this convenient in practice because it mirrors `is_null()` in program code.)

Armael:
Soundly updating the specifications in `ghost_ptr.rs` without reintroducing `addr = addr ==> ptr = ptr` doesn't seem so obvious...
One way would be to first introduce a new module for a single pointer with ghost ownership, and use it to implement `ghost_ptr.rs` (which defines a *map* -- or region -- of pointers).

Armael:
Ah, and note (in the current version of the PR) how I've tried to handle `DeepModel`: by choosing a `DeepModelTy` composed of the address and a private "runtime metadata" about which we know nothing. (Conceptually, a pointer would then be composed of three things: its address, runtime metadata, and other metadata not tested by the equality (e.g. provenance))

creusot-contracts/src/ptr_own.rs
```diff
@@ -0,0 +1,107 @@
...
+pub fn as_ref<T: ?Sized>(ptr: RawPtr<T>, own: GhostBox<&PtrOwn<T>>) -> &T {
+    unsafe { &*ptr }
+}
+
+/// Mutably borrows the underlying `T`.
+#[trusted]
+#[allow(unused_variables)]
+#[requires(ptr == own.ptr())]
+#[ensures(*result == *own.val())]
+// TODO: why is .inner_logic() needed instead of *?
```
Armael, line 73: @arnaudgolfouse do you know what's happening here? (below)

creusot/tests/should_succeed/linked_list.rs
```diff
@@ -0,0 +1,109 @@
...
+    #[logic]
+    #[open(self)]
+    fn view(self) -> Self::ViewTy {
+        pearlite! {
+            seq_map(*self.seq, |ptr_own: PtrOwn<Cell<T>>| ptr_own.val().v)
+        }
+    }
+}
+
+// TODO: allow "- 1" on Ints in ghost code directly
```
Armael, line 53: maybe we should have an issue to track this TODO…

arnaudgolfouse: I opened #1251 to track this.

creusot-contracts/src/ptr_own.rs
arnaudgolfouse: Yeah, I give more details in #1257 ... The bottom line is that using `*` was a hack anyways, and it does not work in this case 😢

creusot-contracts/src/ptr_own.rs
```diff
@@ -0,0 +1,107 @@
...
+#[ensures(result.1.ptr() == result.0 && *result.1.val() == v)]
+pub fn new<T>(v: T) -> (RawPtr<T>, GhostBox<PtrOwn<T>>) {
+    from_box(Box::new(v))
+}
+
+/// Creates a ghost [PtrOwn] and associated [RawPtr] from an existing [Box].
+#[trusted]
+#[ensures(result.1.ptr() == result.0 && *result.1.val() == *val)]
+pub fn from_box<T: ?Sized>(val: Box<T>) -> (RawPtr<T>, GhostBox<PtrOwn<T>>) {
+    assert!(core::mem::size_of_val::<T>(&*val) > 0, "PtrOwn doesn't support ZSTs");
```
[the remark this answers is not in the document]
Armael: That's a good question, I don't know; I took the assert from `ghost_ptr` (cc @dewert99 ?).
Could there be issues with using `Box::{into,from}_raw` on ZSTs maybe?

Armael:
Thanks for the review!
I've addressed the easy changes, and I've moved the functions in ptr_own to be associated functions of `PtrOwn` as discussed with @arnaudgolfouse .

I'm not sure what to do regarding the ZST-related assert, but I'd rather merge this now and remove the assert later if it is needed and sound.

creusot-contracts/src/std/ptr.rs
```diff
@@ -0,0 +1,137 @@
...
+    #[trusted]
+    #[logic]
+    #[open(self)]
+    fn addr_logic(self) -> Int {
+        absurd
+    }
+
+    #[logic]
+    #[open]
+    #[ensures(result == (self.addr_logic() == 0))]
```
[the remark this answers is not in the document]
Armael: yeah, it is in fact required because it is in the trait declaration, so we need to restate it to prove refinement

jhjourdan:
> I'm not sure what to do regarding the ZST-related assert, but I'd rather merge this now and remove the assert later if it is needed and sound.

It is indeed needed (see above), and the current situation is "unsound" (Creusot is supposed to prevent panics). Butfixing this would require reasonning about the size of a value pointed to by a `Box`, which we don't do currently. (That's probably not very difficult to add, but this requires some design decisions, etc...).

creusot-contracts/src/ptr_own.rs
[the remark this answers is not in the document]
Armael: could it be achieved using a `const` failure (instead of a dynamic one)?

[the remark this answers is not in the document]
jhjourdan: If `T` is not Sized, then this is dynamic information only.

PR #1170, opened 2024-10-18
Improve `GhostBox`'s API

arnaudgolfouse:
This is extracted from the ghost structures PR, with some changes


arnaudgolfouse:
@Armael what do you think?

PR #1172, opened 2024-10-18
Add option to generate one file for each module (enabled by default)

Lysxia:
Module `M_krate__mod__f` becomes file `krate/mod/f.coma` containing a single module `M`. We keep the single-file option for the test suite.

This avoids producing a single massive file, which slows down Why3 and the Creusot language server.

An extra empty line at the beginning of generated coma files was also removed.


creusot-rustc/src/options.rs
```diff
@@ -29,33 +29,42 @@ impl CreusotArgsExt for CreusotArgs {
...
+            (false, Some(f), None) => Ok(Output::File(f)),
+            (false, None, Some(d)) => Ok(Output::Directory(d)),
+            (true, Some(_), _) => Err("--stdout and --output-file are mutually exclusive"),
+            (true, None, Some(_)) => Err("--stdout and --output-dir are mutually exclusive"),
+            (false, Some(_), Some(_)) => {
+                Err("--output-file and --output-dir are mutually exclusive")
+            }
+            (false, None, None) => {
+                Err("please specify either --output-dir, --output-file or --stdout")
+            }
```
xldenis, line 44: Ah i see we were already doing it by hand.

creusot-args/src/options.rs
```diff
@@ -32,7 +32,9 @@ pub struct CommonOptions {
     pub stdout: bool,
     /// Print to a file.
     #[clap(group = "output", long, env)]
-    pub output_file: Option<String>,
+    pub output_file: Option<PathBuf>,
+    #[clap(group = "output", long, env)]
+    pub output_dir: Option<PathBuf>,
```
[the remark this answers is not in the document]
Lysxia: I can try to make that work

Lysxia:
This should reduce time to IDE open (for example red-black-tree takes 10s because it has to replay all the transformations). I can try some simple experiments, after fixing some remaining issues (some `use` are broken).

Creusot IDE needs some adjustment.

Armael:
I haven't looked at the code yet, but what's the end-user UI like for `cargo creusot`? Currently there is only one .coma file so `cargo creusot why3 ide` opens it; now that there are several, is there a way to list them and open the ide on a specific one?

I imagine this also makes it harder (for the moment) to check that all proofs pass across all files, until we add why3find to the commandline? I think that's probably fine...

xldenis:
> I haven't looked at the code yet, but what's the end-user UI like for cargo creusot? Currently there is only one .coma file so cargo creusot why3 ide opens it; now that there are several, is there a way to list them and open the ide on a specific one?

Indeed I expect it won't work super well with `cargo creusot why3 ide`, though we can add multiple coma files to the same session (this would be the same thing as what we actually do). I still support this feature in the context of projects using `cargo creusot` but not `cargo creusot why3 ide`. I think afterwards we need to take a step back and as you point out consider the overal UX and expected workflow both for "success" and "failure" cases.

Lysxia:
Right now `cargo-creusot` runs `why3 ide [OPTIONS] target/creusot/` which starts as an empty session to which you can add files manually. We could also run `why3 ide [OPTIONS] target/creusot/ target/creusot/**/*.coma` (first the root directory (where the session is stored), then the files to load) which would result in an equivalent UX to what we have now (including the slowness if there is a lot of code).

We can of course load individual files, but doing so from the command line is a bit tedious. The hope is to do it via the IDE. Next to each function there would be a command to run `why3 ide` just for that function.

Armael:
OK! In the meantime, if the current (clunky but functional) workflow where `cargo creusot why3 ide` still works to open everything, then I'm happy.

Lysxia:
I added the glob so `why3 ide` doesn't start with an empty session.

The argument parsing is indeed getting messy, I'll try to make it better in another PR.

xldenis:
> The argument parsing is indeed getting messy, I'll try to make it better in another PR.

sgtm

go ahead and ship this, its already incredibly useful as is.

xldenis:
Oh god, I just realized that by merging #1176, you'll have some awful conflicts to resolve. let me know if I can help.

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -343,13 +343,16 @@ impl DepElab for TyElab {
...
+                        ctx.module_path_with_suffix(def_id, "_Type")
                     } else {
-                        module_name(ctx.tcx, def_id)
+                        ctx.module_path(def_id)
                     };
                     let name = if name.module.is_empty() { name } else { name.module_qname() };
-                    let use_decl =
-                        Use { as_: Some(name.clone()), name: modl.as_str().into(), export: false };
+                    let use_decl = Use {
+                        as_: Some(Ident::from_string(name.to_string())),
```
xldenis, line 352: I think here you can use `as_ident().unwrap()`

xldenis (approved):

Lysxia:
> Oh god, I just realized that by merging #1176, you'll have some awful conflicts to resolve. let me know if I can help.

Np I resolved it!

Armael:
There seem to be some remaining issues.
Repro: https://github.com/Armael/creusot-playground (latest commit, `904e734405`); build with github.com/Armael/creusot/tree/contracts-tweaks. 
Then, launching `cargo creusot why3 ide` yields:
```
File "creusot_playground/list2_cleaned/qyi1331831093243978089/new.coma", line 177, characters 194-245: 
unbound function or predicate symbol 'T_creusot_playground__list2_cleaned__Cell.t_Cell__v'
File "creusot_playground/list2_cleaned/qyi1331831093243978089/push_front.coma", line 300, characters 195-246: 
unbound function or predicate symbol 'T_creusot_playground__list2_cleaned__Cell.t_Cell__v'
File "creusot_playground/list2_cleaned/qyi1331831093243978089/push_back.coma", line 401, characters 195-246: 
unbound function or predicate symbol 'T_creusot_playground__list2_cleaned__Cell.t_Cell__v'
```

Lysxia:
I'm looking into it.

Armael:
A few additional comments:
- first of all I very much like the more hierarchical view and the smaller .coma files!
- it seems we have lost in readability of names: before I would have a goal tree named (approximately) `myfunction` for a rust function `myfunction`. Now I have a mangled .coma name and proof tree node, and `myfunction` doesn't seem to appear in the goal tree anymore. This makes it quite hard to identify what goals correspond to for functions (and types?).
- opening all those .coma files at once overloads the tab bar of why3ide ; unfortunately this bar apparently doesn't support horizontal scrolling, this makes it hard to navigate between the coma files and I think it makes it impossible to switch to the related `.rs` files. 
- these `M` modules look a bit strange, could they have a better name or their contents be included at the toplevel?

Lysxia:
Thanks for testing it!

>     * it seems we have lost in readability of names: before I would have a goal tree named (approximately) `myfunction` for a rust function `myfunction`. Now I have a mangled .coma name and proof tree node, and `myfunction` doesn't seem to appear in the goal tree anymore. This makes it quite hard to identify what goals correspond to for functions (and types?).

I don't quite see what you mean. Doesn't a Rust function `f` give you a file `.../f.coma`?

>     * opening all those .coma files at once overloads the tab bar of why3ide ; unfortunately this bar apparently doesn't support horizontal scrolling, this makes it hard to navigate between the coma files and I think it makes it impossible to switch to the related `.rs` files.

I could look into adding support for many tabs to Why3 IDE. One workaround for now is that you can click on a tab to focus it and then use the left/right keys to switch to other tabs.

>     * these `M` modules look a bit strange, could they have a better name or their contents be included at the toplevel?

I wasn't completely sure of the syntax but it seems that they can be omitted indeed. (It looks like Why3 just wraps it in a `Coma` module, so that doesn't make Why3 IDE's tree listing shorter.)

Armael:
> I don't quite see what you mean. Doesn't a Rust function f give you a file .../f.coma?

Apparently not, I only get the mangled identifier part of the name:
why3ide
(here, looking at the contents of the `.coma` file, I guess this file is about `push_back`)

Lysxia:
Re: the compilation error. The cause is that there are two crates in that package, the library and the executable, and creusot compiles them in order, and when compiling the executable it overwrites the files from compiling the library. I will tell creusot to use a different path when compiling executables.

Armael:
Ah, then this is probably the same issue behind #1178 ! It could be interesting to see if the issue could be also fixed for the single-coma mode at the same time.

Lysxia:
The mangled files come from the `ghost!` blocks which are closures, it seems. Why3 IDE only shows the file's basename, which is useless for closures. The full file paths are more informative:

```
creusot_playground/
  list2_cleaned/
    qyi133.../
      push_back.coma
      push_back/
        qy123zclosureqy35z1qy125z_Type.coma
        qy123zclosureqy35z1qy125z.coma
```

Maybe Why3 IDE should display the directory tree too.

PR #1173, opened 2024-10-18
Improvements for `GhostBox`

arnaudgolfouse:


arnaudgolfouse:
@xldenis I added a hack to dereference a `GhostBox` in pearlite, does it look reasonable?

xldenis:
Yea, its definitely a hack but far from the worst we've had. Do you expect to add other similar cases?

arnaudgolfouse:
Not really. If more cases are ever needed I'd rather find a more general solution.

arnaudgolfouse:
If the question is "will you add this for `Seq` and `Map`" the answer is not 🙂

xldenis (approved):

PR #1174, opened 2024-10-18
Support type annotations in pearlite

### LINKED ISSUE #1171: Unsupported pattern kind AscribeUserType
```
thread 'rustc' panicked at creusot/src/translation/pearlite.rs:839:23:
not yet implemented: lower_pattern: unsupported pattern kind AscribeUserType { ascription: Ascription { annotation: CanonicalUserTypeAnnotation { user_ty: Canonical { value: Ty(list2::Cell<T/#0>), max_universe: U0, defining_opaque_types: [], variables: [] }, span: src/list2.rs:41:25: 41:32 (#0), inferred_ty: list2::Cell<T/#0> }, variance: + }, subpattern: Pat { ty: list2::Cell<T/#0>, span: src/list2.rs:41:17: 41:23 (#0), kind: Binding { name: "cell_i", mode: BindingMode(No, Not), var: LocalVarId(HirId(DefId(0:74 ~ creusot_playground[c63a]::list2::{impl#0}::invariant).205)), ty: list2::Cell<T/#0>, subpattern: None, is_primary: true } } }

error: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/creusot-rs/creusot/issues/new
```

:-)

arnaudgolfouse:
Fixes #1171.

This simply ignores type annotations, ideally we should try to support them somehow.

### DIFF
--- a/creusot/src/translation/pearlite.rs
+++ b/creusot/src/translation/pearlite.rs
@@ -858,6 +858,10 @@ impl<'a, 'tcx> ThirTerm<'a, 'tcx> {
                 }
                 Ok(Pattern::Boolean(value.try_to_bool().unwrap()))
             }
+            // TODO: this simply ignores type annotations, maybe we should actually support them
+            PatKind::AscribeUserType { ascription: _, subpattern } => {
+                self.pattern_term(subpattern, mut_allowed)
+            }
             ref pk => todo!("lower_pattern: unsupported pattern kind {:?}", pk),
         }
     }


xldenis (approved):
lgtm. I think in practice rust inference is so bad that we shouldn't need annotations on the coma / why3 code.

jhjourdan:
I think this is fine to do so.

I don't know how complete is Why3 type inference, but if it is not, we should rather annotate *all* binders, because we can never assume Why3 type inference will behave the same as Rust's (i.e., nobody can claim that Why3 can only fail if Rust does).

PR #1175, opened 2024-10-20
Do not generate refinement obligations on trusted trait impls

arnaudgolfouse:
This makes some amount of sense, but the motivation is not super strong.
AFAIK there is only one place where this is useful right now: `Snapshot::deref`.

### DIFF
--- a/creusot/src/backend/traits.rs
+++ b/creusot/src/backend/traits.rs
@@ -4,8 +4,7 @@ use super::{
     GraphDepth, Why3Generator,
 };
 use crate::{
-    backend,
-    backend::{all_generic_decls_for, own_generic_decls_for, Namer},
+    backend::{self, all_generic_decls_for, own_generic_decls_for, Namer},
     ctx::ItemType,
     util::{self, item_name, module_name},
 };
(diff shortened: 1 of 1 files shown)


xldenis:
> This makes some amount of sense, but the motivation is not super strong.

Reading your logic, it seems like you're suggesting that marking individual items as `#[trusted]` should disable the refinement? I definitely don't agree with that, more generally I don't think we should provide any way to opt out of that, otherwise we are going to be unsound.

arnaudgolfouse:
Hm ok, I can accept that. The thing that is bothering me is that this means we cannot prove `Snapshot::deref`, even though it is `#[trusted]` 😥

xldenis:
> Hm ok, I can accept that. The thing that is bothering me is that this means we cannot prove Snapshot::deref, even though it is #[trusted]

I _could_ imagine `#[trusted]` around the impl potentially having this effect, what do you think @jhjourdan?

Armael:
What about a different (new?) attribute? I don't particularly want to encourage the proliferation of arcane attributes, but there seem to be two orthogonal usecases here: trusting that the implementation satisfies the specification (the current effect of `#[trusted]` AFAIU), and trusting that the specification of the impl refines the specification of the trait (what you need here).
It seems wrong to conflate the two, I can imagine that one may want one or the other independently.

jhjourdan:
`Deref` is already special cased in many ways, so I would not be shocked if we decided not to derive a refinement VC for it. After all, the impl spec and the trait spec have two different meanings: the first is a lemma function about the deref logic function, while the other is pre/post-cond of a program function.

arnaudgolfouse:
Alright, I just special-cased `Snapshot::deref`, the general mechanism seems not very useful and too error-prone without putting more thought into it.

PR #1177, opened 2024-10-20
builtin strategy

xldenis:
I've been using custom why3 strategies for ages, which save soooo much time when regenerating proofs. I've put together a template for my "tuned" Auto 3 strategy which reduces useless timeouts and attempts to avoid unnecessary proof tasks. 

When we move to `why3find` this won't be necessary, but this PR was easy to throw up in the meantime.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -512,6 +512,7 @@ dependencies = [
  "creusot-args",
  "directories",
  "hex",
+ "indoc",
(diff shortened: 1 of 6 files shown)


Armael:
neat! I will definitely be using this.

Armael:
I just tried it, it is much faster indeed, thanks!

PR #1181, opened 2024-10-21
extern spec for `Arguments::new_const`

arnaudgolfouse:


arnaudgolfouse:
Note that formatting (`println!("{}", 1)`) is not yet supported

Armael:
nice! Would it make sense to add some test(s) to show what is already supported?

arnaudgolfouse:
That "strings are experimental" warning is a bit annoying 😅

PR #1184, opened 2024-10-21
Ghost contracts: tweaks

Armael:
The first commit should be non-controversial.
The second one (changing the visibility of `into_inner` and `inner_logic`) to `open` may be, perhaps? I made this change after discovering that I was stuck on a proof because the SMTs were apparently not able to exploit the spec of `inner_logic`, whereas computing with the definition helped.
And anyway from a logical perspective the spec just paraphrases the definition, so I don't know if we lose anything.
This also makes me wonder if the `#[ensures ...]` can simply be removed...


arnaudgolfouse:
I think I agree with this, but why is there `#[open]` on `into_inner`? It should not be needed since it is not a logical function, no?

Armael:
Indeed, it does not make sense on `into_inner`! I don't know why I added it.

jhjourdan:
The fact that provers are not able to use the `ensures` clause is a bit worrying, but this is not the point of this PR.

In general, I agree that we should prefer transparent definitions over `ensures` clauses that state the same thing. Also, in this case, we should remove the `ensures` clause because it adds a lemma in the SMT goal, which can confuse the provers.

PR #1185, opened 2024-10-21
Output crates of different types in separate directories

### LINKED ISSUE #1178: VCs for libraries are dropped if a binary is present
repro example:

Cargo.toml:
```
[package]
name = "toto"
version = "0.1.0"
edition = "2021"

[dependencies]
creusot-contracts = { path = "/home/armael/recherche/creusot/creusot-contracts" }
```

`src/main.rs`:
```rust
use creusot_contracts::*;

fn main() {
    let x = toto::foo();
    proof_assert!(x@ >= 0);
}
```

`src/lib.rs`:
```rust
use creusot_contracts::*;

#[ensures(result@ == 42)]
pub fn foo() -> i32 {
   -1
}
```

Then launching `cargo creusot why3 ide` only shows VC for `main.rs`, and verification "passes".

Lysxia:
This fixes the issue for both multi-file and single-file workflows. Closes #1178.

`creusot-rustc` now chooses the output location under `--output-dir` depending on whether the crate is a lib (`target/creusot/mycrate`) or bin (`target/creusot/bin/mycrate`) or other. (That used to be done in `cargo-creusot` instead of `creusot-rustc` before #1172 but didn't really work; [more details here](https://github.com/creusot-rs/creusot/issues/1178#issuecomment-2427501817))

For single-file output, a new flag `--monolithic` is added.

Basically there are the following output modes:

- User calls `cargo creusot`, which defaults to multi-file output (calls `creusot-rustc` with `--output-dir=target/creusot/`).
- User calls `cargo creusot --monolithic` for single-file output (calls `creusot-rustc` with `--monolithic --output-dir=target/creusot/`, and it's `creusot-rustc` which sets the file name to `crate_name.coma`, possibly with a `bin/` directory in the middle); we keep this option around while we fix issues with the multi-file workflow (or in case we decide to revert back to single-file).
- Creusot's test suite calls `creusot-rustc --stdout`, no changes there.
- The option `--output-file=${FILE}` previously used for single-file output is now unused (but it uses the same code as `--stdout` so it doesn't hurt to keep it).

### DIFF
--- a/cargo-creusot/src/main.rs
+++ b/cargo-creusot/src/main.rs
@@ -35,10 +35,16 @@ fn main() -> Result<()> {
         coma_glob = None;
     } else {
         // default to --output-dir=target/creusot
-        let dir = make_coma_target(&cargo_md)?;
-        coma_src = dir.clone();
-        cargs.options.output_dir = Some(dir);
-        coma_glob = coma_src.to_str().map(|s| s.to_string() + "/**/*.coma");
+        let dir = match &cargs.options.output_dir {
+            Some(dir) => dir.clone(),
+            None => {
+                let dir = make_coma_target(&cargo_md)?;
+                cargs.options.output_dir = Some(dir.clone());
+                dir
+            }
+        };
+        coma_glob = dir.to_str().map(|s| s.to_string() + "/**/*.coma");
+        coma_src = dir;
     }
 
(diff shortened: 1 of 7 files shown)


xldenis:
> For single-file output, a new flag --monolithic is added.

You might need to update `ui.rs` in consequence.

PR #1187, opened 2024-10-22
Do not generate refinement obligations for `Snapshot::deref`

arnaudgolfouse:
Like #1175, but only for `Snapshot::deref`.


arnaudgolfouse:
This is probably the only place we need this anyways

xldenis (approved):

PR #1188, opened 2024-10-22
Make closure names more readable

Lysxia:
Before: `qy123zclosureqy35z42qy125z`  
After: `qyClosure42`


xldenis (approved):
Seems like an easy win. :+1: 

I think unless there are major issues going forward, you should try to avoid making too many changes to the legacy workflow to instead focus on next-generation stuff :)

PR #1189, opened 2024-10-22
Refactor generation of `use` declarations

Lysxia:
While reworking the structure of the generated modules, I kept getting lost in the logic of `clone_map.rs` and `clone_map/elaborator.rs`, because there are some module names that are computed in two ways.

For example in `impl DepElab for TyElab`:

- we call `names.insert(dep)` which computes modules names one way
- we discard the result
- then we recompute the module names from scratch to fill the `as_` and `name` fields of `UseDecl`.

So when changing the shape of module names I was never sure whether I've forgotten a branch somewhere. This PR tries to unify that logic by preserving the structure of the module names returned by `names.insert()` (not flattened to a `Symbol`) so we can reuse it.

- This PR also makes it more explicit which modules are prefixed with `M_` and which are `T_`
- For closure types, instead of adding a `_Type` suffix, we change the prefix from `M_` to `T_`
- In one-file-per-module mode we remove the now redundant `module` declaration. Why3 then implicitly wraps it in `Coma` so that changes how it is imported by other files.

- [x] ~~I still need to update the why3sessions~~ There are no proofs in `qyClosure*_Type` modules, so no changes to why3sessions are needed.

### DIFF
--- a/creusot/src/backend/clone_map/elaborator.rs
+++ b/creusot/src/backend/clone_map/elaborator.rs
@@ -53,11 +53,10 @@ impl<'a, 'tcx> Namer<'tcx> for ExpansionProxy<'a, 'tcx> {
         self.namer.normalize(ctx, ty)
     }
 
-    fn insert(&mut self, dep: Dependency<'tcx>) -> Kind {
+    fn insert(&mut self, dep: Dependency<'tcx>) -> &Kind {
         let dep = dep.erase_regions(self.namer.tcx);
         self.expansion_queue.push_back((self.source, dep));
-        let k = self.namer.insert(dep);
-        k
+        self.namer.insert(dep)
     }
 
     fn tcx(&self) -> TyCtxt<'tcx> {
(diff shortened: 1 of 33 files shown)


xldenis:
> we call names.insert(dep) which computes modules names one way
> we discard the result

Yea, this is definitely vestigial of the older approaches, good cleanups!

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -409,21 +409,16 @@ impl<'a, 'tcx> Expander<'a, 'tcx> {
...
                 } else if ctx.is_constant(dep.to_trans_id(ctx.tcx, self.param_env).unwrap()) {
                     LogicElab::expand(self, ctx, dep)
                 } else if matches!(ctx.def_kind(def_id), DefKind::Field) {
-                    if util::item_type(self.tcx, def_id) == ItemType::Field {
-                        def_id = self.tcx.parent(def_id);
-                    };
-                    let name = self.namer.insert(dep).qname();
-                    let modl = ctx.module_path(def_id);
-                    let name = if name.module.is_empty() { name } else { name.module_qname() };
+                    let Kind::Used(modl, alias, _) = self.namer.insert(dep) else { todo!() };
```
xldenis, line 418: Is this meant to be a `todo!` or unreachable?

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -330,25 +329,26 @@ impl DepElab for TyElab {
...
 
-                    let use_decl = Use { as_: None, name: qname.module_qname(), export: false };
+                    let use_decl = Use { as_: None, name: name.clone(), export: false };
 
                     vec![Decl::UseDecl(use_decl)]
                 } else {
-                    let name = names.insert(dep).qname();
+                    let k = names.insert(dep);
+                    let Kind::Used(modl, alias, _) = k else {
+                        debug!("TODO: expand: {:?}", k);
```
xldenis, line 343: leftover debug statement?

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -330,25 +329,26 @@ impl DepElab for TyElab {
...
+                        return vec![];
```
xldenis, line 344: why is this correct?

xldenis (commented):
I have a few questions, but overall I am highly favorable to this.

creusot/src/backend/clone_map/elaborator.rs
Lysxia: I'm actually not sure why. Unlike the above `todo!()` this branch is reached by the test suite and yet returning `vec![]` here doesn't change the output from before.

creusot/src/backend/clone_map/elaborator.rs
Lysxia: It was really a todo but the test suite doesn't reach it :shrug:

Lysxia: I checked and it should be unreachable. Replaced with `panic!`

creusot/src/backend/clone_map/elaborator.rs
Lysxia: I found that the vector returned in this case doesn't matter because it gets put in a map and is never looked up. I refactored this part again and left a comment about this.

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -330,26 +329,26 @@ impl DepElab for TyElab {
...
-                    let name = if name.module.is_empty() { name } else { name.module_qname() };
-                    let use_decl = Use { as_: Some(name.as_ident()), name: modl, export: false };
-                    vec![Decl::UseDecl(use_decl)]
+                match names.insert(dep) {
+                    Kind::Unnamed | Kind::Named(_) => {
+                        // Locally available names.
+                        // We return an empty vector because we have to return something,
+                        // but it will be ignored anyway: it is inserted into a map at a
+                        // key which will never be looked up.
+                        vec![]
```
xldenis, line 338: It would be good to figure out which types are causing this, but not a blocker. Normally everything in the graph here should be connected and thus used eventually...

xldenis (approved):

PR #1190, opened 2024-10-22
Test creusot contracts

arnaudgolfouse:
A lot of proofs should be revisited in the future to remove `trusted`, especially around iterators. But this is a start!


arnaudgolfouse:
@xldenis Does that look ok? (btw thanks a lot for the custom why3 strategy!)

xldenis (approved):
> (btw thanks a lot for the custom why3 strategy!)

I would encourage you to tune it further, it was just a first approximation better strategy, I'm convinced we could make it ~2x faster in practice.

Armael:
After making the following changes https://github.com/Armael/creusot/commit/b6da39b8e295d8d0132a78659a28d372e72f3d60 , running `cargo test --test creusot-contracts -- --bless` results in an empty `creusot-contracts.coma` file.
Any idea what the issue is?

Armael:
I'm also intruiged about the `FIXME` related to spans in the test runner code (and indeed when launching `./ide` on `creusot-contracts.coma` we can see that the current spans are wrong).
There seem to be two issues here:
- I would have expected `--spans-relative-to` to work here, when passed `../creusot-contracts` (since the test runner is executed from the `creusot` subdirectory)
- For some values of `--spans-relative-to`, we end up with a `.coma` file that contains references to spans but no `let%spans` (this makes why3ide fail to open the file completely)

arnaudgolfouse:
> After making the following changes [Armael@b6da39b](https://github.com/Armael/creusot/commit/b6da39b8e295d8d0132a78659a28d372e72f3d60) , running `cargo test --test creusot-contracts -- --bless` results in an empty `creusot-contracts.coma` file. Any idea what the issue is?

Ah the `Seq::concat_assoc` call is ill-typed, so compilation of creusot-contracts fails: then because you specified `--bless`, the test assumes that this is normal and the output is just empty... I'll add a guard for this case.

arnaudgolfouse:
> I'm also intrigued about the `FIXME` related to spans in the test runner code (and indeed when launching `./ide` on `creusot-contracts.coma` we can see that the current spans are wrong). There seem to be two issues here:
> 
>     * I would have expected `--spans-relative-to` to work here, when passed `../creusot-contracts` (since the test runner is executed from the `creusot` subdirectory)
> 
>     * For some values of `--spans-relative-to`, we end up with a `.coma` file that contains references to spans but no `let%spans` (this makes why3ide fail to open the file completely)

Yeah I fixed this, I had the order of `spans-relative` wrong: it should point to the directory of the `.coma` file, aka `--spans-relative-to=creusot/tests/creusot-contracts`.

PR #1191, opened 2024-10-23
Extern spec with a body

arnaudgolfouse:
Allow functions in `extern_spec!` to have a body, with corresponding proof obligations.


Armael:
What's the usecase/motivation behind this change?

arnaudgolfouse:
@Armael sorry, I did not see your message!

When writing `extern_spec!`, it can be easier to convince yourself that the spec is actually correct if there is an implementation attached to it. My personal motivation was `Option`'s methods in #1136, like this one:
```rust
#[requires(match self {
    None => true,
    Some(t) => predicate.precondition((&t,))
})]
#[ensures(match self {
    None => result == None,
    Some(t) => match result {
        None => predicate.postcondition_once((&t,), false) && resolve(&t),
        Some(r) => predicate.postcondition_once((&t,), true) && r == t,
    },
})]
fn filter<P: FnOnce(&T) -> bool>(self, predicate: P) -> Option<T>
{
    match self {
        None => None,
        Some(t) => if predicate(&t) { Some(t) } else { None },
    }
}
```
The spec is not easy to grasp, while the implementation is really straightforward. And in any case, we know that we did not introduce inconsistencies!

Note that it does not guarantee that the actual function has that actual behaviour: in this sense this still introduces an axiom.

Armael:
Thanks for the explanation, that makes sense indeed!

PR #1196, opened 2024-10-24
Improve perfomance of problematic tests

arnaudgolfouse:
These are the tests that failed the last 2 nightly.

I just ran bisect_hypothesis during lunch, and then cleaned the result ; maybe this is not what we want?


xldenis:
>  maybe this is not what we want?

My only suggestion would be to see if those `remove`s allow you to remove tactics.

arnaudgolfouse:
Nope 🤷

PR #1200, opened 2024-10-24
Closure Specification Inference

xldenis:
This PR is the reason we adopted Coma in the first place, it allows us to infer the specifications of closures! 

The implementation has one weakness: `postcondition_mut` expects a `&mut F` which means we need to be able to create a borrow in Pearlite. For this reason I added `TermKind::Borrow`, which is not super desirable. A motivated soul would refactor `closure_capture_subst` so that we can change the signature of `postcondition_mut` but I unfortunately don't currently have enough time to tackle this. 

I think that this is still a worthwhile addition in the meantime which makes closures actually usable.


arnaudgolfouse:
I see you updated some tests to use this, it looks good! Do you think `should_succeed/option.rs` could also benefit from this? 😁

PR #1201, opened 2024-10-25
Take why3 master into account when fixing creusot-contracts

arnaudgolfouse:
On those 2 goals, why3 master (currently b4960e45) causes regressions in performances of Z3. I don't know why that is, this PR is only a fix for nightly CI.


arnaudgolfouse:
Issue is tracked in why3 upstream: https://gitlab.inria.fr/why3/why3/-/issues/886

PR #1205, opened 2024-10-28
Why3find integration

Lysxia:
Add scripts to install why3find in `creusot-setup` and to run it in `cargo-creusot`.

```
cargo creusot config
cargo creusot prove
```

I wrote some instructions in `quickstart.md`.

### DIFF
--- a/cargo-creusot/src/main.rs
+++ b/cargo-creusot/src/main.rs
@@ -10,11 +10,16 @@ mod helpers;
 use helpers::*;
 mod why3_launcher;
 use why3_launcher::*;
+mod why3find_wrapper;
+use why3find_wrapper::{why3find_config, why3find_prove};
+
 enum Subcommand {
     // subcommand to pass on to creusot-rustc
     Creusot(Option<CreusotSubCommand>),
-    // subcommand to handle in cargo-creusot
+    // subcommands to handle in cargo-creusot
     Setup(SetupSubCommand),
+    Config(ConfigArgs),
+    Prove(ProveArgs),
 }
 use Subcommand::*;
 
@@ -26,6 +31,8 @@ fn main() -> Result<()> {
         None => Creusot(None),
         Some(CargoCreusotSubCommand::Creusot(cmd)) => Creusot(Some(cmd)),
         Some(CargoCreusotSubCommand::Setup { command }) => Setup(command),
+        Some(CargoCreusotSubCommand::Config(args)) => Config(args),
+        Some(CargoCreusotSubCommand::Prove(args)) => Prove(args),
     };
 
     match subcommand {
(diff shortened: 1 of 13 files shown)


xldenis (approved):
We need to start shipping `why3find` as part of the creusot distribution asap.

Armael (changes requested):
This is nice and I agree that we eventually want this, but right now the way why3find is installed, looked up in the PATH and configured feels a bit "bolted on on the side".
I would feel more comfortable merging commandline support for calling why3find once we have a nice way of installing it with creusot (like why3 is installed today).

I don't think this is necessarily very hard, off the top of my head, I'm thinking:
- add `why3find` to `creusot-deps.opam` (pinning it to a specific commit)
- extend `creusot setup` to find the `why3find` binary and symlink it in creusot's `bin` directory
Then there's the question of config files, I need to look more closely into how why3find does things, but I would very much like to avoid editing `~/.why3.conf` and instead continue using the one generated by `creusot setup`.

So my suggestion is to first do these changes in a separate PR and then rebase this one on top of it.

Lysxia:
Sounds reasonable.

> Then there's the question of config files, I need to look more closely into how why3find does things, but I would very much like to avoid editing `~/.why3.conf` and instead continue using the one generated by `creusot setup`.

I fixed the `why3find` commands to use our `why3.conf`.

> add why3find to creusot-deps.opam (pinning it to a specific commit)

The only temporary difficulty is that why3find depends on the released version of why3 at the moment. We could fork it for a while but it seems simpler to wait for the upcoming why3 release.

Armael:
> We could fork it for a while but it seems simpler to wait for the upcoming why3 release.

As you prefer; I think forking would be reasonable as well, if it helps making progress.

Lysxia:
This PR now includes the installation of why3find.

cargo-creusot/src/main.rs
```diff
@@ -10,11 +10,16 @@ mod helpers;
...
+mod why3find_wrapper;
+use why3find_wrapper::{why3find_config, why3find_prove};
+
 enum Subcommand {
     // subcommand to pass on to creusot-rustc
     Creusot(Option<CreusotSubCommand>),
-    // subcommand to handle in cargo-creusot
+    // subcommands to handle in cargo-creusot
     Setup(SetupSubCommand),
+    Config(ConfigArgs),
```
Armael, line 21: Are there cases where it's useful for the user to manually call the `config` subcommand? Or could we instead not expose it and always rely on `prove`?

creusot-setup/src/tools.rs
```diff
@@ -349,3 +370,7 @@ pub fn symlink_file<P: AsRef<Path>, Q: AsRef<Path>>(original: P, link: Q) -> std
         std::os::windows::fs::symlink_file(original, link)
     }
 }
+
+pub fn why3find_install() {
+    Command::new("why3find").arg("install").arg("prelude").status().unwrap();
```
Armael, line 375: where does the prelude gets installed when running this command?

cargo-creusot/src/main.rs
Lysxia: Actually I misunderstood the purpose of `config` yesterday. The `why3find.json` that it generates is meant to be committed in git, so `config` should only be run when creating a project, and it shouldn't be run in `prove`.

I was confused because I was only thinking of the simplest projects which would just work with a basic config.

Armael: OK, then having a separate config command makes sense for now! (Then ultimately when we make `creusot setup` project-local we should be able to combine `config` and `setup`)

Armael (approved):
LGTM now  :)

Issue #943, opened 2024-02-20
`cargo creusot why3 ide` does not launch after a successful `cargo creusot`


Armael:
If I run `cargo creusot` and it terminates successfully, then running `cargo creusot why3 ide` immediately terminates "successfully" without launching why3ide.
This had me scrambling for a while trying to understand what was wrong with my installation of why3ide.

I guess this is caused by caching of build results performed by cargo, but I don't know where to start looking to fix the issue.

xldenis:
> I guess this is caused by caching of build results performed by cargo, but I don't know where to start looking to fix the issue.

Yea, there is a lot that needs improving here. Off the top of my head some things I think we should do:

1. Set a custom cargo target so that creusot artifacts are not built in `target/debug` but instead in `creusot/debug` or even potentially `target/creusot` (if we can figure out that second one). This would also potentially avoid trashing the cargo cache as often and make `creusot-contracts` more reliable when switching between rustc and creusot modes.
2. To better control caching of cargo, we can leverage the various cargo instructions which can be passed back up: https://doc.rust-lang.org/cargo/reference/build-scripts.html

laurentder:
Currently, calling `cargo creusot why3 ide` will execute the `cargo check` command.
- On the first call, `cargo check` will execute `creusot-rustc`, which will launch `why3 ide`.
- On subsequent calls, if the source has not been modified, `cargo check` will not call `creusot-rustc` (you can see documention [here](https://doc.rust-lang.org/cargo/commands/cargo-check.html)). If `creusot-rustc` is not called, `why3 ide` will not be launched.

If the goal is to launch `why3 ide`, the approach of executing the `cargo check` command may not be the right one. Another approach might be to modify `cargo creusot`, when calling `cargo creusot why3 ide`. We could do this in two stages
 - first run cargo check.
 - then launch from cargo creusot why3 ide.

 @xldenis, Is this approach right for you? Currently, when `creusot-rustc` launches `why3 ide`, it uses context information from its compilation (TranslationCtx, etc.). Is this information necessary to launch `why3 ide`?

> To better control caching of cargo, we can leverage the various cargo instructions which can be passed back up: https://doc.rust-lang.org/cargo/reference/build-scripts.html

[...]

PR #1211, opened 2024-10-30
Add --check option to not generate output

### LINKED ISSUE #1207: Feature request: a `cargo creusot check` command that does not output .coma files (for IDE support)
The current workflow for editing creusot sources in an editor is to override rust-analyzer `cargo check` command with `cargo creusot`. However this currently overwrites the .coma files that are generated, with is super annoying when running `cargo creusot` separately from the commandline, e.g. to use different options, or debug something.

I think a solution could be to have a separate command (`cargo creusot check`) that only runs the "diagnostics" part of `cargo creusot", but without writing to the filesystem.

Lysxia:
Close #1207


Armael:
Thanks! I had a look yesterday but timeouted, and this looks just like what I had in mind.

Lysxia:
There's still something going on because after `cargo creusot` (`--check` or not), `cargo creusot` again doesn't do anything if the source files haven't changed (`touch` works to invalidate whatever cache is responsible).

Armael:
Ah, that's a problem indeed... I remember that we looked at a similar caching issue with @laurentder : https://github.com/creusot-rs/creusot/issues/943#issuecomment-2025432532

Lysxia:
`cargo` doesn't have a `--force` option according to this issue https://github.com/rust-lang/cargo/issues/1101. A workaround is to do `cargo clean -p yourcrate` beforehand, which is better than `cargo clean` because it doesn't clean your built dependencies.

xldenis:
The real solution is that creusot should use a different target directory or should change the cache key used by cargo. I think using a different target directory is likely better

Lysxia:
We still want `cargo creusot check` for diagnostics to still use the usual target, don't we?

xldenis:
I don't think so ? The creusot diagnostics should be a superset right?

Armael:
Is this summary of the issue correct?:
- `cargo` runs `creusot-rustc` (our `rustc` wrapper). This command is cached, it is not run again if the source files are the same and the command is the same
- however, parameters from `cargo creusot` to `creusot-rustc` are passed in an environment variable, so `cargo` does not see them
- these parameters (e.g `--check`, `--monolithic`, etc) change the output of `creusot-rustc` 

If so could we make `cargo` aware somehow of the parameters passed to `creusot-rustc` to take them into account in its caching strategy?

xldenis:
Yes and cargo has a mechanism to tell it to care about specific env vars https://doc.rust-lang.org/cargo/reference/build-scripts.html#change-detection

Lysxia:
> Yes and cargo has a mechanism to tell it to care about specific env vars https://doc.rust-lang.org/cargo/reference/build-scripts.html#change-detection

That seems to be about rerunning `build.rs` specifically. Or does that imply rebuilding one's library?

Lysxia:
To answer my own question, I tested it and rerunning the build script causes the library to be rebuilt. It seems strange to have to use a build script for that though.

xldenis:
I think we can trigger the same behavior without a build script since we're a cargo wrapper

Lysxia:
The issue with caching is addressed in #1212

Armael:
Thanks a lot!!

PR #1212, opened 2024-10-30
Make cargo rerun creusot when options change

Lysxia:
See https://github.com/creusot-rs/creusot/pull/1211#issuecomment-2446789994

We pass `CreusotArgs` to creusot-rustc via a `--creusot=` argument instead of an environment variable.

---

Old approach (outdated):

It isn't possible to directly tell `cargo` to include an environment variable (`CREUSOT_ARGS`) in its fingerprints. There is an option `cargo::rerun-if-env-changed` that a `build.rs` can emit but which is not accessible anywhere else. The workaround implemented in this PR is to insert a hash of `CREUSOT_ARGS` in the command line arguments of `creusot-rustc` which `cargo` includes in its fingerprint.

### DIFF
--- a/cargo-creusot/src/main.rs
+++ b/cargo-creusot/src/main.rs
@@ -133,6 +133,17 @@ fn invoke_cargo(args: &CreusotArgs) {
         .env("RUSTC_WRAPPER", creusot_rustc_path)
         .env("CARGO_CREUSOT", "1");
 
+    // Append flags to any pre-existing ones
+    // CARGO_ENCODED_RUSTFLAGS contains options to pass to rustc, separated by '\x1f'.
+    // https://doc.rust-lang.org/cargo/reference/environment-variables.html
+    let mut cargo_encoded_rustflags = match std::env::var("CARGO_ENCODED_RUSTFLAGS") {
+        Ok(flags) => flags + "\x1f",
+        Err(_) => String::new(),
+    };
+    cargo_encoded_rustflags.push_str("--creusot=");
+    cargo_encoded_rustflags.push_str(&serde_json::to_string(&args).unwrap());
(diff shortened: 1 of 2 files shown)


Armael (approved):
Looks good to me! I've checked that the refactored logic is equivalent, AFAICT.

Armael:
For reference, the doc about `CARGO_ENCODED_RUSTFLAGS` can be found here: https://doc.rust-lang.org/cargo/reference/environment-variables.html . Maybe a link to the page could also be added in the code as a comment? (to explain the magic `\x1f` separator)

Lysxia:
I added a link :)

Armael:
I only realize now (when testing this + #1213 on top) that this still causes e.g. `cargo creusot --monolithic why3 ide` and `cargo creusot --check` to invalidate each other's caches.
This is a logical consequence of the current implementation: the two commandlines are different so cargo correctly rebuilds everything. (This would be indeed necessary if the two commandlines were `cargo creusot --monolithic` and `cargo creusot`.)

This made me wonder if your initial approach was in fact the right one, since it gives us more flexibility: when computing the hash we can chose to ignore some options that we know are not relevant wrt caching.
But I believe this doesn't help for the specific conflict I'm mentionning: I can't think of a way of computing hashes so that `cargo creusot --check` would not cause rebuilds with respect to both `cargo creusot --monolithic` and `cargo creusot`, while still allowing the last two to conflict with each others…

Armael:
Actually, making `cargo creusot --check` and `cargo creusot [--monolithic]` equivalent wrt caching is also not what we want; if one runs `cargo creusot --check` first then this does not generate any coma file, then we still want to generate them when running `cargo creusot`...
I guess I need to take a step back and think about what is really needed at a high level...

PR #1213, opened 2024-10-31
Refactor creusot-rustc argument parser

Lysxia:
Instead of guessing when `creusot-rustc` is being called by `cargo` to decide whether rust options or creusot options come first, we make rust options always come first to simplify the parsing logic.

New usage:

```
creusot-rustc [RUST_OPTIONS] -- [CREUSOT_OPTIONS]
creusot-rustc [RUST_OPTIONS] --creusot=JSON_CREUSOT_OPTIONS
```

There was some funny thing going on with `rust_flags` because it was used also by `cargo-creusot` to forward arguments to `cargo` all while being uselessly sent to `creusot-rustc`. The original purpose of that field (forwarding arguments from `creusot-rustc`  to `rustc`) is made obsolete by this PR, so I renamed it to `cargo_flags` and moved it accordingly.

### DIFF
--- a/cargo-creusot/src/main.rs
+++ b/cargo-creusot/src/main.rs
@@ -54,10 +54,9 @@ fn main() -> Result<()> {
                 why3_path: config_args.why3_path.clone(),
                 why3_config_file: config_args.why3_config.clone(),
                 subcommand: creusot_rustc_subcmd.clone(),
-                rust_flags: cargs.rust_flags,
             };
 
-            invoke_cargo(&creusot_args);
+            invoke_cargo(&creusot_args, cargs.cargo_flags);
 
             if let Some((mode, coma_src, args)) = launch_why3 {
                 let mut coma_files = vec![coma_src];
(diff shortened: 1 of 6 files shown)


cargo-creusot/src/main.rs
```diff
@@ -129,8 +128,8 @@ fn invoke_cargo(args: &CreusotArgs) {
     let mut cmd = Command::new(cargo_path);
     cmd.arg(format!("+{toolchain}"))
         .arg(&cargo_cmd)
-        .args(args.rust_flags.clone())
-        .env("RUSTC_WRAPPER", creusot_rustc_path)
```
Armael, line 133: why the change from `RUSTC_WRAPPER` to `RUSTC`? (I have no idea what these variables do)

Armael:
I've forgotten the details of how this code is running but this looks like a good idea. I remember stumbling onto this `rust_flags` business and finding it confusing.

cargo-creusot/src/main.rs
Lysxia: `RUSTC=foo` replaces `rustc` with `foo`.
`RUSTC_WRAPPER=foo`  replaces `rustc` with `foo /path/to/rustc`, presumably for tools that just end up calling the underlying `rustc`. We abused this option to disambiguate when `cargo` called `creusot-rustc`.

PR #1214, opened 2024-10-31
test: run cargo creusot with a separate --target-dir

Lysxia:
Maybe we want to do this change systematically in cargo-creusot


arnaudgolfouse (approved):

Armael:
I think we should also do that in `cargo-creusot` indeed.

PR #1220, opened 2024-11-05
Change the algorithm in validate_terminates

arnaudgolfouse:
@jhjourdan this is what we talked about.

The new algorithm is stricter:
```rust
trait Foo {
    fn foo1();
    fn foo2();
}

impl Foo for i32 {
    fn foo1() {}
    fn foo2() { g() }
}

fn f<T: Foo>() { <T as Foo>::foo1() }
fn g() { f::<i32>(); }
```
is now rejected, saying that `f might call i32::foo2` because of the `T: Foo` bound.


arnaudgolfouse:
This is technically incorrect, but there should be a path to fix that (see #1232). Also usage of rustc's internal APIs is better, avoiding some crashes.

jhjourdan:
We discussed this with @arnaudgolfouse. We think that this is an improvement over the current state, even though there are some remaining problems. Still, we need this to be merged before #1228, so I'm going to merge this, and postpone solving #1232 later.

PR #1222, opened 2024-11-05
More informative `expl:` labels for loop invariants

Lysxia:
- Loop invariants are numbered `loop invariant 1`, etc. to make it easier to map Why3 goals to their source location at a glance.

- Invariants obtained by desugaring `for` loops and mutable borrows in loops are annotated with something different than "loop invariant": "for invariant" or "mut invariant" respectively.

- I fixed spans for the desugaring of `for` loops, so invariants stop reporting their source location as the first invariant. The reason was that the macro for the first invariant is the one that desugars `for` loops, and the spans in the generated code default to the macro that generated it. That is the behavior of the `quote!` macro; the solution is to use `quote_spanned!` which takes an explicit span parameter.

- Replacing `quote!` with `quote_spanned!` introduced some warnings about redundant semicolons, that were somehow swallowed before?! Those are now fixed.


creusot/src/translation/fmir.rs
```diff
@@ -213,9 +213,25 @@ impl<'tcx> Branches<'tcx> {
     }
 }
 
+#[derive(Debug, Clone, Copy)]
+pub enum InvariantKind {
+    /// Invariant that the user wrote
+    Source,
+    /// Invariant generated by desugaring `for` loops
+    For,
+    /// Invariant for mutated variables with type invariants
```
arnaudgolfouse, line 222: Is this really what this kind is used for? Looking at its usage, I thought this was about mutable variables that are not mutated in the loop. Are type invariants involved?

Lysxia: I misread the comment on `infer_proph_invariants`. It looks like it's about remembering the prophecies of mutable borrows, and doesn't depend on whether they are being mutated (e.g., test `while_let.rs`)

Lysxia:
@Armael does that fix #1225 ?

Armael:
Let me try…

Armael:
No, it looks like the issue is still there, unfortunately.

Lysxia:
Fixed:

image
image

Those two invariants desugar to terms that start respectively with `::creusot_contracts::model::View::view` and `::creusot_contracts::__stubs::implication` and the spans for the `::` were being set to the call site, i.e., the first `invariant` macro. Fixed by setting the right span explicitly.

xldenis (approved):
Seems like an across the board improvement in spans from specifications! Glad to see this long standing issue with invariants fixed! There should be a few gh issues you can close.

PR #1224, opened 2024-11-06
Seq: rename {push,pop}->{push,pop}_back, add {push,pop}_front

Armael:


creusot-contracts/src/logic/seq.rs
```diff
@@ -134,11 +134,19 @@ impl<T: ?Sized> Seq<T> {
         absurd
     }
 
+    #[trusted]
+    #[logic]
+    #[open(self)]
+    #[creusot::builtins = "prelude.prelude.Seq.cons_"]
```
arnaudgolfouse, line 140: this uses "prelude.prelude.Seq.cons_", but push_back uses "seq.Seq.snoc", is there a reason for this difference?

Armael: `seq.Seq.snoc` is from the why3 stdlib. Ideally we would also use `seq.Seq.cons` (also from the why3 stdlib), but the order of arguments is different from the one we want! (`self` has to be the first argument, right?)
Hence the `cons_` wrapper to swap the arguments.

xldenis: I guess this is moving the wrapper down one level.

xldenis: My recollection was that Seq had an internal wrapper that flipped the arguments

Armael: I couldn't see such a wrapper, but maybe I missed it?

xldenis: https://github.com/creusot-rs/creusot/blob/master/creusot-contracts/src/logic/seq.rs#L48 ?

xldenis: Ah i guess i only did it for sets: https://github.com/creusot-rs/creusot/blob/master/creusot-contracts/src/logic/fset.rs#L50-L52

xldenis: we're not 100% consistent either way, I have a slight preference for the rust wrapper approach but i wouldn't consider it blocking.

Armael: I agree that it is better. I pushed a new version with a rust wrapper instead of one in why3

Armael:
I keep getting owned by tests that replay locally but not in CI… I have the suspicion that the new "Creusot" strategy in why3ide is generating more fragile sessions than the others because of the short initial timeout for z3.

xldenis:
> I keep getting owned by tests that replay locally but not in CI… I have the suspicion that the new "Creusot" strategy in why3ide is generating more fragile sessions than the others because of the short initial timeout for z3.

Oh that's a good point. It really reinforces that we should make steps work for our sessions. This would ensure CI is as reproducible as local.

PR #1226, opened 2024-11-06
Fix a ICE in the final places analysis

arnaudgolfouse:
When trying to write to a shared reference, creusot generated a ICE file, which was annoying: this fixes that.


xldenis:
Include a test?

PR #1227, opened 2024-11-07
hide generated files for tests from github diffs 

Armael:
This is an attempt to use https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github to hide generated .coma/session files from github diffs by default, to make it easier to review PRs.


jhjourdan:
Yeah, I think this is an improvement (even though reading the diffs for coma and xml files is important).

PR #1228, opened 2024-11-07
Inline type translations in the client modules instead of cloning modules containing polymorphic definitions

jhjourdan:
This is accompanied with major refactoring/simplification of the backend.


jhjourdan:
The only failing test should be `iterator/17_filter`, which is waiting for #1220 to fix a bug in normalization.

creusot/src/backend/clone_map.rs
```diff
@@ -324,89 +281,41 @@ impl<'tcx> CloneNames<'tcx> {
...
 impl NameSupply {
     pub(crate) fn freshen(&mut self, sym: Symbol) -> Symbol {
         let count: usize = *self.name_counts.entry(sym).and_modify(|c| *c += 1).or_insert(0);
+        // FIXME: if we don't do use the initial ident when count == 0, then the ident clashes
+        // with local variables
+        /*if count == 0 {
+            sym
+        } else {*/
         Symbol::intern(&format!("{sym}'{count}"))
+        /*}*/
```
xldenis, line 318: what's the status of this

creusot/src/translation/specification.rs
```diff
@@ -32,7 +43,8 @@ impl<'tcx> PreContract<'tcx> {
 
     pub(crate) fn normalize(mut self, tcx: TyCtxt<'tcx>, param_env: ParamEnv<'tcx>) -> Self {
         for term in self.terms_mut() {
-            normalize(tcx, param_env, term);
+            *term =
+                normalize(tcx, param_env, std::mem::replace(term, /*Dummy*/ Term::mk_true(tcx)));
```
xldenis, line 47: Why change the signature like this? If you want it to be consisent wouldn't it be better to change `PreContract::normalize` instead?

creusot/src/backend/clone_map.rs
jhjourdan: I think we should change the design of name handling in Creusot. Here is my thoughts:
- The `Ident` type of the Why3 crate should carry a non-freshened (but mangled) string, and a unique identifier (e.g., u64).
- `Ident::new` should generate a fresh u64, and the creusot crate should carry around the same `Ident` when the refer to one variable in the same scope.
- When printing, the Why3 crate should track the content of the environment, and freshen identifiers if there are some clashes.

I already discussed with @Lysxia about all that. He said he would work on this, but we are waiting for this PR because his change would conflict a lot.

creusot/src/backend/dependency.rs
```diff
@@ -165,54 +100,55 @@ impl<'tcx> Dependency<'tcx> {
...
+            _ => None,
         }
     }
 
     pub(crate) fn is_closure_spec(&self) -> bool {
         matches!(self, Dependency::ClosureSpec(_, _, _))
     }
 
-    #[inline]
+    // FIXME: this function should not be necessary, dependencies should not be created non-normalized
```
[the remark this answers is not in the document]
jhjourdan: Do we? Where?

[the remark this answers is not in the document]
xldenis: I don't know if _we_ do, but rustc definitely can, whether `erase_regions` is needed should not be based off whether we do normalization but rather whether we erase.

creusot/src/translation/specification.rs
jhjourdan: Yeah, I know.... My goal was to make these function normalize types in addition to resolving traits, but I failed because of `PartialEq::eq`. Given that rustc normalization consumes its argument, these function should as well.

I'm open to reverting this change. But the whole situation wrt when normalization happens is really bad, it is unclear when we have a type whether it is normalized or not (these functional invariants are not documented, and probably not maintained everywhere as well). I think there are two reasonable choices:

- If we decide to normalize eagerly, then this signature has to change, because `normalize` should normalize all the types, too. But we need to find a solution with `PartialEq::eq`.
- If we decide to normalize lazilly, then this is a rather big change. Hash tables of dependencies need to use type unification for testing equality, etc...

I think the last solution is better, but it requires more work.

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -147,54 +154,40 @@ impl DepElab for LogicElab {
...
-                | Dependency::StructuralResolve(_)
-                | Dependency::ClosureSpec(_, _, _)
+            Dependency::Item(_, _) | Dependency::TyInvAxiom(_) | Dependency::ClosureSpec(_, _, _)
         ));
 
         // TODO: Fold into `term`, but requires first some sort of
         // handling for axioms
         if let Dependency::TyInvAxiom(ty) = dep {
-            return expand_ty_inv(elab, ctx, ty);
+            return expand_ty_inv_axiom(elab, ctx, ty);
```
xldenis, line 163: Any reason why you couldn't do that?

xldenis:
I've been trying to review this but the size of the changes makes this challenging at best. The refactoring seems good though, both because we finally get rid of type cloning but also all the other simplifications and code motion that came along with it. 

As we discussed, there is still this concern about the `ParamEnv` that we need to figure out, but hopefully this can be delayed to a future date. 

I think we should merge roughly as is, I don't see how I'll be able to effectively review it further. The generated code changes seem correct, and the rust code is cleaner.

creusot/src/backend/clone_map/elaborator.rs
jhjourdan: I could not do what?

I guess you mean folding this into term, but I have to say I don't really see how this could work. `TyInvAxioms` means inserting a new *axiom*, while this piece of code inserts a new logical definition.

creusot/src/validate_terminates.rs
```diff
@@ -102,7 +105,7 @@ pub(crate) fn validate_terminates(ctx: &mut TranslationCtx) {
     }
 
     // detect mutual recursion
-    let cycles = petgraph::algo::kosaraju_scc(&call_graph);
+    let cycles = petgraph::algo::tarjan_scc(&call_graph);
```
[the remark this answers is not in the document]
jhjourdan: Not much. Tarjan algorithm is usually preferred, because it has a lower memory footprint, and is probably more efficient too (it requires only one traversal of the graph).

Issue #421, opened 2022-06-13
Ensures and requires clause order is reversed when translated into Why3


jhjourdan:

xldenis:
This is still an issue in Coma, though its importance is less clear there.

jhjourdan:
I guess the problem is the same for `invariant` closes?

jhjourdan:
Is it really still reversed? `contract_clauses_of` seems to contain a call to `rev` to reverse the order....

Lysxia:
Closed in #1229

PR #1229, opened 2024-11-08
Reorder invariants and contracts and more informative 'expl:' labels for function contracts

Lysxia:
Goals now have the following labels, replacing the dry `loop invariant`, `precondition`, `postcondition`:

- `loop invariant {#INDEX}` (i.e., `loop invariant` or `loop invariant #0`, `#1`, etc. for multiple invariants)
- `FNAME ensures {#INDEX}` (where `FNAME` is the name of the function)
- `FNAME provides {#INDEX}`
- `FNAME 'ARG' type invariant` (where `ARG` is the name of the relevant argument)
- `FNAME result type invariant`

### Summary of changes

- Restore ensures, requires, invariant in user-provided order
- Change 'loop invariant 0' to 'loop invariant #0' for consistency with ensures and requires
- Only add `#0` if more than one clause
- Also add `expl:` labels in translation of logic functions

---

The type inference issue from before is solved by putting "for invariants" first (I added a comment about it in `creusot-contracts-proc/src/invariant.rs`).

OUTDATED SUMMARY BELOW

Original title: Reorder loop invariants

Loop invariants appear in reverse order in the Coma output. In addition to being confusing visually, the order also matters because later assertions in Coma have earlier assertions in their context. One probably expects later invariants on the same loop to be able to use earlier ones, rather than the opposite.

This PR preserves the position of invariants generated by creusot relative to user-written invariants in the Coma output to minimize proof breakage: "mut invariants" first, then "user invariants", then "for invariants". My intuition would be to always put generated invariants first because they should be easy to prove and may be used by user-written invariants, but in practice that breaks even more proofs. I have to check how many more exactly.

There are two kinds of breakage: broken proofs, and broken type inference.

- [x] Broken proofs: `iterators/17_filter.rs`, `knapsack_full.rs`, fixed by reordering the invariants in the source
- [x] Broken type inference: `hillel.rs`, `iterators/03_std_iterators.rs`. The issue is that the desugaring of `for` loops produces the following unannotated let bindings:

    https://github.com/creusot-rs/creusot/blob/52a602aaf341733c3347c53a67c8a342da8e5125/creusot-contracts-proc/src/invariant.rs#L158-L160

    Morally, we have `it: T`  where `T: Iterator`, and `produced: Snapshot<Seq<&T::Item>>>`, but the fact that they use the same `T::Item` is not inferrable from those definitions. Before this PR, unification happens when checking one of the generated invariants:

    ```
    ::creusot_contracts::std::iter::Iterator::produces(#iter_old.inner(), #produced.inner(), #it)
    ```

    However, this PR moves those generated "for invariants" *after* user-written invariants for the purposes of typechecking (before this PR that happened after typechecking). That makes type inference of generated code (`it` and `produces`) dependent on user-written code. That fails when you try to use a trait method on `produces`  when its type hasn't been inferred yet.

    I can think of two fixes:

    - We could move the generated "for invariants" back before user-written invariants. That has the issue of breaking more proofs.
    - Add some kind of annotation to infer the right types. I found this one which seems not too intrusive, although an ideal solution would have no impact on the generated code at all:

        ```rust
        let mut #it = ::std::iter::IntoIterator::into_iter(#iter);
        let #iter_old = snapshot! { #it };
        let mut #produced = snapshot! { ::creusot_contracts::logic::Seq::EMPTY };
        { type Unify<'a, T> = (&'a T, &'a Snapshot<Seq<<T as ::std::iter::Iterator>::Item>>);
          let _: Unify<'_, _> = (&#it,&#produced); }
        ```


jhjourdan:
Related to this PR is issue #421, which we are not sure is an actual issue. Could you please check, and fix it in this PR?

jhjourdan:
> The resulting Coma: https://gist.github.com/Lysxia/a7354b9f158ab8194d8b6de46830a471

This just passes easily with auto level 1. I think you submitted knapsack.coma instead of knapsack_full.coma

Lysxia:
> > The resulting Coma: https://gist.github.com/Lysxia/a7354b9f158ab8194d8b6de46830a471
> 
> This just passes easily with auto level 1. I think you submitted knapsack.coma instead of knapsack_full.coma

Oops. Updated.

jhjourdan:
> `knapsack_full.rs` I don't understand, the failing invariant is not on a for loop, the only difference in the Task seems to be the order of assumptions.

If you do `split_vc`, then the goal is easilly discharged by Z3. I think this is a reasonable fix. Of course, we could investigate more about the matching loop that is making provers loop in this case, and add relevant triggers, but this is irrelevant for this PR.

jhjourdan:
Weird, on my machine, the timing indicates that raising the timeout should not be necessary even on a slow machine:

image

Lysxia:
I'm looking at goal 118

Lysxia:
`sum.rs` passes if I  add `proof_assert!(i@ == produced.len());`

jhjourdan:
What if you use:

```rust
    #[invariant(2 * sum@ == produced.len() * (produced.len() + 1))]
```

instead of:

```rust
    #[invariant(sum@ == produced.len() * (produced.len() + 1) / 2)]
```

creusot-contracts-proc/src/lib.rs
```diff
@@ -392,14 +392,15 @@ impl Parse for Assertion {
 pub fn proof_assert(assertion: TS1) -> TS1 {
     let assert = parse_macro_input!(assertion as Assertion);
     let assert_body = pretyping::encode_block(&assert.0).unwrap_or_else(|e| e.into_tokens());
+    let location = format!("proof_assert L{}", Span::call_site().start().line);
```
Lysxia, line 395: I'm trying to store the line number of each `proof_assert` here but I keep getting 0.

Lysxia: Postponing to other PR

Lysxia:
After reordering function contracts (not only loop invariants), the problem with `sum.rs` fixed itself...

Lysxia:
Would it make sense to rename "loop invariant" to "invariant" to match what the user wrote?

xldenis:
I wonder if in a future PR we could introduce explicit syntax for labels in preconditions/postconditions I could imagine a good amount of them can be concisely summarized.

Lysxia:
That's definitely feasible. My thought is that as we take some of the workflow out of why3 IDE that will reduce the usefulness of more human-readable labels. In any case we can wait and see if a use case appears.

Lysxia:
In the end I did have to change `knapsack_full.rs` (move an invariant) and `sum.rs` (rewrite `... = ... / 2` to `... * 2 = ...`). Tell me if there should be a better way.

creusot/tests/should_succeed/sum.rs
```diff
@@ -2,10 +2,10 @@ extern crate creusot_contracts;
 use creusot_contracts::*;
 
 #[requires(n@ < 1000)]
-#[ensures(result@ == n@ * (n@ + 1) / 2)]
+#[ensures(result@ * 2 == n@ * (n@ + 1))]
```
jhjourdan, line 5:
```suggestion
#[ensures(result@ == n@ * (n@ + 1) / 2)]
```

I understand about the change bellow.  But this one should not be necessary.

jhjourdan:
Let's merge as soon as CI passes.

PR #1233, opened 2024-11-13
Add assertion locations

Lysxia:
This is half a PR and half an issue: I can't figure out how to use the `span-location` feature of `proc-macro2` (which is really the unstable feature `proc_macro_span` in rustc). This compiles but the line numbers are all `0`. Is there something wrong with how we wrap rustc?

Beyond this PR I was thinking of getting Creusot to output span information somewhere so that Creusot IDE can work without parsing rust code.

### DIFF
--- a/creusot-contracts-proc/Cargo.toml
+++ b/creusot-contracts-proc/Cargo.toml
@@ -16,5 +16,5 @@ quote = "1.0.35"
 uuid = { version = "1.3", features = ["v4"] }
 pearlite-syn = { version = "0.3", path = "../pearlite-syn", features = ["full"] }
 syn = { version = "2.0.15"}
-proc-macro2 = "1.0.29"
+proc-macro2 = { version = "1.0.29", features = ["span-locations"] }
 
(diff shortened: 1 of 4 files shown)


arnaudgolfouse:
This is weird: it seems `Span::call_site()` contains the information we want, but it is lost when calling `.start()` ??
However using `Span::call_site().unwrap().line()` works... I have no idea what the difference is.

Lysxia:
> Why can't you get the span from inside creusot instead?

That could be another way. I'm not fond of it because half the tokens at that point don't come from the source so I have to think a bit to get the right span from the right token.

>However using `Span::call_site().unwrap().line()` works... I have no idea what the difference is.

That's interesting because `start()` returns (a struct containing) 0 if a flag `proc_macro_span` is not set, and that seems to come from a `build.rs`. I'm not sure we handle build scripts completely right.

https://docs.rs/proc-macro2/latest/src/proc_macro2/wrapper.rs.html#478-489

Lysxia:
It seems that we are already getting decent spans for assertions so I'll postpone investigating this issue to hopefully never.

Issue #1232, opened 2024-11-12
Remaining unsoundness in the termination check


arnaudgolfouse:
After #1220, the termination check must be reworked (again), as it allows the following two incorrect examples:

### Recursive instantiation

```rust
trait Tr1 {
    fn f();
}

trait Tr2 {
    fn g();
}

impl<T> Tr1 for T where T : Tr2 {
    fn f() {
        T::g()
    }
}

impl Tr2 for u32 {
    fn g() {
        h()
    }
}

fn call_f<T: Tr1>() {
    T::f()
}

fn h() {
    call_f::<u32>()
}
```

When calling `call_f`, the bound `T: Tr2` does not appear, but it is used. To visit trait bounds correctly, see in `rustc_traits/src/codegen.rs`, the implementation of function `codegen_select_candidate`.

### Default function uses trait instance

```rust
trait Tr {
    fn f();
}

default impl<T> Tr for T {
    fn f() { g::<Self>(); }
}

fn g<T: Tr>() {
    <T as Tr>::f();
}
```

This is never flagged by the current check, even if we do something like
```rust
impl Tr for i32 {}

fn h() {
    <i32 as Tr>::f();
}
```

We should forbid `f` from using the trait instance it is defining.

Lysxia:
To solve this, there's a simpler way than visiting the proof tree. We just need to keep track of unimplemented function calls. For example:

```rust
impl<T> Tr1 for T where T : Tr2 {
    fn f() {
        T::g()
    }
}
```

Here we remember that `f` calls `T::g`, which is not yet defined. Then:

```rust
fn call_f<T: Tr1>() {
    T::f()
}
```

We remember that `call_f` calls `T::f`.

```rust
fn h() {
    call_f::<u32>()
}
```

`h` calls `call_f::<u32>`, which we remembered that it calls `<u32>::f`, which calls `<u32>::g`, which calls `h`.

jhjourdan:
That's a nice idea, thanks. We already discussed something that look like this, and a weak version of this is already used for default methods. However, this causes two problems in the general case:

1- this breaks abstraction, because `f` may be opaque and therefore callers should not rely on its body.

2- if we instantiate `f` with a type `T` such that `T::g` calls `f`, then it seems to me that in some situations the loop can only appear at the instantiation point of `f`. This is very misleading because the definition of `f` and `g` may appear in different crates, and the end user of `f` is expecting things that are already declared compose nicely.

[...]

PR #1234, opened 2024-11-13
Use the new trait solver.

jhjourdan:
We will nee too do that anyway, and this may help for #1232.

The main issue I encountered is that the new trait solver does not allow ambiguity between a user impl and an impl coming from ParamEnv when searching for an instance. This is fixed by removing all such ambiguities in our code base. At some point, we will need to figure out a way to report a pleasant error message when finding trait method instances fails because of such an ambiguity.


creusot-contracts/src/std/ops.rs
```diff
@@ -16,7 +18,7 @@ pub trait FnOnceExt<Args: Tuple>: FnOnce<Args> {
 /// `FnMutExt` is an extension trait for the `FnMut` trait, used for
 /// adding a specification to closures. It should not be used directly.
 #[rustc_diagnostic_item = "fn_mut_spec"]
-pub trait FnMutExt<Args: Tuple>: FnMut<Args> + FnOnceExt<Args> {
```
xldenis, line 19: 🎉

creusot-contracts/src/std/ops.rs
```diff
@@ -5,7 +5,9 @@ pub use ::std::ops::*;
 /// `FnOnceExt` is an extension trait for the `FnOnce` trait, used for
 /// adding a specification to closures. It should not be used directly.
 #[rustc_diagnostic_item = "fn_once_spec"]
-pub trait FnOnceExt<Args: Tuple>: FnOnce<Args> {
+pub trait FnOnceExt<Args: Tuple> {
+    type Output;
```
xldenis, line 9: Interesting, you needed to do this?

jhjourdan: This creates an ambiguity in trait resolution, which the new solver does not like.

jhjourdan:
Oh, and by the way: this PR magically fixes our problem with `PartialEq::eq`.

xldenis:
> Oh, and by the way: this PR magically fixes our problem with PartialEq::eq.

Magical fixes are the best kinds of fixes

PR #1235, opened 2024-11-14
Remove buggy check that we do not instantiate Deref predicates with ghost

jhjourdan:
The test is bogus (in particular because it does not handle nested constraints), and is useless because Deref::deref has no specification (and hence has a false pre).

Cc @arnaudgolfouse **when you will be back from vacation**.


jhjourdan:
@arnaudgolfouse Could you please have a quick look into this?

jhjourdan:
Alright, thanks, merging!

PR #1236, opened 2024-11-14
Get rid of ClosureSpecKind.

jhjourdan:
Thanks to #1228, this can be handled in a much simpler way directly when elaborating these items.

The only remaining case is closure accessors.


jhjourdan:
>  should help with turning postcondition_mut from &mut F, Args -> bool to F, Args, F -> bool

Wish granted! Sadly, it seems like this opens some simplifications in map iterators proofs, but implementing these make the proofs much more difficult...

PR #1243, opened 2024-11-20
add weak dependencies

xldenis:
The catch all "AllTyInvAxiom" dependency causes problems when trying to add closure contract inference, by introducing false cycles which get rejected by the `clone_map`. 

To work around this, we introduce a notion of weak dependency. Weak dependencies are allowed to form cycles and are used to perform an initial stratification of the dependencies in the clone map. 

We first find all SCCs respecting weak edges, and then within each of these we perform a second topological sort to print the dependencies according to their *strong* edges. 

This check allows for much more fine-grained ordering in the presence of type invariants. @jhjourdan it should also allow you to address the issues around laws you were identifying. 

Several sessions broke in very weird ways given the no-op nature of the changes, particularly `01_resolve_unsoundness` and `878` which had goals go from sat -> unknown, however I wasn't able to obtain the sat even without my changes? is this a linux vs mac issue?


creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -167,8 +166,13 @@ impl DepElab for LogicElab {
...
         if is_inv_function(ctx.tcx, def_id) {
-            elab.expansion_queue
-                .push_back((Dependency::AllTyInvAxioms, Dependency::TyInvAxiom(subst.type_at(0))));
+            elab.expansion_queue.push_back((
+                dep,
+                Strength::Weak,
+                Dependency::TyInvAxiom(subst.type_at(0)),
+            ));
+            // elab.expansion_queue
+            //     .push_back((Dependency::AllTyInvAxioms, Dependency::TyInvAxiom(subst.type_at(0))));
```
jhjourdan, line 175:
```suggestion
            // elab.expansion_queue
            //     .push_back((Dependency::AllTyInvAxioms, Dependency::TyInvAxiom(subst.type_at(0))));
```

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -293,8 +297,8 @@ impl<'a, 'tcx> Expander<'a, 'tcx> {
             self_key,
             param_env,
             expansion_queue: initial
-                .chain(iter::once(Dependency::AllTyInvAxioms))
-                .map(|b| (self_key, b))
+                // .chain(iter::once(Dependency::AllTyInvAxioms))
```
jhjourdan, line 300:
```suggestion

```

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -359,14 +368,13 @@ impl<'a, 'tcx> Expander<'a, 'tcx> {
...
-        // Make every declaration that creates a goal depend on every TyInvAxiom to make sure that
-        // TyInvAxioms are visible at this point
-        if decls.iter().any(|d| matches!(d, Decl::Coma(..) | Decl::Goal(_))) {
-            self.expansion_queue.push_back((dep, Dependency::AllTyInvAxioms));
-        }
+        // // Make every declaration that creates a goal depend on every TyInvAxiom to make sure that
+        // // TyInvAxioms are visible at this point
+        // if decls.iter().any(|d| matches!(d, Decl::Coma(..) | Decl::Goal(_))) {
+        //     self.expansion_queue.push_back((dep, Dependency::AllTyInvAxioms));
+        // }
```
jhjourdan, line 377:
```suggestion
```

jhjourdan:
> @jhjourdan it should also allow you to address the issues around laws you were identifying.

Well, except that in the current state, it does not, because there is no weak dep from the triggering item to the law.

xldenis:
> Well, except that in the current state, it does not, because there is no weak dep from the triggering item to the law.

I didn't say that it _addresses_ but it _allows addressing_ :P

jhjourdan:
> I didn't say that it addresses but it allows addressing :P

Oh, sorry then! I'll add a patch.

> Something weird is going on with the CI. on my computer the generated coma is ok and the proofs pass.

Looking at it!

jhjourdan:
I fixed the sessions. It should work now. I think you did not used an up-to-date version of Why3.

PR #1244, opened 2024-11-20
Error instead of panic for unhandled matchee types

### LINKED ISSUE #482: Better error message when using `char`

Lysxia:
Closes #482 

Note that we already warn about uses of `char`. This is just to provide a cleaner error message for unhandled cases.


jhjourdan:
LGTM

PR #1253, opened 2024-11-26
Fix spec of `FMap::get_mut_ghost`

arnaudgolfouse:
It was too weak: in particular, you could not prove that the domain of the map stays the same after the call.


creusot-contracts/src/logic/fmap.rs
```diff
@@ -291,11 +291,12 @@ impl<K, V: ?Sized> FMap<K, V> {
     #[ensures(if self.contains(*key) {
             match result {
                 None => false,
-                Some(r) => *(*self).lookup_unsized(*key) == *r &&
+                Some(r) => (^self).contains(*key) &&
+                           *(*self).lookup_unsized(*key) == *r &&
                            *(^self).lookup_unsized(*key) == ^r,
```
xldenis, line 296: Shouldn't contains be implied by that?

arnaudgolfouse: I don't think so... First the following function
```rust
#[requires(map.contains(*key))]
#[ensures(forall<k: K> (*map).contains(k) == (^map).contains(k))]
pub fn foo<K, V>(map: &mut FMap<K, V>, key: &K) {
    let _ = map.get_mut_ghost(key);
}
```
cannot be proved automatically with the current spec.

But beyond that, `lookup_unsized` does not make the assumption that the key is present in the map, it just specify the behavior in the case it _is_ present.

xldenis: > But beyond that, lookup_unsized does not make the assumption that the key is present in the map, it just specify the behavior in the case it is present.

Right then your spec makes sense :+1:

PR #1254, opened 2024-11-26
Indexing of `FMap` in logic

arnaudgolfouse:
Allow indexing of a `FMap` in logic. It only works with `Sized` values though.

Also add `FMap::create` and `FSet::create` for consistency with `Seq`.


creusot-contracts/src/logic/fmap.rs
```diff
@@ -107,18 +111,19 @@ impl<K, V: ?Sized> FMap<K, V> {
         self.get_unsized(k) != None
     }
 
+    /// Create a new, empty map.
     #[trusted]
     #[logic]
     #[ensures(result.len() == 0)]
     #[ensures(result.view() == Mapping::cst(None))]
-    pub fn empty() -> Self {
+    pub fn create() -> Self {
```
xldenis, line 119: Since this is logic not ghost I'm not sure create is really the write name, all empty sets should be the same, there is no other notion of identity right?

xldenis: I would prefer that over create. `create` implies something _fresh_ to me so the fact that they are all the same unit element of their type would suggest a different name.

PR #1255, opened 2024-11-26
Run tests using why3find

Lysxia:
Sequel to #1205 

Run tests with `why3find` by default.

Some test cases must still be run with `why3` because they use ad hoc transformations, they are indicated by a new header `WHY3PROVE`.

There is one test case that requires `inline_goal` (`iterators/10_once.rs`), enabled with the header `TACTIC +inline_goal`.


Lysxia:
There is still a detail I need to figure out about how why3find records prover times. I asked about it on the why3find issue tracker: https://git.frama-c.com/pub/why3find/-/issues/117

jhjourdan:
Also, now that we are using Why3find, which is supposed to be more stable than why3 wrt. proof replay failure, would it be relevant to remove the "UNSTABLE" annotation of some tests? (Recall: this annotation tells CI to skip these tests, because they are too unstable.)

@xldenis ; anything to say?

xldenis:
I would love to remove that annotation if it does become more stable!

Lysxia:
The remaining tests with manual proofs (= those that need `why3session.xml`):

```
creusot/tests/should_succeed/sparse_array/why3session.xml
creusot/tests/should_succeed/iterators/07_fuse/why3session.xml
creusot/tests/should_succeed/iterators/05_map/why3session.xml
creusot/tests/should_succeed/iterators/04_skip/why3session.xml
creusot/tests/should_succeed/iterators/06_map_precond/why3session.xml
creusot/tests/should_succeed/iterators/17_filter/why3session.xml
creusot/tests/should_succeed/iterators/09_empty/why3session.xml
```

xldenis (approved):
Looks great to me, eventually we'll need to find a solution for the remaining tests but this is step in the good direction. It also gives us a simpler metric to target perf wise

Lysxia:
I just noticed that `--replay=none` is ignored by our why3find tests (so CI currently runs them twice)... I think the closest why3find equivalent is to pass `--replay` to why3find and store its cache in the Github Actions cache.

jhjourdan:
Indeed, this is great, and exceeds my expectations !

> The remaining tests with manual proofs (= those that need `why3session.xml`):
> 
> ```
> creusot/tests/should_succeed/sparse_array/why3session.xml
> ```

We should write a recursive proof in Creusot ourselves. This is probably not very difficult. It amounts to prove that an injective mapping (the `back` mapping) with a finite range which coincides with its domain is necessarily surjective.

This should be proved by induction by iterating the mapping. @Lysxia do you feel like you could give it a try?

> ```
> creusot/tests/should_succeed/iterators/07_fuse/why3session.xml
> creusot/tests/should_succeed/iterators/05_map/why3session.xml
> creusot/tests/should_succeed/iterators/04_skip/why3session.xml
> creusot/tests/should_succeed/iterators/06_map_precond/why3session.xml
> creusot/tests/should_succeed/iterators/17_filter/why3session.xml
> creusot/tests/should_succeed/iterators/09_empty/why3session.xml
> ```

I guess the solution for them is to redesign the `Iterator` trait by adding more snapshot parameters/ return values to avoid quantifiers at all prices.

We also discussed with @Lysxia a design that would allow manually introducing forall quantifiers in logic code.

jhjourdan:
Anyway, I thin kwe can merge this. This is awesome!

PR #1256, opened 2024-11-27
Ghost `int` literal

### LINKED ISSUE #1251: `Int` literals in ghost code
There should be an easy way to create an `Int` in ghost code. Currently you can do
```rust
ghost! {
    let x: Int = *Int::new(42);
};
```
But this is very inelegant.

arnaudgolfouse:
Create `Int`s in ghost code using a `int` suffix, like so:
```rust
ghost! {
    let x = 1int;
};
```

This is also added in pearlite for consistency.

Fixes #1251.


arnaudgolfouse:
I'm unhappy about the duplicated `ghost.rs` file in `creusot-contracts-proc` and `creusot-contracts-dummy`, but using a `#[path = "..."]` annotation does not work... If someone has a better idea I'm interested!

arnaudgolfouse:
> I'm unhappy about the duplicated `ghost.rs` file in `creusot-contracts-proc` and `creusot-contracts-dummy`, but using a `#[path = "..."]` annotation does not work... If someone has a better idea I'm interested!

Well apparently it _does_ work!

Issue #1265, opened 2024-11-28
Handle `#[test]`


Lysxia:
Good rust projects have test suites, and it would be nice to be able to check assertions in test suites using creusot.

`cargo creusot -- --tests` on a random project fails with:

```
thread 'rustc' panicked at creusot/src/backend/ty_inv.rs:69:18:
not implemented: dyn [Binder { value: Trait(std::ops::Fn<(&mut test::Bencher,)>), bound_vars: [Region(BrNamed(DefId(26:1712 ~ test[7e83]::types::TestFn::'_#3), '_))] }, Binder { value: Projection(Output = std::result::Result<(), std::string::String>), bound_vars: [Region(BrNamed(DefId(26:1712 ~ test[7e83]::types::TestFn::'_#3), '_))] }, Binder { value: AutoTrait(DefId(2:32809 ~ core[98a5]::marker::Send)), bound_vars: [] }] + 'static
```

Those `dyn` seem to come from [`test::TestFn`](https://doc.rust-lang.org/test/enum.TestFn.html).

Maybe we can just tell creusot to ignore the generated test wrappers when `--test` is on.

PR #1266, opened 2024-11-28
Add `new` command to create projects

Lysxia:
Set up a minimal project with some contracts.

- This uses the released `creusot-contracts`, which is out of date because it's missing the ghost primitives (I am manually testing this PR by replacing it with a `{ path = ... }` dependency).
- In the future I'd like to be able to check test suite with creusot (blocked by #1265 and also an anomaly in Why3 due to tuples generated by `assert_eq!`, reported at https://gitlab.inria.fr/why3/why3/-/issues/890)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -50,29 +50,10 @@ jobs:
         cp ci/creusot-config-dummy.toml ~/.config/creusot/Config.toml
     - name: Run tests
       run: cargo test
-  why3:
-    runs-on: ubuntu-22.04
+  why3-deps:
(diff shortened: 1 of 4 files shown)


arnaudgolfouse:
Maybe also a command `init` for consistency with cargo?

xldenis:
It's always a little more touchy but if even go so far as to propose a CI workflow since it may not be obvious.

Lysxia:
I beat CI into submission!

- Added `init`
- Added some common feature flags to build with rustc
- Added test for this new functionality in CI

PR #1269, opened 2024-11-29
Add why3find dependency to nightly

Lysxia:
D:

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -34,6 +34,5 @@ jobs:
     - name: run cargo creusot setup install
       if: steps.cache-creusot-setup.outputs.cache-hit != 'true'
       run: |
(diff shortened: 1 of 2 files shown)


Armael:
I think you also need to fix `nightly.yml` to pass `--no-check-version` for why3find

jhjourdan:
> I think you also need to fix `nightly.yml` to pass `--no-check-version` for why3find

Really? If the version does not match, then we need to know that, and update Why3find.

Armael:
Well the point of the nightly job is to see whether creusot works with the newest (possibly different) version of why3find. (We already pass the same option for why3 right now.)

jhjourdan:
I don't think the version check of why3find checks the individual commit.

Lysxia:
Nightly went green again

PR #1270, opened 2024-11-29
creusot-setup: Clarify error about whether a tool is missing or has the wrong version

Lysxia:


Armael (approved):
Looks like a clear improvement!

PR #1274, opened 2024-12-02
Fix formatting in cargo new templates

Lysxia:


Lysxia:
@xldenis  can you add the `install` job (maybe also `contracts-build`) to the required CI jobs for auto merge?

PR #1276, opened 2024-12-02
Use path to why3find stored in Creusot configuration

### LINKED ISSUE #1275: `cargo creusot config`: No such file or directory
`cargo creusot config` fails on a new project with
```
Error: why3find config failed

Caused by:
    No such file or directory (os error 2)
```
This seems to be because `why3find` is not available outside of creusot's directory. Since I installed why3find on a local switch inside that directory, it cannot find it elsewhere.

Looking at `cargo-creusot/src/why3find_wrapper.rs`, it seems this is also a problem for `cargo creusot prove`.

Lysxia:
Fix #1275

### DIFF
--- a/cargo-creusot/src/why3find_wrapper.rs
+++ b/cargo-creusot/src/why3find_wrapper.rs
@@ -1,6 +1,6 @@
 use anyhow::Result;
 use clap::*;
-use creusot_setup::{get_why3_config_file, PROVERS};
+use creusot_setup::{status_for_creusot, CreusotFlags, PROVERS};
 use std::{
     path::{Path, PathBuf},
     process::{Command, Stdio},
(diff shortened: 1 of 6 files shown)


ci/creusot-config-dummy.toml
```diff
@@ -1,4 +1,4 @@
-version = 5
+version = 6
```
Lysxia, line 1: I forgot to bump the version when I added why3find

cargo-creusot/src/why3find_wrapper.rs
```diff
@@ -27,8 +27,8 @@ fn why3find_json_exists() -> bool {
     Path::new("why3find.json").exists()
 }
 
-fn raw_config(args: &Vec<String>, why3_config_file: &PathBuf) -> Result<()> {
-    let mut why3find = Command::new("why3find");
+fn raw_config(args: &Vec<String>, why3find: &PathBuf, why3_config_file: &PathBuf) -> Result<()> {
```
arnaudgolfouse, line 30: Why not pass the full `&creusot_setup::CreusotFlags`, like in `raw_prove` ?

arnaudgolfouse (approved):

cargo-creusot/src/why3find_wrapper.rs
Lysxia: I think I'm going to do that for consistency. At first I was just not sure if that's the right way to use `CreusotFlags` and `status_for_creusot` because of their not very descriptive and probably outdated names.

PR #1277, opened 2024-12-03
Refactor handling of why3 paths in cargo-creusot and tests

Lysxia:
- Remove library compoenent of creusot-dev-config; test suite depends on creusot-setup instead
- Rename `status_for_creusot` and `CreusotFlags` to `creusot_paths` and `Paths`


Lysxia:
Oops. Fixing that in #1282.

PR #1278, opened 2024-12-03
Condense test logs by hiding successful tests

Lysxia:


xldenis:
this is nice, but we may also just want to have proper GitHub actions support which includes test reports with nice summaries etc..

Lysxia:
I hadn't thought about that! This is more for keeping our local terminals clean rather than Github Actions'.

Lysxia:
I had a look at [the various kinds of messages we can emit in Github Actions](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#adding-a-job-summary) but I don't see one that would make us really more productive than looking at the logs directly.

PR #1281, opened 2024-12-03
`Int` ghost methods

### LINKED ISSUE #1271: Operations on `Int` in ghost code
So right now it is not possible to perform arithmetic operation on `Int` in ghost code, e.g. the following:
```rust
ghost! {
    let x = 1int;
    let y = 2int;
    let z = x + y;
};
```
Fails, because `+` on `Int` is a logic-only operation.

arnaudgolfouse:
Allow using operators (`+`, `*`...) on `Int` in ghost code.

This works because it is never possible to get an `Int` in normal (non-ghost) code.

Fixes #1271.


arnaudgolfouse:
Right now this bumps into limitations of the purity check, so
```
#[pure]
fn add(x: Int, y: Int) -> Int {
    x + y
}
```
is rejected because the _trait function_ `Add::add` is not `#[pure]` (although `<Int as Add>::add` _is_)

jhjourdan:
> is rejected because the trait function Add::add is not #[pure] (although <Int as Add>::add is)

Huh, you mean that we don't resolve trait methods in the purity check? We should!

arnaudgolfouse:
Hum nevermind, I just forgot a `#[pure]` annotations 😅

creusot-contracts/src/logic/int.rs
```diff
@@ -164,68 +165,149 @@ impl Int {
...
+impl AddLogic for Int {
+    type Output = Self;
+    #[logic]
     #[trusted]
     #[creusot::no_translate]
     #[creusot::builtins = "add_int"]
-    fn add(self, _: Int) -> Self {
-        panic!()
+    fn add(self, other: Self) -> Self {
+        let _ = other;
```
jhjourdan, line 175:
```suggestion
    fn add(self, _: Self) -> Self {
```

arnaudgolfouse: ahhh, this is so it renders better in doc :)

jhjourdan: Well, then, you can use `#[allow(unused)]`.

creusot/src/validate.rs
```diff
@@ -127,8 +127,7 @@ pub(crate) fn validate_traits(ctx: &mut TranslationCtx) {
 fn is_overloaded_item(tcx: TyCtxt, def_id: DefId) -> bool {
     if let Some(name) = tcx.get_diagnostic_name(def_id) {
         match name.as_str() {
-            "mul" | "add" | "sub" | "div" | "rem" | "neg" | "box_new" | "deref_method"
-            | "deref_mut_method" => true,
+            "box_new" | "deref_method" | "deref_mut_method" => true,
```
jhjourdan, line 130: :tada:

creusot/tests/should_succeed/mc91.rs
```diff
@@ -3,7 +3,7 @@ extern crate creusot_contracts;
 use creusot_contracts::*;
 
 #[ensures(x <= 100u32 ==> result == 91u32 &&
-    x > 100u32 ==> result == x - 10u32)]
+    x > 100u32 ==> result@ == x - 10u32)]
```
jhjourdan, line 6: Why does this typecheck? `x` and 10u32 are machine integers, which should no longer have arithmetic operators.

jhjourdan:
Overall, LGTM.

@laurentder : this should fix your issues about machine arithmetics in pearlite specifications, since they are all rewritten.

creusot/tests/should_succeed/mc91.rs
arnaudgolfouse: I added an implementation `add(i32, i32) -> Int` (and for other integers too). I was trying to keep some of the existing behaviour, but to be honest I would prefer to remove them. If you agree I will do that (and then we would have to write `result@ == x@ - 10`)

PR #1287, opened 2024-12-04
cargo-creusot: Disable incremental compilation

### LINKED ISSUE #1286: Error when rebuilding code with closures
In a fresh project with only the following `lib.rs` (a pretext to use a closure),

```rust
#![cfg_attr(not(creusot), feature(stmt_expr_attributes, proc_macro_hygiene))]
use creusot_contracts::*;

pub fn add(v: Vec<u64>) -> u64 {
    v.iter().map(|x| x + 1).sum()
}
```

```
cargo creusot
touch src/lib.rs
cargo creusot
```

yields

```
thread 'rustc' panicked at creusot/src/ctx.rs:281:36:
did not find body for DefId(0:5 ~ clo[5629]::add::{closure#0})
```

Creusot [overrides the `mir_borrowck` query](https://github.com/creusot-rs/creusot/blob/7f18d4f42b1c039b6a2d9a8171e13d8ef8f5c8e3/creusot/src/callbacks.rs#L90) to collect function and closure bodies in a map, and it seems that that query is sometimes skipped when rebuilding code that has not changed.

Lysxia:
Closes #1286

### DIFF
--- a/cargo-creusot/src/main.rs
+++ b/cargo-creusot/src/main.rs
@@ -137,6 +137,9 @@ fn invoke_cargo(args: &CreusotArgs, cargo_flags: Vec<String>) {
         .args(cargo_flags)
         .env("RUSTC", creusot_rustc_path)
         .env("CARGO_CREUSOT", "1");
+    // Incremental compilation causes Creusot to not see all of a crate's code
+    // (the `mir_borrowck` hook in `creusot/src/callbacks.rs` is not called on all closures).
+    cmd.env("CARGO_INCREMENTAL", "0");
 
     // Append flags to any pre-existing ones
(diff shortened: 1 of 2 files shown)


xldenis (approved):

jhjourdan:
This looks more like a hack than a proper solution... (I don't know what would be a proper solution, though.)

Lysxia:
I think the hack is the whole process of getting MIR out of rustc, and a proper solution would involve rustc having an actual API for alternative backends.

jhjourdan:
I don't really understand what this overriding mechanism does, but I am surprise that it is impossible to make it work together with caching.

PR #1289, opened 2024-12-04
Clone ghost containers

arnaudgolfouse:
Implement `Clone` on `Seq`, `FMap` and `FSet`.


jhjourdan:
What's the motivation? Is there any ghost object which is clone but not copy ?

arnaudgolfouse:
None really, bu we need to implement `Clone` to be `Copy` anyways.

PR #1290, opened 2024-12-04
creusot-contracts: Add Iterator impls for [T;N], &mut I

Lysxia:
These came up in examples I've been playing with.


arnaudgolfouse:
Could you add some test cases for those?

creusot-contracts/src/std/iter.rs
```diff
@@ -211,3 +213,33 @@ extern_spec! {
...
+    #[open]
+    #[predicate(prophetic)]
+    fn produces(self, visited: Seq<Self::Item>, o: Self) -> bool {
+        (*self).produces(visited, *o)
+    }
+
+    #[open]
+    #[predicate(prophetic)]
+    fn completed(&mut self) -> bool {
+        (**self).completed()
```
jhjourdan, line 227:
```suggestion
        (*self).completed() && ^*self == ^^self
```

jhjourdan: (I think, if I'm not mistaken.)

creusot-contracts/src/std/iter/array.rs
```diff
@@ -0,0 +1,48 @@
...
+    fn into_iter_pre(self) -> bool {
+        pearlite! { true }
+    }
+
+    #[trusted]
+    #[predicate(prophetic)]
+    fn into_iter_post(self, _res: Self::IntoIter) -> bool {
+        dead
+    }
+}
```
jhjourdan, line 48: Could you please give actual specification which state something usable? As is, I don't see what these specifications allow proving. You could take inspiration from iterators of vectors, for example.

Also, I am a bit surprised that these specifications are accepted by Creusot. Usually, we have very limited support for const generics in Creusot.

creusot-contracts/src/std/iter/str.rs
```diff
@@ -0,0 +1,31 @@
...
+    #[trusted]
+    #[law]
+    #[requires(inv(a))]
+    #[requires(inv(b))]
+    #[requires(inv(c))]
+    #[requires(a.produces(ab, b))]
+    #[requires(b.produces(bc, c))]
+    #[ensures(a.produces(ab.concat(bc), c))]
+    fn produces_trans(a: Self, ab: Seq<Self::Item>, b: Self, bc: Seq<Self::Item>, c: Self) {}
+}
```
jhjourdan, line 31: Same here. This specification says nothing.

Lysxia:
I removed `Lines` from this PR because strings are a pain. I added specs for the remaining impls and basic test examples.

PR #1295, opened 2024-12-09
`PtrOwn` disjoint lemma

arnaudgolfouse:


xldenis:
Why is this not enforced by the model of the type system itself? I feel like we shouldn't need to state / prove this property no?

jhjourdan:
This is a guarantee of the modelof the type system indeed, but without this ghost "lemma", there is no way to leverage this in Creusot.

PR #1303, opened 2024-12-11
Clean verif/ directory before writing to it

Lysxia:
When you remove a function in your Rust code, the corresponding Coma file(s) should be removed.


arnaudgolfouse:
Hm, this is not enough unfortunately: removed files are still listed in the `why3session.xml`, and they cause a crash in why3...
Granted, why3find handles it without issues ;) (But I need the ide and its tactics)

Lysxia:
I think we want to fix why3 here.

Lysxia:
This should resolve the crash https://gitlab.inria.fr/why3/why3/-/merge_requests/1189

jhjourdan:
> This should resolve the crash https://gitlab.inria.fr/why3/why3/-/merge_requests/1189

Is this really what we want? Who is going to remove the session file in this case?

Lysxia:
My Why3 MR addresses the current/old use case of having one big `why3session.xml` at the root of the project for all `.coma` files, by fixing the crash when a file is removed and instead treating its session as detached.

Since then, we've also decided to switch to a workflow where each `.coma` file has its own `why3session.xml` or `proof.json`, and this workflow does not run into the crash above. As far as I can tell, dangling `why3session.xml` files are harmless. Removing them as soon as we remove their `.coma` file seems dangerous (the user could change their mind and re-add the `.coma` file, or they may want to move the session/proof file rather than delete it). Instead I propose to add a new command `cargo creusot clean` to remove dangling sessions and proofs. `cargo creusot` could suggest this command in a warning.

jhjourdan:
Sounds good indeed!

jhjourdan:
Ok, this looks good to me. Merging as soon as CI passes.

PR #1316, opened 2025-01-16
Union-Find example

arnaudgolfouse:
Add an example of a union-find data structure, using ghost code. Adapted from the why3 proof at <https://github.com/ocaml-gospel/vocal/blob/main/proofs/why3/UnionFind_impl.mlw>.

Note that it uses the `compute_in_goal` tactic in order to unwrap the definition of the invariant. Fortunately one can set it in the `why3find.json` file when using why3find, and everything passes automatically.


jhjourdan:
LGTM, go.

PR #1326, opened 2025-01-27
ci: Fix deploy.yml

Lysxia:


Lysxia:
Is there a way to test this or do we just merge it?

jhjourdan:
I'm fine with merging this, but why do we need this change ?

Lysxia:
The [deploy build is broken](https://github.com/creusot-rs/creusot/actions/runs/12946442942/job/36110984703) since I changed the installation steps in #1314 and forgot to update `deploy.yml`.

jhjourdan:
Too bad, this still does not work!

PR #1329, opened 2025-01-27
Add command `cargo creusot clean`

Lysxia:
As proposed in [#1303](https://github.com/creusot-rs/creusot/pull/1303#issuecomment-2615269970)


jhjourdan:
There is no documentation about this new command. Perhaps we could have some central document with the description of all the available `cargo creusot` commands?

Lysxia:
Good idea. I've added a page about `cargo creusot` commands.

jhjourdan:
Ok, thanks!

PR #1331, opened 2025-01-28
Contracts build stable

Lysxia:


xldenis (approved):
Step is only needed for extern specs right? if so then its fine to just make it conditional.

jhjourdan:
Do we agree that this does not really solves the problem that `#[invariant(..)]` does not work on stable?

Sure, the user can write `#[cfg_attr(creusot, invariant(...))]`, but this is both ugly and maybe it should be rejected by Rust, because this is still an attribute on a statement...

xldenis:
yes, agreed. but this is still an improvement.

Lysxia:
> Do we agree that this does not really solves the problem that #[invariant(..)] does not work on stable?

Yes, I'm working on that, and this PR does not need it.

creusot-contracts/src/std/iter/range.rs
```diff
@@ -50,6 +50,7 @@ pub fn range_inclusive_len<Idx: DeepModel<DeepModelTy = Int>>(r: RangeInclusive<
     }
 }
 
+#[cfg(creusot)]
 impl<Idx: DeepModel<DeepModelTy = Int> + Step> Iterator for RangeInclusive<Idx> {
```
Lysxia, line 54: This is where `Step` is used. This PR currently just removes these `impl`s. Is that okay or is there a possible replacement?

creusot-contracts/src/std/ops.rs
```diff
@@ -254,3 +261,27 @@ extern_spec! {
...
+#[cfg(not(creusot))]
+impl<A, O, F: FnOnce(A) -> O> FnOnceExt<(A,)> for F {
+    type Output = O;
+}
+
+#[cfg(not(creusot))]
+impl<A, O, F: FnMut(A) -> O> FnMutExt<(A,)> for F {}
+
+#[cfg(not(creusot))]
+impl<A, O, F: Fn(A) -> O> FnExt<(A,)> for F {}
```
Lysxia, line 287: TODO: finish this hack or find another replacement (this change is because the `Tuple` and `Fn<_>` traits are unstable).

Lysxia:
I added the handling of `invariant` here because testing it depends on this PR anyway, even though the actual changes orthogonal.

creusot-contracts-dummy/src/lib.rs
```diff
@@ -106,3 +112,36 @@ pub fn derive_deep_model(_: TS1) -> TS1 {
...
+    fn visit_expr_loop_mut(&mut self, i: &mut syn::ExprLoop) {
+        delete_invariants_attrs(&mut i.attrs);
+        syn::visit_mut::visit_expr_loop_mut(self, i)
+    }
+}
+
+fn delete_invariants_attrs(attrs: &mut Vec<syn::Attribute>) {
+    attrs.retain(|attr| {
+        if let syn::Meta::List(meta) = &attr.meta {
+            !meta.path.is_ident("invariant")
```
Lysxia, line 138: should we handle `#[creusot_contracts::invariant(...)]` too?

creusot-contracts/src/std/iter/range.rs
xldenis: Yes you can remove them since that's purely for specifications.

jhjourdan: Yeah, the trait and the implementations should remain in the non-creusot config.

Isn't there a way to force Rustc to accept unstable features in a crate used by a stable Rust compiler? That would be unfair, Rust's stdlib is using plenty of unstable features to target stable clients, but we couldn't!

creusot-contracts/src/std/ops.rs
jhjourdan: We should keep them for the same reason!

creusot-contracts-dummy/src/lib.rs
jhjourdan: I guess? Is it significantly more difficult?

Lysxia: no, I just wanted to be sure.

creusot-contracts/src/std/iter/range.rs
Lysxia: As far as I can tell, rustc provides special attributes for std to mark APIs as stable or unstable, and to allow itself to use its own unstable APIs in stable code. That's how it's able to expose an `impl` that depends on `Step` (or `Allocator` or some other unstable trait). Those attributes can't be used outside of std.

Lysxia:
I changed the flag from `creusot` to `feature = "nightly"` so that we can still use the original impls when building with a nightly rustc. The hacks to approximate those impls are only needed for stable rustc.

Lysxia:
Ready for review.

PR #1333, opened 2025-01-31
Add a derive macro for `Default`

arnaudgolfouse:
## Example

```rust
#[derive(Default)]
struct S {
    x: i32, y: i32
}
```

Generates

```rust
struct S {
    x: i32, y: i32
}

impl ::std::default::Default for S {
    #[ensures(result.is_default())]
    fn default() -> Self {
        Self { x: <i32>::default(), y: <i32>::default() }
    }
}

impl ::creusot_contracts::std::default::Default for S {
    #[predicate(prophetic)] #[open]
    fn is_default(self) -> bool {
        self.x.is_default() && self.y.is_default()
    }
}
```


creusot-contracts/src/std/default.rs
```diff
@@ -1,6 +1,24 @@
 use crate::*;
-pub use ::std::default::*;
 
+#[cfg(creusot)]
+pub use creusot_contracts_proc::Default;
+
+/// Extension of the standart [`Default`](::std::default::Default) trait.
```
Lysxia, line 6:
```suggestion
/// Extension of the standard [`Default`](::std::default::Default) trait.
```

Lysxia (approved):

creusot-contracts/src/std/default.rs
arnaudgolfouse: Eheh, thanks

PR #1334, opened 2025-01-31
Bump rustc.

jhjourdan:


jhjourdan:
@xldenis @Lysxia any idea why spans behave differently on CI than on my machine?

jhjourdan:
I guess there is some kind of configuration somewhere which enables more spans in rustc's ASTs?

PR #1335, opened 2025-02-01
creusot-setup: Print issues on stderr instead of stdout when bailing

Lysxia:
@xldenis that would provide a more informative error message in https://github.com/creusot-rs/creusot/pull/1294#issuecomment-2628449229


xldenis:
```
Testing creusot-contracts ... failure
Error: Why3 has version 1.7.2, expected version is 1.8.0
Error: why3find has version 1.0.0, expected version is 1.1.1
Error: Please run 'cargo creusot setup status' to diagnostic and fix the issue(s)
thread 'main' panicked at creusot/tests/creusot-contracts.rs:120:9:
output of creusot-contracts failed
```
ah! opam strikes again.

PR #1346, opened 2025-02-05
Logical functions aliases

dianegolfouse:
# Example

```rust
// In `creusot_contracts`

impl<K, V> FMap<K, V> {
    #[trusted]
    #[logic]
    #[ensures(result >= 0)]
    pub fn len_logic(self) -> Int {
        dead
    }

    #[trusted]
    #[pure]
    #[has_logical_alias(Self::len_logic)]
    pub fn len(self) -> Int {
        panic!()
    }
}

// User code

pub fn foo() {
    let mut m = FMap::new();
    proof_assert!(m.len() == 0); // logic call
    ghost! {
        m.insert_ghost(1int, 1int);
        let l = m.len(); // program (ghost) call
        proof_assert!(l == 1);
        m.insert_ghost(2int, 2int);
        proof_assert!(m.len() == 2);
    };
}
```

### DIFF
--- a/creusot-contracts-proc/src/creusot/specs.rs
+++ b/creusot-contracts-proc/src/creusot/specs.rs
@@ -72,6 +72,10 @@ pub fn requires(attr: TS1, tokens: TS1) -> TS1 {
 }
 
 pub fn ensures(attr: TS1, tokens: TS1) -> TS1 {
+    ensures_inner(attr, tokens, false)
+}
+
+fn ensures_inner(attr: TS1, tokens: TS1, logical_alias_path: bool) -> TS1 {
     let documentation = document_spec("ensures", doc::LogicBody::Some(attr.clone()));
 
     let mut item = parse_macro_input!(tokens as ContractSubject);
@@ -80,6 +84,12 @@ pub fn ensures(attr: TS1, tokens: TS1) -> TS1 {
 
     let ens_name = crate::creusot::generate_unique_ident(&item.name(), Span::call_site());
     let name_tag = ens_name.to_string();
+    let logical_alias = if logical_alias_path {
+        quote!(#[creusot::decl::logical_alias_path = #name_tag])
+    } else {
+        quote!()
+    };
+
     match item {
         ContractSubject::FnOrMethod(mut s) if s.is_trait_signature() => {
             let attrs = std::mem::take(&mut s.attrs);
@@ -96,6 +106,7 @@ pub fn ensures(attr: TS1, tokens: TS1) -> TS1 {
             TS1::from(quote! {
               #ensures_tokens
               #[creusot::clause::ensures=#name_tag]
+              #logical_alias
               #(#attrs)*
               #documentation
               #s
@@ -113,6 +124,7 @@ pub fn ensures(attr: TS1, tokens: TS1) -> TS1 {
             }
             TS1::from(quote! {
                 #[creusot::clause::ensures=#name_tag]
+                #logical_alias
                 #(#attrs)*
                 #documentation
                 #f
@@ -129,6 +141,7 @@ pub fn ensures(attr: TS1, tokens: TS1) -> TS1 {
                 #ensures_tokens
                 #res_id
             });
+            // FIXME: forbid logical aliases in this case
             TS1::from(quote! {
                 #[creusot::clause::ensures=#name_tag]
                 #clos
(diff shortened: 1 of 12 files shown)


creusot-contracts-proc/src/creusot/specs.rs
```diff
@@ -224,6 +237,29 @@ pub fn check(args: TS1, tokens: TS1) -> TS1 {
...
 
+pub fn has_logical_alias(attr: TS1, tokens: TS1) -> TS1 {
+    let logic_path = match syn::parse::<Path>(attr.clone()) {
+        Ok(path) => path,
+        Err(err) => {
+            return syn::Error::new(err.span(), "`has_logical_alias` should contain a path to a logical function with the same signature").to_compile_error().into()
+        }
+    };
+    let tokens_2 = tokens.clone();
+    let func = parse_macro_input!(tokens_2 as syn::ImplItemFn);
```
jhjourdan, line 248: This means that you force the method to have a body, which forbids logical aliases for trait methods. This seems restrictive. Is this intended? A simple fix may be to use `FnOrMethod` here instead.

creusot-contracts-proc/src/creusot/specs.rs
```diff
@@ -224,6 +237,29 @@ pub fn check(args: TS1, tokens: TS1) -> TS1 {
...
+    let args = func.sig.inputs.iter().map(|a| match a {
+        syn::FnArg::Receiver(receiver) => {
+            let self_t = receiver.self_token;
+            quote!(#self_t)
+        }
+        syn::FnArg::Typed(pat_type) => {
+            let pat = &pat_type.pat;
+            quote!(#pat)
```
jhjourdan, line 256: This assumes that the pattern is also a term, which is not always the case.

This is probably fine in this case, but it would be better to detect this in the macro, and emit an error.

creusot/src/ctx.rs
```diff
@@ -318,6 +322,168 @@ impl<'tcx> TranslationCtx<'tcx> {
...
+    /// The returned span is the span of the attribute.
+    pub(crate) fn logical_alias(&self, def_id: DefId) -> Option<(Span, DefId)> {
+        self.logical_aliases.get(&def_id).copied()
+    }
+
+    /// Load all the logical aliases defined in the current crate
+    ///
+    /// They can later be retrieved using [`Self::logical_alias`].
+    pub(crate) fn load_logical_aliases(&mut self) {
+        // FIXME: what about functions from another crate?
```
jhjourdan, line 338: Cannot we simply reject them?

creusot/src/ctx.rs
```diff
@@ -318,6 +322,168 @@ impl<'tcx> TranslationCtx<'tcx> {
...
+                                }
+                                _ => unreachable!(),
+                            };
+                            let (t2, name2) = match c2.kind() {
+                                rustc_type_ir::ConstKind::Param(p) => {
+                                    (p.find_const_ty_from_env(tcx.param_env(logic_id)), p.name)
+                                }
+                                _ => unreachable!(),
+                            };
+                            if t1 != t2 {
```
jhjourdan, line 392: Why don't we check the equality of constant param names, too? It seems weird to do that for types and lifetimes, but not for constants.

creusot/src/ctx.rs
```diff
@@ -318,6 +322,168 @@ impl<'tcx> TranslationCtx<'tcx> {
...
+        for local_id in self.hir_body_owners() {
+            let def_id = local_id.to_def_id();
+            // TODO: Put this in another file for better organisation
+            if let Some((span, logic_id)) = function_has_logical_alias(self, def_id) {
+                let program_span = tcx.def_span(def_id);
+                let logic_span = tcx.def_span(logic_id);
+                let program_subst = erased_identity_for_item(tcx, def_id);
+                let logic_subst = erased_identity_for_item(tcx, logic_id);
+                let mut program_subst = program_subst.iter().map(|arg| arg.kind());
+                let mut logic_subst = logic_subst.iter().map(|arg| arg.kind());
```
jhjourdan, line 349: Doing that directly in the loop would be clearer and shorter.

creusot/src/ctx.rs
```diff
@@ -318,6 +322,168 @@ impl<'tcx> TranslationCtx<'tcx> {
...
+                loop {
```
jhjourdan, line 350: From itertools:

```rust
for (prog_arg, logic_arg) in program_subst.zip_longest(logic_subst) { ... }
```

jhjourdan (commented):
This is a very interesting feature, howeer, there are a few important parts missing:

- Support for aliases in traits. A simple constraint could be to force the alias to belong to the same trait. Then, we sould add a refinement check for implementations (i.e., implementors declare compatible aliases).

- Uses in creusot_contracts, to get rid of all the `_ghost` and `_logic` functions.

- Use this for `Ghost::Deref`, `Rc::Deref`, `Snapshot::Deref`, etc... and get rid of the special cases for this in Creusot's code.

jhjourdan:
Perhaps we could ask Maël Coulmance to work on this?

xldenis:
What is the current thinking on this? I think this would be really nice and would really contribute to making proofs & specs easier to follow by getting rid of ubioquitous "_logic" suffixes, especially if there's a macro to help you automatically generate these aliases.

jhjourdan:
Superseeded by #2218.

PR #1351, opened 2025-02-08
Add basic support for RPIT

xldenis:
A proof I was working on was using `impl Iterator` and I got annoyed at us not supporting it. Return position impl trait is a lot simpler than the argument position since we can just compile it to an opaque type outside the defining scope (now that we "monomorphize" everything).


creusot/src/translation/function/terminator.rs
```diff
@@ -436,7 +438,7 @@ fn resolve_function<'tcx>(
 
 // Try to extract a function defid from an operand
 fn func_defid<'tcx>(op: &Operand<'tcx>) -> Option<(DefId, GenericArgsRef<'tcx>)> {
-    let fun_ty = op.constant().unwrap().const_.ty();
+    let fun_ty = op.constant()?.const_.ty();
```
xldenis, line 441: this just allows us to report a nicer error rather than panic.

creusot/src/backend/dependency.rs
```diff
@@ -57,9 +58,13 @@ impl<'tcx> Dependency<'tcx> {
...
                 }
-                TyKind::Alias(_, aty) => {
-                    Some(Symbol::intern(&type_name(tcx.item_name(aty.def_id).as_str())))
-                }
+                TyKind::Alias(AliasTyKind::Opaque, aty) => Some(Symbol::intern(&format!(
+                    "opaque{}",
+                    tcx.def_path(aty.def_id).data.last().unwrap().disambiguator
+                ))),
+                TyKind::Alias(_, aty) => Some(Symbol::intern(&type_name(
+                    tcx.opt_item_name(aty.def_id).unwrap_or(Symbol::intern("synthetic")).as_str(),
```
xldenis, line 66: I'm not sure _how_ but I'm getting these new "synthetic" types in the code I'm looking at which I don't yet know how to handle cleanly here.

jhjourdan:
Thanks, Xavier, I'll have a look into that.

In particular I'd like to better understand the choices about the type checker mode. Ot does not seem like that post analysis mode is the right mode...

Also, I don't see why these types in parameter position are more difficult to handle : cannot we handle them as abstract types in the body, and ask the type checker what they at the call site ?

xldenis:
>Also, I don't see why these types in parameter position are more difficult to handle : cannot we handle them as abstract types in the body, and ask the type checker what they at the call site ?

It's not _fundamentally_ more difficult but Rust compiles them to additional generic parameters literally called `"impl TraitWhatever"` with the expected bounds. It's just more plumbing to make it work properly that's all. For my purposes I only had a few `impl Iterators` returned by functions so I didn't care about that for the moment.

xldenis:
>In particular I'd like to better understand the choices about the type checker mode. Ot does not seem like that post analysis mode is the right mode...

I agree but normalization panics when in `analysis_in_body`, I couldn't find any clear explanation for why that would happen. Perhaps it has to do with the fact that we recover MIR after borrowck?

xldenis:
I forgot I actually did have some argument position impl traits: which generate invalid coma:

```
type impl Document'0
```

xldenis:
hmm I may just have added support for APIT by replacing spaces with `_` in the argument name.

jhjourdan:
Please try to avoid name clashes...

xldenis:
The whole thing goes through the clone map after so it should get a unique count.

xldenis:
Specifying functions with RPIT is tricky because our encoding of ensures won't work, specifically we create a closure with `result : ResultTy` but you can't have `impl Trait` arguments to closures. I think this requires changing the generation of `ensures` to look the one we give to closures, but that's not without its own complications!

creusot/src/backend/dependency.rs
```diff
@@ -40,6 +40,7 @@ impl<'tcx> Dependency<'tcx> {
                 TyKind::Adt(def, substs) => Some((def.did(), substs)),
                 TyKind::Closure(id, substs) => Some((*id, substs)),
                 TyKind::Alias(AliasTyKind::Projection, aty) => Some((aty.def_id, aty.args)),
+                TyKind::Alias(AliasTyKind::Opaque, aty) => Some((aty.def_id, aty.args)),
```
jhjourdan, line 43: I guess you could factorize too?

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -252,6 +250,13 @@ impl DepElab for TyElab {
                 ty_name: names.ty_param(ty).as_ident(),
                 ty_params: vec![],
             })],
+            TyKind::Alias(AliasTyKind::Opaque, _) => {
+                let (def_id, subst) = dep.did().unwrap();
+                vec![Decl::TyDecl(TyDecl::Opaque {
+                    ty_name: names.ty(def_id, subst).as_ident(),
+                    ty_params: vec![],
+                })]
+            }
```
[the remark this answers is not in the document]
jhjourdan: I guess we could also replace the assertion with a pattern checking that the alias kind is either opaque or a projection?

creusot/src/backend/dependency.rs
jhjourdan: I guess we could also replace the assertion with a pattern checking that the alias kind is either opaque or a projection?

creusot/tests/should_succeed/syntax/15_impl_trait.rs
```diff
@@ -0,0 +1,24 @@
...
+
+pub fn returns_iterator() -> impl MyTrait {
+    ()
+}
+
+#[ensures(true)]
+pub fn main() {
+    let x = returns_iterator().a();
+
+    proof_assert!(x);
```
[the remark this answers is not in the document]
jhjourdan: We should make sure that `should_succeed` only contain tests whose proof succeed. So yes, please move this to `should_fail`.

Also, except if there is a problem I don't see, could you please add a `#[ensures(result = true)]` to the impl above, just to make sure the prover indeed checks something?

creusot/src/backend/dependency.rs
jhjourdan: Do you have a repro? If possible, I'd like to have it in our tests.

jhjourdan:
@xldenis : I did the changes I wanted, I think this is ready to merge. Feel free to hit the merge button if you don't have anything to add.

PR #1354, opened 2025-02-11
test why3: add `--no-cache` and `// TIME (\d+)` options

Lysxia:


why3tests/tests/why3.rs
```diff
@@ -213,8 +217,13 @@ fn main() {
             why3find.env("WHY3CONFIG", &paths.why3_config);
             why3find.arg("prove").arg(file);
             if let Some(tactic) = tactic_re.captures_iter(&header_line).next() {
-                why3find.arg("--tactic");
-                why3find.arg(tactic.get(1).unwrap().as_str());
+                why3find.args(["--tactic", tactic.get(1).unwrap().as_str()]);
+            }
+            if let Some(time) = time_re.captures_iter(&header_line).next() {
```
xldenis, line 222: is it possible to use steps instead of time?

Lysxia: Not in why3find. :/

I'm still unclear on the situation with steps. In principle it seems like what we want, but jh mentioned that sometimes they were nondeterministic, that may be a Why3 bug. Do you recall any more details from experimenting with steps?

xldenis: yea I just recall that why3 would fail when given a stepcount (ie the same prover on the same machine on the same goal uses differing steps). However I also know AdaCore uses this religiously...

jhjourdan:
LGTM, merging

PR #1357, opened 2025-02-14
Bitwise operators

### LINKED ISSUE #743: Support for bitwise operations

### LINKED ISSUE #1124: Implement casts in pearlite

jhjourdan:
The goal is to fix #743 by giving a meaning to bitwise operations.

In order to be able to do proofs efficiently, we added the `#[bitwise_proof]` attributes, which makes Creusot interpret machine integers using bit vectors instead of bounded integers. Bitwise operators have an abstract specification when this attribute is not on.

This will also add support for arithmetic operations and casts in pearlite for machine integers, with a wrapping semantics.

Fixes #743.
Fixes #1124.


jhjourdan:
Merging !! :fireworks: :tada: :partying_face:

PR #1367, opened 2025-02-18
Ghost macro inline

arnaudgolfouse:
Make the ghost macro into an actual block, rather than a hacked closure. This should in particular help when borrowing part of a struct (Creusot will complain less), and when using types invariants (more proofs will pass).


jhjourdan:
So what's the status of this?

arnaudgolfouse:
I think this is good to go

PR #1368, opened 2025-02-19
Check that we do not reference uninit variables in logic

arnaudgolfouse:


creusot/tests/should_fail/final_borrows.rs
```diff
@@ -11,18 +11,20 @@ pub fn not_final_borrow<T: Copy>(bor: &mut T) {
 
 pub fn store_changes_prophecy<T: Copy>(bor: &mut T) {
     let b1 = &mut *bor;
+    let b1_snap = snapshot!(b1);
     let val = *b1;
```
jhjourdan, line 15: Wow, so this use of `b1` *uninitializes* it?

creusot/tests/should_fail/final_borrows.rs
```diff
@@ -11,18 +11,20 @@ pub fn not_final_borrow<T: Copy>(bor: &mut T) {
...
 }
 
 pub fn store_changes_prophecy_2<T: Copy>(bor: &mut T, x: T) {
     let b1 = &mut *bor;
     *b1 = x;
+    let b1_snap = snapshot!(b1);
     // The prophecy of `bor` changed here after the reborrow
     *bor = x;
-    proof_assert!(b1 == bor);
+    proof_assert!(*b1_snap == bor);
```
jhjourdan, line 27: *That* is weird. I guess you added the snap in order to compensate the fact that `b1` gets uninitialized here. But What is the instruction that does so?

creusot/tests/should_fail/final_borrows.rs
```diff
@@ -56,6 +58,8 @@ pub fn move_place_without_deref<T>(bor: Box<&mut T>) {
     #[ensures(**x == ^*x)]
     fn inner<T>(x: Box<&mut T>) {}
     let b1 = &mut **bor;
+    let b1_snap = snapshot!(b1);
+    let bor_snap = snapshot!(bor);
     inner(bor);
-    proof_assert!(b1 == *bor);
+    proof_assert!(*b1_snap == **bor_snap);
```
jhjourdan, line 64: Same here.

creusot/tests/should_fail/final_borrows.rs
jhjourdan: Ok, now I tested and I see that removing the snap does not change anything. Why did you add it?

arnaudgolfouse: Does it ?? I think I got confused then

creusot/tests/should_fail/final_borrows.rs
arnaudgolfouse: Ok, I was _really_ not thinking... none of those are actually uninitialized, except for `bor` in `move_place_without_deref` (which makes sense). I'm fixing this now.

PR #1370, opened 2025-02-20
Rename Why3find package "prelude" to "creusot"

### LINKED ISSUE #1369: Rename `prelude` to `creusot`?
"prelude" is a very generic name for what is now a proper why3find package. A very green newcomer might think it's some kind of standard library rather than something specific to this project.

Lysxia:
Close #1369


jhjourdan:
Thanks:!

PR #1371, opened 2025-02-20
Move prelude build script into creusot-install

Lysxia:


jhjourdan:
I'm not so sure about this one. What sense does it makes to make the Why3 test depend on creusot-install?

Lysxia:
`creusot-setup` itself doesn't need the prelude. It happens that the two places that need the prelude (installing Creusot and the Why3 tests) depend on `creusot-setup`, but that's more of a coincidence.

This spurious dependency via `build.rs` also means that a mere `cargo check` uselessly builds the prelude.

That's why `build.rs` is a strange place to build the prelude. Moreover, `creusot-install` is a good place to consolidate knowledge of how to build things that `cargo build` isn't suited to handle. If we think of `creusot-install` as our custom build system, rather than just an installer, then it makes sense for the Why3 tests to ask it to build the prelude.

PR #1372, opened 2025-02-20
Always run all the 'ui' tests

arnaudgolfouse:


jhjourdan:
Good!

While you are at it, could you please add one feature: don't run the creusot-contracts tests if the string "creusot-contracts" does not match the pattern?

jhjourdan:
Thanks!

PR #1374, opened 2025-02-21
ci: Reduce size of .cargo cache

Lysxia:


Lysxia:
On my machine `.cargo/registry/src` has grown to 1Gb but dropping it doesn't seem to make a difference in CI...

PR #1378, opened 2025-02-24
Remove spurious cargo-creusot dependency on why3

Lysxia:


jhjourdan:
@Armael I think you set up this. Do you confirm this is spurious ?

PR #1381, opened 2025-02-25
Simplify injections in prelude, and move to Rust edition 2024

### LINKED ISSUE #1380: char doesn't implement View
I would expect:

```rust
impl View for char {
    type ViewTy = Int;
}
```

That's probably an oversight.

jhjourdan:
See commits.

Fixes #1380.


ia0:
This would be fixing #1380 right? I can see `ord_logic_impl!(char, "prelude.prelude.Char");` in the "Simplify prelude and generated coma code" commit. However if I understand correctly it's not modeled as `Int` but `Char` and I must use `to_int` to get the Unicode scalar value.

jhjourdan:
> This would be fixing #1380 right? I can see `ord_logic_impl!(char, "prelude.prelude.Char");` in the "Simplify prelude and generated coma code" commit. However if I understand correctly it's not modeled as `Int` but `Char` and I must use `to_int` to get the Unicode scalar value.

Indeed.

Issue #667, opened 2022-12-18
Support for bit-level operations


ia0:
Hi,

I wanted to try proving the correctness of [`data-encoding`](https://crates.io/crates/data-encoding) but I hit the following error:

```
error[creusot]: unsupported binary operation: Shl
```

Is support for bit-level operations planned? I could not find an existing issue about it. Would be happy to subscribe to such issue and give it another try once there's support.

Thanks!

xldenis:
> Is support for bit-level operations planned? 

Yes, but with no firm date yet. It actually is already possible to some degree by swapping out the relevant modules in the `prelude`. 

> Would be happy to subscribe to such issue and give it another try once there's support.

Let's consider this the issue then!

jhjourdan:
This is a duplicate of #743.

ia0:
It's more of the opposite :-) but I'll subscribe to the other one then. Thanks for the pointer and looking into it!

jhjourdan:
Well, the other one has a longer discussion.

PR #1382, opened 2025-02-25
add get_unchecked_*

xldenis:
cc @ia0


xldenis:
@Lysxia @arnaudgolfouse can one of you update the creusot contracts coma? I can't do from my machines since it will change the `BitMask` type which varies based on architecture!

ia0:
Thanks! I confirm it works in my project (so far[^1]).

One thing that surprises me is `#[ensures((^self)@.len() == self@.len())]`. In Rust this comes for free because you can't modify the length of a `&mut [T]` parameter. I guess we need it in Creusot because the model for `&mut T` doesn't enforce that the metadata is preserved between the current value and the prophecy value. Is it because the metadata is not modeled yet? Or there's a more fundamental reason?

[^1]: I'm still at the very beginning (unblocked by #667) and expect quite a few issues.

xldenis:
Yea the model of mutable borrows of slices loses the invariant that the length can't change

Issue #1361, opened 2025-02-17
Install script fails on MacoOS because there is a space in some path


jhjourdan:

Lysxia:
@xldenis do you have a repro by any chance?

xldenis:
there's no error actually, it just fails later because nothing is installed at the desired path. A manual `cargo install --path creusot-rustc --root "/Users/xavier/Library/Application Support/creusot.creusot/toolchains/nightly-2025-01-31"` works

Lysxia:
The status is that `creusot-rustc` installs fine now, but `opam` is mishandling that space in MacOS so it can't set up a switch in the `data_dir`.

Lysxia:
> there's no error actually, it just fails later because nothing is installed at the desired path. A manual `cargo install --path creusot-rustc --root "/Users/xavier/Library/Application Support/creusot.creusot/toolchains/nightly-2025-01-31"` works

I know what happened!

```
    let _ = fs::remove_dir_all(&paths.config_dir);
    let _ = fs::remove_dir_all(&paths.data_dir.join("bin"));
```

I added `.join("bin")` to not remove everything in `data_dir`, but `config_dir` and `data_dir` are the same on MacOS so...

PR #1383, opened 2025-02-27
install: Remove lossy path-to-string conversions

Lysxia:


Lysxia:
@xldenis does this fix #1361 ?

xldenis:
no luck:

```
[ERROR] The compilation of ocaml.5.2.1 failed at "ocaml /Users/xavier/Library/Application Support/creusot.creusot/_opam/share/ocaml-config/gen_ocaml_config.ml
        5.2.1 ocaml".
```

Lysxia:
Yeah that one looks like a bug in opam. But wasn't there another issue with the `cargo install` step not working either? Does `./INSTALL --external why3-and-why3find` install `creusot-rustc` in the right path?

xldenis:
That seems to be fix now, running that command gives:

```
Building prelude...
Installing cargo-creusot...
Installing creusot-rustc...
Error: Why3 not found. Please install Why3 version 1.8.0
```

which is better than before?

Lysxia:
Yeah that looks fine :) I'm not sure that this is what fixed it but in any case it's good to have fewer `to_string_lossy`.

I'm not sure what to do about the opam issue. Maybe we can put the switch in `$HOME` for Mac users? You may want to submit the bug report (I would if I could reproduce it but my opam doesn't have trouble with spaces in switch paths.)

xldenis:
> I'm not sure what to do about the opam issue. Maybe we can put the switch in $HOME for Mac users? You may want to submit the bug report (I would if I could reproduce it but my opam doesn't have trouble with spaces in switch paths.)

I'll start by attempting to reproduce it outside of creusot, I find this such a weird issue anyways.

PR #1384, opened 2025-02-27
Fix install script for MacOS

### LINKED ISSUE #1361: Install script fails on MacoOS because there is a space in some path

Lysxia:
Fixes #1361

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -110,10 +110,14 @@ jobs:
         cat ~/.config/creusot/why3.conf
     - run: cargo test --test why3
   install:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v4
     - name: Cache Cargo build
+      if: matrix.os == 'ubuntu-latest'
       uses: actions/cache@v4
       with:
         path: |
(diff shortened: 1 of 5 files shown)


Lysxia:
I can reproduce @xldenis's problem!

ia0:
This smells like a `sandbox-exec` problem (so it's like `creusot -> opam -> sandbox-exec` blame chain :sweat_smile:). Can you run the following command on MacOS and see if you get `foo      bar` and not `foo bar`?

```
~/.opam/opam-init/hooks/sandbox.sh build echo 'foo      bar'
```

This works on Linux which suggests that `opam` does call this script correctly. If that's really the issue that's quite worrying of `sandbox-exec`.

xldenis:
```
~/.opam/opam-init/hooks/sandbox.sh build echo 'foo      bar'
foo      bar
```
seems fine.

ia0:
Ok weird, then it's probably in `opam` but specific to MacOS?

Lysxia:
That does look like an opam issue. I propose to put Creusot's switch in `$HOME/.creusot` on MacOS for now.

Lysxia:
The issue appears to be known: https://github.com/ocaml/opam/issues/5679 according to which the cause is in ocaml's build script.

I could actually reproduce the error on Linux too. (When I tested this earlier I somehow managed to skip recompiling ocaml.) So the bug exists regardless of the OS, but it's only on MacOS that the data directory we are using for a switch contains a space.

PR #1399, opened 2025-03-03
Remove warning about strings being experimental

Lysxia:
After this change I'm happy to make a release


xldenis (approved):

Lysxia:
The MacOS runner somehow regressed from rust 1.85 to 1.84 overnight???

PR #1402, opened 2025-03-03
Fix `Seq::get_mut_ghost`

arnaudgolfouse:
There was an inconsistency here, that allowed the proof of `linked_list` to pass even though it was incorrect.


Armael:
Well spotted!

PR #1411, opened 2025-03-04
Update Rust edition in all crates

Lysxia:
I was wondering why `cargo fmt` was not behaving the same as `rustfmt`. `rustfmt` defaults to the latest edition and "style edition".

### DIFF
--- a/creusot-contracts-dummy/Cargo.toml
+++ b/creusot-contracts-dummy/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "creusot-contracts-dummy"
 version = "0.4.0"
-edition = "2021"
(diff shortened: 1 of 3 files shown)


jhjourdan:
The problem is that the backend of Creusot does not support partial captures yet (TODO !), so we cannot use recent editions for code which is read by Creusot....

jhjourdan:
Recent editions also generate different code for panic, which we don't handle.

Lysxia:
Oh I see. Maybe we can at least update the non-creusot-contracts crates, and the dummy one.

jhjourdan:
Ok, LGTM.

PR #1416, opened 2025-03-04
Unsafe operations in `ptr_own`

arnaudgolfouse:


creusot-contracts/src/ptr_own.rs
```diff
@@ -16,32 +16,42 @@ pub type RawPtr<T> = *const T;
...
+/// When using Creusot to verify the code, all methods should be safe to call. Indeed,
+/// Creusot ensures that every operation on the inner value uses the right [`PtrOwn`] object
+/// created by [`PtrOwn::new`], ensuring safety in a manner similar to [ghost_cell](https://docs.rs/ghost-cell/latest/ghost_cell/).
+pub struct PtrOwn<T: ?Sized> {
+    #[cfg(creusot)]
+    ptr: RawPtr<T>,
+    #[cfg(creusot)]
+    val: Box<T>,
+    #[cfg(not(creusot))]
+    _val: std::marker::PhantomData<T>,
```
jhjourdan, line 31: Why do we need a different implementation in the case we compile without Creusot?

PR #1420, opened 2025-03-07
Add difference operation in Fset 

Halbaroth:
This operation is supported by Why3's finite set theory:
https://www.why3.org/stdlib/set.html

This PR introduces this operation for Fset and includes the support
for the `Difference` iterator of `Vec`.

Note: after rebased on #1419, I got: 
```
Testing tests/creusot-contracts/creusot-contracts.coma ... not proved 
```
Is it normal?


jhjourdan:
> ```Testing tests/creusot-contracts/creusot-contracts.coma ... not proved```

You need to update the session file associated with the creusot-contracts file:

```
$ ./ide tests/creusot-contracts/creusot-contracts
```

and then fix the proofs (which, in your case, should be rather trivial).

Halbaroth:
I thought that I ran this command before running `cargo test --test why3 -- --update`. After running it, I got a `obselete` warning for this coma file. I guess that there are obsolete proofs in the why3 sessions. Finally, I restored the Why3 session of master and ran again these commands.

tests/creusot-contracts/creusot-contracts/why3session.xml
```diff
@@ -560,41 +750,49 @@
 </theory>
 <theory name="M_creusot_contracts__stdqy35z1__cmp__qyi8793870187857001919__refl" proved="true">
  <goal name="vc_refl&#39;0" proved="true">
+ <proof prover="0"><result status="valid" time="0.018866" steps="2896"/></proof>
```
Lysxia, line 753: These are duplicate proofs. Are you manually selecting Z3 in the provers? Strategies like "auto level 1" (shortcut key 1) clean up the tree when multiple provers succeed.

jhjourdan:
I updated the session file. Thanks for the contribution.

PR #1421, opened 2025-03-07
Clarify documentation for update proof.json

Halbaroth:


CONTRIBUTING.md
```diff
@@ -28,7 +28,7 @@ Once you are satisfied with the coma output, you can validate the proofs of Creu
 If you add a test that you believe should include a proof, you can add it using the `./ide` script provided in Creusot.
 Load your test case in the Why3 IDE, solve the proof and save the result, it will now be checked as part of CI.
 
-Options:
+Options of `cargo test -- test why3`:
```
jhjourdan, line 31:
```suggestion
Options of `cargo test --test why3`:
```

why3tests/tests/why3.rs
```diff
@@ -239,7 +239,7 @@ fn main() {
             if sessionfiles.len() > 0 {
                 write!(out, "{current}").unwrap();
                 out.set_color(ColorSpec::new().set_fg(Some(Color::Red))).unwrap();
-                writeln!(&mut out, "unused {sessionfiles:?}").unwrap();
+                writeln!(&mut out, "unused {sessionfiles:?}. Please update the proof.json file.").unwrap();
```
jhjourdan, line 242:
```suggestion
                writeln!(&mut out, "unused {sessionfiles:?}. Please do not use Why3 sessions files for this test. Instead, update the proof.json file.").unwrap();
```

jhjourdan:
In addition, could you please fix code formatting by running the `fmt` script at the root of the repo?

Halbaroth:
I applied your suggestion but it didn't manage to reproduce the same output that the fmt step of the CI. I tried:
1. Installed zsh and ran `./fmt`, no effect. 
2. Ran `cargo fmt`, and I got a huge diff. I also tried `cargo fmt --check --package why3`. 
3. Finally, I executed the Ci commands in bash, and no effect again. 

My version of Cargo is given by your rust-toolchain. I am not a Rust developer, so I might be missing something obvious?

Besides, there is no mention to this script in your guide line. As `cargo fmt` produced a huge diff, I believed that you didn't use it at all.

Lysxia:
Try `rustfmt **/*.rs` (which should be what `fmt` runs so it's odd that it has no effect).

Indeed `cargo fmt` produces a different result, but maybe we can add some configuration so it works the same as `rustfmt` (it's just that some editions are not 2024, but we can probably set the "style edition" idependently).

PR #1422, opened 2025-03-07
rustfmt: Set style_edition = "2024"

Lysxia:
This makes `cargo fmt` produce the same output as a manual `rustfmt` command. I guess the problem is that we have some non-2024 edition crates and `rustfmt` doesn't know that because it ignores `Cargo.toml`.


jhjourdan:
LGTM.

PR #1425, opened 2025-03-11
Specify `std::hint::*`, `std::mem::drop` and `std::mem::forget`

arnaudgolfouse:


creusot-contracts/src/std/hint.rs
```diff
@@ -0,0 +1,35 @@
...
+
+extern_spec! {
+    mod std {
+        mod hint {
+            #[pure]
+            #[requires(cond)]
+            unsafe fn assert_unchecked(cond: bool) {}
+
+            #[pure]
+            #[ensures(true)]
```
ia0, line 11: I was expecting `#[ensures(result == dummy)]` (similarly to `must_use` below). One reason I could see, is that this is the spec the compiler sees (in particular, it can't look at the implementation to figure out a better spec). But I'm not sure users expect this spec.

jhjourdan: Agreed. We should add a spec here.

jhjourdan (approved):

PR #1430, opened 2025-03-12
Rename `GhostBox` to `Ghost`

arnaudgolfouse:
Also make it a wrapper around `T` rather than `Box<T>`.


jhjourdan:
Next step is to make the field public, and check in Creusot that we are only accessing this field/using the constructor in a ghost! block.

PR #1433, opened 2025-03-12
cargo creusot new: Set specific version of creusot-contracts

Lysxia:


creusot-setup/src/lib.rs
```diff
@@ -231,3 +231,10 @@ pub fn toolchain_channel() -> String {
         .expect("Expected `cargo-creusot` to be built with a valid toolchain file");
     toolchain["toolchain"]["channel"].as_str().unwrap().to_string()
 }
+
+pub fn creusot_contracts_version() -> String {
+    let creusot_contracts: toml::Value =
+        toml::from_str(include_str!("../../creusot-contracts/Cargo.toml"))
```
ia0, line 237: If creusot-setup is a binary shipped to users, it will contain the full creusot-contracts toml file just for the version. An optimization would be to do this extraction in the build.rs and only embed the version.

Lysxia: By "const code" do you mean for `build.rs` to generate `const VERSION = "0.4.0";`?

ia0: I think Xavier meant `const fn`, but those are not yet well supported in Rust and you might suffer, but it's worth trying.

(Also it's not necessarily a big issue to embed the whole toml file, I just mentioned it for awareness regarding performance and binary size, which are probably not big concerns for this tool.)

PR #1436, opened 2025-03-13
Do not check that trusted impls are correct.

jhjourdan:
These impls may require additional predictates from extern specs, and these may not be implemented.


jhjourdan:
Cc @Halbaroth : that should fix the issue we discussed today.

PR #1440, opened 2025-03-17
Refactor internal ASTs

jhjourdan:
These three commits were preleminary work for a failed project.

They are:
1- Use more smart constructors in the Why3 ASTs.
2- Translate Rust tuples into records instead of Why3 tuples. This makes projections easier, and avoids some Why3 polymorphism.
3- Add type invariants in signatures after having instantiated them. This makes more type invariant being detected as trivial, and simplifies the coma files a lot.
4- Store types in Pearlite patterns.


xldenis:
> 3- Add type invariants in signatures after having instantiated them. This makes more type invariant being detected as trivial, and simplifies the coma files a lot.


I'm curious if you are able to measure an improvement in proof times as a result? Looking over the coma files indicates that yes, maybe, but those times are unnormalized (unless we're calibrating why3find?)

PR #1442, opened 2025-03-18
Give a body to almost every extern spec in `option.rs`

dianegolfouse:
This did uncover a spec (`is_some_and`) which was incorrect wrt resolution.


jhjourdan:
> This did uncover a spec (`is_some_and`) which was incorrect wrt resolution.

Do you have more detail?

creusot-contracts/src/std/slice.rs
```diff
@@ -365,6 +365,18 @@ extern_spec! {
         #[ensures(ix.has_value(self@, *result))]
         fn index(&self, ix: I) -> &<[T] as Index<I>>::Output;
     }
+
+    mod std {
+        mod slice {
+            #[pure]
+            #[ensures(result@.len() == 1 && result@[0] == *s)]
```
jhjourdan, line 372:
```suggestion
            #[ensures(result@.len() == 1)]
            #[ensures(result@[0] == *s)]
```

creusot-contracts/src/std/slice.rs
```diff
@@ -365,6 +365,18 @@ extern_spec! {
...
+            fn from_ref<T>(s: &T) -> &[T];
+
+            #[pure]
+            #[ensures(result@.len() == 1 && (*result)@[0] == *s && (^result)@[0] == ^s)]
```
jhjourdan, line 376:
```suggestion
            #[ensures(result@.len() == 1)]
            #[ensures(result@[0] == *s)]
            #[ensures((^result)@.len() == 1)]
            #[ensures((^result)@[0] == ^s)]
```

We need to ensure that the length of the prophecy is 1, too.

Some day we will figure out a way to ensure freely that the length of the current and final values are equal.

tests/should_succeed/option.rs
```diff
@@ -45,91 +32,31 @@ pub fn map() {
...
-        some.map_or(
-            -1,
-            #[requires(x < i32::MAX)]
-            #[ensures(result@ == x@ + 1)]
-            |x| x + 1
-        ) == 2
-    );
+    assert!(none.map_or(2, |_| panic!()) == 2);
+    assert!(some.map_or(-1, |_| 3) == 3);
+    assert!(some.map_or(-1, |x| x + 1) == 2);
```
jhjourdan, line 55: These diffs are so good to see...

jhjourdan:
> The resolve(&t) was incorrect, because the t will be passed to the f closure, which will handle resolution.

Indeed. Could you fix the bug?

arnaudgolfouse:
I don't understand: the spec no longer contains the resolve. It did before, and was (blindly) accepted because it was axiomatized (since `is_some_and` did not have a body).

jhjourdan:
> I don't understand: the spec no longer contains the resolve. It did before, and was (blindly) accepted because it was axiomatized (since is_some_and did not have a body).

Sorry. I did not notice in the diff that you indeed fixed the bug.

creusot-contracts/src/std/option.rs
```diff
@@ -210,6 +283,7 @@ extern_spec! {
                     }
                 }
 
+                // FIXME: cannot add a body, deref is mistranslated (in logic?)
                 #[ensures(match *self {
                     None => result == None,
                     Some(t) => result == Some(&t.deref()),
```
jhjourdan, line 289: huh this is weird... `Deref::deref` is not a logic function, but rather a program function. This should actually be rejected. Instead, we should add a `DerefLogic` trait, with the spec of the `Deref` function...

(There is an exception for `Box` and friends, but here we are considering the generic case.)

arnaudgolfouse: Yes, I think the exception accepts too much (every instance of deref and deref_mut is accepted).
Actually, let me try to fix that...

arnaudgolfouse: Ok, done: I also removed this function, since I think it was unusable in practice anyways. I definitely agree that we should have `DerefLogic` at some point though.

jhjourdan:
Alright, LGTM, merging!

PR #1444, opened 2025-03-18
Infer the FnMut spec of a Fn closure (and other combinations)

### LINKED ISSUE #1443: Incomplete spec inference for `FnOnce`
```rust
pub fn foo() {
    let clos = |x: i32| x + 1;
    let y = call_with_one(clos);
    assert!(y == 2);
}

#[requires(f.precondition((1i32,)))]
#[ensures(f.postcondition_once((1i32,), result))]
fn call_with_one<F: FnOnce(i32) -> i32>(f: F) -> i32 {
    f(1)
}
```

The following does not work, but it should. In particular it works with `Fn`.
This seems to be because the generated `postcondition_once` predicate is not given any specification: this looks like a simple oversight.
@xldenis where is the code that generate this stuff please? 😁

arnaudgolfouse:
Fix #1443


arnaudgolfouse:
Should I give a number to the new test, like the others in `tests/closures`?

tests/should_succeed/closures/inference.rs
```diff
@@ -0,0 +1,53 @@
+extern crate creusot_contracts;
+use creusot_contracts::*;
+
+#[requires(f.precondition((1i32,)))]
+#[ensures(f.postcondition((1i32,), result))]
+fn call_with_one<F: Fn(i32) -> i32>(mut f: F) -> i32 {
```
jhjourdan, line 6:
```suggestion
fn call_with_one<F: Fn(i32) -> i32>(mut f: &F) -> i32 {
```

jhjourdan:
Oh, and btw, the definition you generate for postcondition_once in the mut case are slightly too weak, because they don't say that the closure is resolved a the end of the call.

PR #1448, opened 2025-03-19
Update Why3/Why3find versions

Lysxia:
- in Why3, a memory leak was fixed (affects Creusot IDE)


Lysxia:
@arnaudgolfouse Does this fix the leak you observed?

PR #1457, opened 2025-03-24
Set the ocaml compiler version

arnaudgolfouse:
I had issues related to the compiler version when installing creusot-ide.


jhjourdan:
Could you specify the ocaml version in the .opal file as a dependency ? This seems more appropriate.

Lysxia (approved):

PR #1458, opened 2025-03-25
Soundly erase `ghost!`

### LINKED ISSUE #1455: Erasing ghost/snapshot code is unsound
... Because it may change type inference.

# Concrete example

```rust
trait Tr {
    fn foo(self) -> i32;
}
impl Tr for i32 {
    #[ensures(result@ == 1)]
    fn foo(self) -> i32 { 1 }
}
impl Tr for u8 {
    #[ensures(result@ == 2)]
    fn foo(self) -> i32 { 2 }
}

fn bar() {
    let x = 0;
    ghost! { if x == 0u8 {} };
    let res = x.foo();
    assert!(res == 2);
}
```

# Solution for `ghost!`

compile the  `ghost!` macro in normal mode with `if false { Ghost::new(<body>) } else { Ghost::conjure() }`

# Solution for `snapshot!`

??? See <https://docs.rs/ghosts/latest/ghosts/>

arnaudgolfouse:
Partially resolve #1455

### DIFF
--- a/creusot-contracts-dummy/src/lib.rs
+++ b/creusot-contracts-dummy/src/lib.rs
@@ -44,7 +44,12 @@ pub fn snapshot(_: TS1) -> TS1 {
 #[proc_macro]
 pub fn ghost(body: TS1) -> TS1 {
     let body = proc_macro2::TokenStream::from(ghost::ghost_preprocess(body));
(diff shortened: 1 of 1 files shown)


jhjourdan:
Are we sure that the ghost piece of code still compiles in non-creusot mode ? For example, don't we have ghost libraries in creusot-contracts that are ony enabled in creusot mode?

PR #1460, opened 2025-04-02
Minor improvements in `why3` crate

arnaudgolfouse:


why3/src/coma.rs
```diff
@@ -135,10 +136,6 @@ impl Defn {
 }
 
 impl Expr {
-    pub fn boxed(self) -> Box<Self> {
-        Box::new(self)
-    }
-
```
jhjourdan, line 141: Why don"t you like this?

That's a function I recently added, wimply because I find it easier to write and read `Constr(blah.boxed())` `Constr(Box::new(blah))`. One immediate advantage is the lower nesting of parentheses.

PR #1463, opened 2025-04-10
Lower program signatures to Coma rather than WhyML

arnaudgolfouse:
This hopefully makes things a bit clearer. But mainly I needed to be able to specify reference arguments (`& x`) easily.


why3/src/coma.rs
```diff
@@ -107,12 +107,23 @@ pub enum Arg {
     Cont(Expr),
 }
 
+/// The signature of a coma handler.
 #[derive(Clone, Debug)]
 #[cfg_attr(feature = "serialize", derive(Serialize, Deserialize))]
-pub struct Defn {
+pub struct Prototype {
```
jhjourdan, line 113:
```suggestion
pub struct HandlerSig {
```

(That's not a strong query from my part, but this proposal is almost as short and more descriptive. Also, I don't quite the word "Prototype" for a functional language like Coma.)

PR #1465, opened 2025-04-15
Improve binder hygiene

Lysxia:
See comments in `why3::name` and `why3::printer`. All variables from rust code now go through `Ident::fresh` so collision should be impossible. There are only five special non-fresh variables used internally to generate specs for closures.

Some TODOs are to make substitution on Why3 expressions capture-avoiding (I only did it for Pearlite terms), and to make sure that module names don't collide with Why3 keywords (because this PR moved the check elsewhere), but users kinda have to go out of their way to poke those holes so it's not pressing.


jhjourdan:
That's really, really nice.

I'd like to have an in-depth look, but I am leaving for vacations on Thursday for almost 2 weeks. If you could leave this on hold, I would be grateful. But if you need this to be merged to continue working on something else, then please go ahead.

Issue #1447, opened 2025-03-19
Handle missing things in creusot-contracts more gracefully


Lysxia:
Whenever we add primitives in `creusot-contracts` that are directly used by `creusot` we start encountering this error:

```
error: The `creusot_contracts` crate is loaded, but the following items are missing: ghost_new, ghost_into_inner, ghost_inner_logic, ghost_deref, ghost_deref_mut, ghost_ty. Maybe your version of `creusot-contracts` is wrong?
```

This happens on projects that just specify a version string in their `Cargo.toml` (`creusot-contracts = "0.4.0"`) which gets the crate from crates.io instead of locally (`creusot-contracts = { path = "/path/to/creusot-contracts" }`).

1. The root of the issue is that the versions of `creusot-rustc` and `creusot-contracts` are tied together, so the error message should really be "expected version X of `creusot-contracts` but got version Y".
2. One challenge with that is that currently, the "dev" version number is the same as the "latest on crates.io" version number. We should modify the "dev" version number. Maybe right after a release we can edit the version number from `X.Y.Z` to something like `X.Y.Z+dev`, but a solution without extra commits would be nicer. UPDATE: There doesn't seem to be a way to do this (there is [a long-standing cargo feature request](https://github.com/rust-lang/cargo/issues/6583)). So I think always bumping the version with a `+dev` right after a release is the way to go.
3. Beyond the error message, it should be easier to use the right version of `creusot-contracts` (we can manually edit the `Cargo.toml` but I'd like to automate that). UPDATE: Cargo has a "source replacement" feature https://doc.rust-lang.org/cargo/reference/source-replacement.html
4. Often the error is spurious because the project doesn't use the missing features. In that case it would be more convenient for that error to be a warning only.

PR #1466, opened 2025-04-15
Added temporary documentation for versioning mismatch issue

HumamAlhusaini:
When I was following the [quickstart](https://creusot-rs.github.io/creusot/guide/), I came across this error: 
`The creusot_contracts crate is loaded, but the following items are missing: ghost_new, ghost_into_inner, ghost_inner_logic, ghost_deref, ghost_deref_mut, ghost_ty. Maybe your version of creusot-contracts is wrong?`

Luckily for me, issue #1447 gave me the solution to this issue, that being to replace `creusot-contracts = "0.4.0"` with `creusot-contracts = { path = "/path/to/creusot-contracts"` in Cargo.toml of the new project. The issue also gives some useful details about the issue.

Until this issue is solved, I think that some special documentation should be added to quickstart.md so nobody will need to look for #1447  in order to get the solution.

### DIFF
--- a/guide/src/quickstart.md
+++ b/guide/src/quickstart.md
@@ -37,6 +37,20 @@ A successful run gives us the certainty that functions defined in this package s
 for all arguments satisfying the preconditions (`requires` clauses), the result of the function will
 satisfy the postconditions (`ensures` clauses).
 
+However, if you get this error 
+```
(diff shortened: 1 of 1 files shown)


Lysxia:
It's good to have a bit of documentation until we have a better fix. Thanks for the help!

PR #1469, opened 2025-04-18
Merge creusot-contracts-dummy into creusot-contracts-proc

### LINKED ISSUE #1464: usage of symbolic link
I am working on a project using Creusot, and sometimes I need to compile it on windows.
The problem is that I cannot build my project with the last release:

```bash
error: expected item, found `..`
 --> creusot\creusot-contracts-dummy\src\ghost.rs:1:1
  |
1 | ../../creusot-contracts-proc/src/ghost.rs
  | ^^ expected item
  |
  = note: for a full list of items that can appear in modules, see <https://doc.rust-lang.org/reference/items.html>
```

I think the issue is with commit 25b9dfd2e5d773e3e927901a43739bb277399d00 where a symbolic link was added.
To my knowledge, symbolic links are different in windows.

Is it something that you would like to fix?

Lysxia:
To avoid a symbolic link for ghost.rs. Fixes #1464


Lysxia:
One drawback of hiding the non-dummy code behind a flag is that rust-analyzer doesn't see it.

@arnaudgolfouse do you know a way to make rust-analyzer see all of the code in the `creusot-contracts-proc` crate, but so that `cargo build` and `cargo creusot` still only build the relevant bits?

Lysxia:
I found a way by adding a feature that gets enabled by both creusot-rustc and rust-analyzer to compile the whole creusot-contracts-proc crate.

Issue #1164, opened 2024-10-17
incorrect spec for `String::len`?


arnaudgolfouse:
```rust
impl ShallowModel for String {
    type ShallowModelTy = Seq<char>;
}

#[ensures(result@ == self@.len())]
fn len(&self) -> usize;
```
The length of the string is in bytes, so this spec is not correct I think? Same for `str::split_at`. Although I guess there is no way no observe this at the moment...

xldenis:
Hmm yea I guess we should relax that to >=?

arnaudgolfouse:
This would work, but shouldn't we actually give the correct length? With a logical function `utf8_len(char) -> usize`
I guess this would make `str::len` hard on provers, but if we use `>=` it would be a bit useless.

xldenis:
Depends on what we need it for really.

jhjourdan:
Do we want to formalized Unicode? No.
So the only alternative we have is not supporting `len` at all, which is definitely an option.

arpj-rebola:
>  So the only alternative we have is not supporting `len` at all, which is definitely an option.

Not really, `String::len` has interesting properties that can be relevant (e.g. the length of string concatenation is the sum of lengths of concatenated strings; and similarly for splitting and substrings). It just doesn't correspond to the notion of length in `Seq`. One possible way there might be formalizing some (very basic) contract of `str::as_bytes`, which behaves nicely and compositionally, and then define `str::len` based on that.

Is `Seq<char>` a good model for `String`, as opposed to `Seq<u8>`, though?

Lysxia:
> Is Seq<char> a good model for String, as opposed to Seq<u8>, though?

I think `Seq<char>` lets us hide encoding details better. One can still switch to a `Seq<u8>`-based representation via `str::as_bytes`.

PR #1484, opened 2025-04-24
Fix and add some extern specs for strings

### LINKED ISSUE #1379: Support for unsafe operations that Creusot can support
Creusot probably can't support unsafe operations with requirements mentioning the memory or aliasing model, but it should be able to support purely arithmetic safety requirements. For example `slice::get_unchecked()`, `u64::unchecked_add()`, or `String::from_utf8_unchecked()`. A related closed issue is #36 which actually mentions `get_unchecked` as something that could be soundly supported.

This issue is about whether such unsafe operations would be something Creusot wants to support.

Lysxia:
Close #1379 and #1164


arnaudgolfouse (approved):
Very nice, this puts us in a good position to eventually add more specs!

Issue #1439, opened 2025-03-13
Match with no branches in pearlite


arnaudgolfouse:
This crashes Creusot:
```rust
#[requires(match opt { })]
fn foo(opt: Option<i32>) {}
```

jhjourdan:
That should be easy, isn't it?

PR #1486, opened 2025-04-25
Handle empty type in logic

### LINKED ISSUE #972: `ExprKind::NeverToAny` => `Term::Absurd` => Invalid Why3
When running CI for the Prusti translation I noticed that `absurd` seem to no longer be valid `Why3`, but it still can get generated from the THIR never to any expression eg.
```rust
#[ensures(match x {Some(x) => x, None => true})]
//                            ^ x is implicitly cast from ! to bool, which generates ExprKind::NeverToAny
fn test(x: Option<!>) {}
```

Lysxia:
Fix #972 and #1439

### DIFF
--- a/creusot/src/backend/term.rs
+++ b/creusot/src/backend/term.rs
@@ -111,9 +111,13 @@ impl<'tcx, N: Namer<'tcx>> Lower<'_, 'tcx, N> {
 
                     Exp::qvar(of_qname).app([Exp::qvar(to_qname).app([self.lower_term(arg)])])
                 }
+                TyKind::Never => Exp::ascribe(
+                    Exp::qvar(self.names.in_pre(PreMod::Any, "any_l")).app([Exp::unit()]),
+                    translate_ty(self.ctx, self.names, term.span, term.ty),
+                ),
                 _ => self.ctx.crash_and_error(
                     DUMMY_SP,
-                    "casting from a type that is not a boolean is not supported",
+                    "casting from a type other than booleans, integers, and `!` is not supported",
                 ),
             },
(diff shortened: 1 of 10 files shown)


creusot/src/backend/term.rs
```diff
@@ -212,7 +216,12 @@ impl<'tcx, N: Namer<'tcx>> Lower<'_, 'tcx, N> {
             }
             TermKind::Old { box term } => Exp::Old(self.lower_term(term).boxed()),
             TermKind::Match { box scrutinee, arms } => {
-                if let PatternKind::Bool(b0) = arms[0].0.kind {
+                if arms.is_empty() {
+                    Exp::ascribe(
+                        Exp::qvar(self.names.in_pre(PreMod::Any, "any_l")).app([Exp::unit()]),
+                        translate_ty(self.ctx, self.names, term.span, term.ty),
+                    )
```
jhjourdan, line 223: The behavior you choose here is to use an indeterminate value when pattern matching on the ! type in Pearlite. Recall that in the logic, this type is interpreted as unit, because Why3 does not support empty types.

I would be rather in favor of simply not allowing empty pattern matching in Pearlite. Two reasons for this:

1- I don't see in which case pattern matching on ! this way can be useful. If the user needs an indeterminate value, then she can use dedicated features.

2- This is confusing, because this gives the wrong belief that ! is empty in Pearlite. I would greatly prefer a pedagogical error message.

Lysxia: I see. I'm still not used to thinking that there are no empty types in Why3.

PR #1490, opened 2025-04-25
Forbid empty matches and casts from ! in logic

Lysxia:
They don't make sense in Why3 because all types are inhabited.

This partly reverts #1486, introducing specific error messages instead

Fix https://github.com/creusot-rs/creusot/issues/972 and https://github.com/creusot-rs/creusot/issues/1439 better


jhjourdan (commented):
LGTM.

PR #1491, opened 2025-04-25
Support `?` operator

### LINKED ISSUE #761: syntax error on empty type
Using this file:

```rust
use creusot_contracts::*;

pub fn f() -> Option<()> { None }
pub fn g() -> Option<()> {
    f()?;
    None
}

```

why3 (master 6b489907fb63b33bd16a52841b0472a601812cb0 ) complains:

> File "../proto-rlib.mlcfg", line 46, characters 0-3: syntax error

The error occurs on the definition of module Core_Convert_Infallible_Type:

```whyml
module Core_Convert_Infallible_Type
  type t_infallible  =
    
end
```

Looks like why3 doesn't like the empty type definition.

As an aside, I am not sure why ! gets brought into the context here. The sample file (and the actual file it was reduced from) doesn't use ! as far as I can tell. But maybe the ? operator needs to know about ! somehow?

Lysxia:
Close #761

Is this the right spec for `from`? The motivation is to be able to add an extern spec to `FromResidual<...> for Result<T, F>`. The test case 761.rs only uses the special case `impl From<T> for T`.


creusot-contracts/src/std/convert.rs
```diff
@@ -0,0 +1,27 @@
+use crate::*;
+
+/// Specification trait for `::std::convert::From`.
+pub trait From<T>: ::std::convert::From<T> {
```
Lysxia, line 4: I'm copying what `Iterator` does. But why do we reuse the same name? Another name like `FromSpec` seems clearer to me.

tests/should_succeed/ite_normalize.rs
```diff
@@ -76,24 +76,43 @@ extern_spec! {
...
     fn from(b: bool) -> Self {
         if b { Self::True } else { Self::False }
     }
 }
 
+impl From<usize> for Expr {
+    #[predicate]
+    #[open]
+    fn comes_from(self, v: usize) -> bool {
+        self == Expr::Var { v }
```
jhjourdan, line 97: Alternatively you could use `true` here. That would be closer to the current code.

tests/should_succeed/ite_normalize.rs
```diff
@@ -76,24 +76,43 @@ extern_spec! {
...
+    }
+}
+
+impl From<bool> for Expr {
+    #[predicate]
+    #[open]
+    fn comes_from(self, b: bool) -> bool {
+        self == if b { Expr::True } else { Expr::False }
```
jhjourdan, line 105: Idem

PR #1492, opened 2025-04-28
Add the 'Persistent arrays' test case

arnaudgolfouse:


jhjourdan:
We are waiting for the version with local invariants and ghost state for merging this, but this should be merged as soon as we have a working version.

tests/should_succeed/persistent_array.rs
```diff
@@ -35,152 +31,180 @@ pub mod implementation {
     ///
     /// If you don't, you must ensure that a [`TokensMap`] object is never used with a
     /// `PersistentArray` that is not derived from it.
-    pub struct PersistentArray {
-        /// Contains the view of the value
-        logical_value: Snapshot<Seq<Value>>,
+    pub struct PersistentArray<T: 'static> {
```
jhjourdan: I guess you no longer need the `'static`, here and at plenty of other places in the file.

tests/should_succeed/persistent_array.rs
```diff
@@ -35,152 +31,180 @@ pub mod implementation {
...
+        type Public = Id;
+
         #[logic]
-        fn invariant_inner(self) -> bool {
+        fn invariant_with_data(self, id: Id) -> bool {
             pearlite! {
-                (forall<id: LogicId> self.own_map.contains(id) ==> self.own_map[id].id() == *id) &&
-                (forall<id: LogicId> self.own_map.contains(id) ==> self.values[*id].len() == *self.length) &&
+                self.values.id() == id &&
+                (forall<id: _> self.own_map.contains(id) == self.values@.contains(*id)) &&
```
jhjourdan: You know you no longer need the `: _` for quantifiers?

dianegolfouse: Oh really? That's good!

tests/should_succeed/persistent_array.rs
dianegolfouse: Well I needed it for the `Ghost<&'static LocalInvariant<CompleteMap<T>>>`. But I removed it by using a `Rc` instead 😛
The only thing bothering me with this approach is that it only works because you cannot access the reference count in Creusot. As soon as we allow this, this becomes unsound, unless we have a `RcGhost` or something...

dianegolfouse: Ah, unless accessing the reference count is not `#[pure]` I guess.

jhjourdan: That's actually a very clever idea. There is no reason not using `Rc` in ghost code. Except of course for reading the count.

creusot-contracts/src/local_invariant.rs
```diff
@@ -193,6 +193,25 @@ impl<'a, T: LocalInvariantSpec> LocalInvariantExt<'a> for Ghost<&'a LocalInvaria
         LocalInvariant::open(self, namespaces, f)
     }
 }
+impl<'a, T: LocalInvariantSpec> LocalInvariantExt<'a>
```
jhjourdan: Instead of these two somewhat ad-hoc instances, couldn't we have an instance like this:

```rust
impl<T: LocalInvariantSpec, U: Deref<Target=LocalInvariant<T>>> LocalInvariantExt<'a> for Ghost<U> {
...
}
```

The spec then should contain something about `U::deref.precondition` (because it is false on e.g., `Snapshot` or `Ghost`) and `U::deref.postcondition(...)`. And of course, for this to be sound, we should add `#[pure]` to the `deref` method, but I think this is very reasonable.

Also, to avoid using `.borrow()` when calling `open`, it should take a `&self`. (It should be sound not to use &'a self).

dianegolfouse: > Also, to avoid using .borrow() when calling open, it should take a &self. (It should be sound not to use &'a self).

Ah, because the namespace lifetime already prevents the reopening of the invariant?

jhjourdan: > Ah, because the namespace lifetime already prevents the reopening of the invariant?

No, this argument is wrong, be cause we may want to tear the invariant down if we have its full ownership (as we do with `into_inner` of `Cell`, `ReffCell` an d`Mutex`).

And actually I think I was wrong saying that we can take a `&self` and not a `&'a self`, I was thinking that ther already is a nested `'a` borrow, but this is no longer the case if we use `Deref` instead of `&'a` inside the `Ghost`.

jhjourdan: One thing that bothers me is that (at least in theory), it is stronger to require a `&'a Ghost<..>` than a `Ghost<&'a ...>`, because `&'a Ghost<..>` freezes a physical memory location for the duration `'a`, which `Ghost<&'a ...>` does not. But it is more convenient, at least in our example.

I don't really know what is best to do... Could we automate the call to `borrow` by creating a well-chosen `Deref` instance?

dianegolfouse: > The spec then should contain something about `U::deref.precondition` (because it is false on e.g., `Snapshot` or `Ghost`)

Ah, that's not quite true... It is false in any generic context, meaning that it cannot be used here :( [1] (on the contrary, the precondition is `true` for `Snapshot` and `Ghost`, but there is a check in Creusot that we do not use `deref` in program for those types)

[1]: My understanding is that `T::foo.precondition` should always be stronger than the precondition of the trait function, right?

jhjourdan: Wait... how do we make sure that we do not call `Ghost::deref` or `Snapshot::deref` through a generic? Isn't this unsound?

I think what should happen instead is that these functions should have `false` as a precondition, and that a special mechanism should be used to make it `true` when called directly in appropriate context.

>  It is false in any generic context, meaning that it cannot be used here

Nope. It is not `false` in generic context, it is undefined. I don't see why we could not use that here.

> [1]: My understanding is that T::foo.precondition should always be stronger than the precondition of the trait function, right?

Right.

dianegolfouse: > Nope. It is not false in generic context, it is undefined. I don't see why we could not use that here.

Well this function:
```rust
#[requires(T::deref.precondition((x,)))]
pub fn foo<T: Deref<Target = i32>>(x: &T) { let y: &i32 = &*x; }
```
is not provable currently.

jhjourdan: See #1644. That should make this possible.

dianegolfouse: Nice, that did it

dianegolfouse:
I think that's good to go if you're ok with it!

creusot-contracts/src/local_invariant.rs
```diff
@@ -194,6 +194,33 @@ impl<'a, T: LocalInvariantSpec> LocalInvariantExt<'a> for Ghost<&'a LocalInvaria
     }
 }
 
+impl<'a, T, L> LocalInvariantExt<'a> for Ghost<&'a L>
+where
+    T: 'a + LocalInvariantSpec,
+    L: ::std::ops::Deref<Target = LocalInvariant<T>>,
+{
+    type Inner = T;
```
jhjourdan: I'm sorry, I'm still not entirely satisfied because

1- There is one instance for no `Deref`, one instance for one `Deref`, but not for, e.g., two derefs or more.
2- We still need to call `borrow()`.

So I propose something in the lines of:

```suggestion
impl<'a, L> LocalInvariantExt<'a> for Ghost<&'a L>
where
    L: ::std::ops::Deref<Target: LocalInvariantExt<'a>>,
{
    type Inner = L::Target::Inner
```

which should recursively call <L::Target as LocalInvariantExt>::open, peeling off all the derefs until using the main instance.

And another instance:

```rust
impl<'a, L> LocalInvariantExt<'a> for &'a Ghost<L>
where
    Ghost<&'a L>: LocalInvariantExt<'a>
{
    type Inner = <Ghost<&'a L> as LocalInvariantExt<'a>>::Inner
```

jhjourdan: (I agree, this is not particularly aesthetic. Maybe we should have a trait like `GhostDeref` to handle all this dereferencing business automatically.)

dianegolfouse: Alright, this works (although it's indeed not very pretty)

jhjourdan: Thanks.

jhjourdan:
Thanks. All my concerns are addressed, I think. Could you please fix the conflicts and merge?

PR #1495, opened 2025-05-02
Define a `FnPure` trait, used for calling closures in `#[pure]` contexts.

arnaudgolfouse:
Ideally I would want to tell the typechecker to implement `FnPure` on all `#[pure]` closures, but I don't know if this is possible.


xldenis:
The best suggestion I would have is for the pure attribute to implicitly wrap the closure in your purefn wrapper.

jhjourdan:
@arnaudgolfouse what's blocking this?

Unfortunately, since #1531, there are some conflicts to be solved.

arnaudgolfouse:
> The best suggestion I would have is for the pure attribute to implicitly wrap the closure in your purefn wrapper.

Thanks, this worked like a charm!

At this point I think nothing is blocking this.

Issue #1496, opened 2025-05-03
Dereferencing raw pointers should result in a Creusot error instead of failing at the Why3 stage


sarsko:
Surely there is a messup here on my part as well, but it still shouldn't panic like this.

```rust
use creusot_contracts::{
    logic::Seq,
    ptr_own::{PtrOwn, RawPtr},
    *,
};

struct Node<T> {
    v: T,
    next: RawPtr<Node<T>>,
}

pub struct List<T> {
    // actual data
    head: RawPtr<Node<T>>,
    last: RawPtr<Node<T>>,
    // ghost
    seq: Ghost<Seq<PtrOwn<Node<T>>>>,
}

impl<T> Invariant for List<T> {
    #[open]
    #[predicate]
    fn invariant(self) -> bool {
        pearlite! {
            (*self.seq == Seq::EMPTY &&
             self.head.is_null_logic() &&
             self.last.is_null_logic())
            ||
            (self.seq.len() > 0 &&
             self.head == self.seq[0].ptr() &&
             self.last  == self.seq[self.seq.len() - 1].ptr() &&
             // the nodes in `seq` are chained properly
             (forall<i:Int> 0 <= i && i < self.seq.len() - 1 ==>
                 self.seq[i].val().next == self.seq[i+1].ptr()) &&
             self.seq[self.seq.len() - 1].val().next.is_null_logic())
        }
    }
}

impl<T> View for List<T> {
    type ViewTy = Seq<T>;

    #[logic]
    #[open(self)]
    fn view(self) -> Self::ViewTy {
        pearlite! {
            seq_map(*self.seq, |ptr_own: PtrOwn<Node<T>>| ptr_own.val().v)
        }
    }
}

#[logic]
#[open(self)]
pub fn seq_map<T, U>(s: Seq<T>, f: logic::Mapping<T, U>) -> Seq<U> {
    Seq::create(s.len(), |i| f.get(s[i]))
}

impl<T> List<T> {
    #[ensures(result@ == Seq::EMPTY)]
    pub fn new() -> List<T> {
        List { head: std::ptr::null(), last: std::ptr::null(), seq: Seq::new() }
    }

    pub fn remove_better(&mut self, toremove: RawPtr<Node<T>>) {
        unsafe {
            let mut p = self.head;
            while p != toremove {
                p = (*p).next;
            }
        }
    }
}

```

Gives:

```
[...]

PR #1511, opened 2025-05-05
Forbid raw ptr deref

arnaudgolfouse:
Fix for #1496.


jhjourdan:
IT seems like this does not cover other related operations, such as casting a raw pointer to a reference.

jhjourdan:
My bad, such conversion involves a place that dereferences a raw pointer.

arnaudgolfouse:
Banning `transmute` and the like, the only way to get a reference from a pointer is to use `&*p` (as far as I know), and that generates a `Deref` place in MIR.

PR #1513, opened 2025-05-05
Correctly generate a false precondition for contractless external functions

### LINKED ISSUE #1493: Calling an unspecified extern function should generate a false VC, but it does not
This is coming from #1477:

```rust
use ::std::collections::HashMap;

use creusot_contracts::{View, ensures, logic, open};

struct Foo(HashMap<u64, u8>);

impl View for Foo {
    type ViewTy = <HashMap<u64, u8> as View>::ViewTy; //FMap<Int, u8>;

    #[logic]
    #[open]
    fn view(self) -> Self::ViewTy {
        self.0.view()
    }
}

impl Foo {
    pub fn new() -> Self {
        Self(HashMap::new())
    }

    #[ensures((^self)@.get(num@) == Some(bar))]
    pub fn add(&mut self, num: u64, bar: u8) {
        *self.0.entry(num).or_insert(bar) = bar;
    }
}
```

Functions `entry` and `or_insert` should generate a false VC, but they don't.

arnaudgolfouse:
Fix #1493.

The issue was that `ctx.externs` contained the `std` crate, and so the false precondition was not generated.
This removed some refinement proofs for `Ghost::deref`, which I assume should not have existed in the first place.


creusot-contracts/src/std/io.rs
```diff
@@ -0,0 +1,18 @@
+use crate::*;
+use ::std::fmt;
+
+extern_spec! {
+    mod std {
+        mod io {
+            // This is an implementation detail of `std`: we specify it so that we can use
+            // `print!` and `println!`.
+            #[ensures(true)]
+            fn _print(args: fmt::Arguments<'_>) {}
```
jhjourdan, line 10: Do we use `println!` or `print!` somewhere in verified code?

It's a bit bold to assume that these functions can be called without VC... This is assuming, for example, the formatting functions do never fail.

creusot/src/translation/specification.rs
```diff
@@ -270,20 +270,19 @@ pub(crate) fn contract_of<'tcx>(ctx: &TranslationCtx<'tcx>, def_id: DefId) -> Pr
...
         let bound = raw_inputs.iter().map(|(ident, _, _)| ident.0);
         let subst = erased_identity_for_item(ctx.tcx, def_id);
         let mut contract = contract_clauses_of(ctx, def_id)
             .unwrap()
             .get_pre(ctx, fn_name, bound)
             .instantiate(ctx.tcx, subst);
 
-        if contract.is_empty()
-            && !def_id.is_local()
-            && ctx.externs.get(def_id.krate).is_none()
```
jhjourdan, line 284: Why was this here in the first place? I'm afraid you are introducing a bug by removing this condition.

arnaudgolfouse: Yep, I applied the patch you sent me!

creusot-contracts/src/std/io.rs
xldenis: i have to pretty strongly disagree here. the implication is that you simply cannot use `print` in verified code since there will be no effective way to specify it, no more than we can for OOM errors. We _absolutely_ should support people having print statements in their verified code.

jhjourdan: I don't know... I wrote thousands of lines of verified code in my life, and I don't think there was any print statement in it!

If one wants to write code that uses IO, then either she needs to specify what it means to perform IO (and what is the global spec of a program that performs IO), or perform IO in unverified code which may call verified functions.

Tracing / debugging print statements is another concern, though. We may provide some infrastructure for this, if we want.

Issue #1519, opened 2025-05-07
Panic when refering to non-implemented trait methods in spec


Armael:
This panics (note the missing implementations of `le` and `idemp` in the impl for (T, U):

```rust
use crate::*;
#[allow(unused_variables)]

pub trait RA: Sized {
    #[logic]
    fn op(self, other: Self) -> Option<Self>;

    #[logic]
    #[ensures(
        (forall<b: Self> ! (b.le(a) && b.idemp())) ||
        (exists<b: Self> b.le(a) && b.idemp() &&
           forall<c: Self> c.le(a) && c.idemp() ==> c.le(b))
    )]
    fn maximal_idemp(a: Self);

    #[logic]
    fn le(self, other: Self) -> bool;

    #[logic]
    fn idemp(self) -> bool;
}

impl<T, U> RA for (T, U)
    where T: RA, U: RA
{
    #[logic]
    #[open]
    fn op(self, other: Self) -> Option<Self> { pearlite!{
        self.0.op(other.0).and_then_logic(|x:T| {
            self.1.op(other.1).and_then_logic(|y:U| {
                Some((x, y))
            })
        })
    }}

    #[logic]
    #[open(self)]
    #[ensures(
        (forall<b: Self> ! (b.le(a) && b.idemp())) ||
        (exists<b: Self> b.le(a) && b.idemp() &&
           forall<c: Self> c.le(a) && c.idemp() ==> c.le(b))
    )]
    fn maximal_idemp(a: Self) { }
}
```

```
thread 'rustc' panicked at creusot/src/validate/terminates.rs:435:39:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: core::option::unwrap_failed
   4: core::option::Option<T>::unwrap
             at /home/armael/.rustup/toolchains/nightly-2025-01-31-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:978:21
   5: creusot::validate::terminates::BuildFunctionsGraph::visit_specialized_default_function
             at ./creusot/src/validate/terminates.rs:435:20
   6: creusot::validate::terminates::BuildFunctionsGraph::function_call
             at ./creusot/src/validate/terminates.rs:357:34
   7: creusot::validate::terminates::CallGraph::build
             at ./creusot/src/validate/terminates.rs:593:17
[...]

PR #1520, opened 2025-05-07
Fix a crash in the termination check

arnaudgolfouse:
Fix for #1519

### DIFF
--- a/creusot/src/validate/terminates.rs
+++ b/creusot/src/validate/terminates.rs
@@ -398,7 +398,7 @@ impl<'tcx> BuildFunctionsGraph<'tcx> {
         Ok(())
     }
 
-    /// This visit the special function that is called when calling:
+    /// This visits the special function that is called when calling:
     /// - a default function in a trait (or in a default impl)
     /// - that is logical
     /// - and visible at the point of implementation, that is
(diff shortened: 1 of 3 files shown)


creusot/src/validate/terminates.rs
```diff
@@ -432,7 +438,11 @@ impl<'tcx> BuildFunctionsGraph<'tcx> {
 
         let impl_id = impl_id.to_def_id();
         let typing_env = ctx.typing_env(impl_id);
-        let term = ctx.term(item_id)?.unwrap();
+        let Some(term) = ctx.term(item_id)? else {
+            // ` ctx.term` can return None if the function we are searching for is not yet implemented.
+            // In this case, there is an Rust error happening: bail out and let the compiler report it normally.
+            return Err(CannotFetchThir);
```
jhjourdan, line 444: That's weird. Type checking is ran before this, isn't it? Then why are we here without all methods implemented?

arnaudgolfouse: No, the termination check is ran in `before_analysis`, and at this stage type checking is lazy.

jhjourdan: That's unfortunate, because the code can be malformed in many ways which we may not anticipate, which would trigger other kinds of ICEs like this one.

Isn't there a way to make sure you're checking is fully completed? Is there another callback hook that we could use?

arnaudgolfouse: We could try running in `after_analysis`, but I remember this causing issues as this needs to access thir? I'll try again just in case.

PR #1521, opened 2025-05-07
Improve `prove` options to open `why3 ide` and fix `cargo creusot why3 ide`

### LINKED ISSUE #1502: Following tutorial and using `cargo creusot why3 ide` ain't working
Followed the tutorial for setup and ran:`cargo creusot new pg && cd pg && cargo creusot prove && cargo creusot why3 ide`, which gives:

Image

Running on an M1 Mac with Sequioa 15.2. Everything else Creusot seems to be working (even the ide when running with `-i` mode).

Lysxia:
Close #1502.

This fixes the `why3 ide` command which was left unmaintained since I added support for why3find.

An important change is that the command now requires a coma file as an argument. Previously this wasn't necessary because the compiler generates a single output mlcfg file. This new requirement makes the command a bit unwieldly. Suggestions to improve it are welcome. I think the main mitigation is to nudge users away from the CLI and more towards editor-integrated tooling, that is Creusot IDE.


xldenis:
>  I think the main mitigation is to nudge users away from the CLI and more towards editor-integrated tooling, that is Creusot IDE.

Shamelessly tagging @sarsko since they told me they had difficulty getting the IDE installed and functional. I wonder if it might be time to start packaging the server binary with the extension? I know @giltho had some success doing something similar for their bug finding tool in written in caml.

sarsko:
Yeah, I wasn’t able to get the IDE plugin to work. I’ll try again once I’m at my laptop, and open an issue if I’m not able to get it workinf.

I agree that VS Code should be the way people use Creusot, but that requires the plugin install to work.

I like using the Why3 IDE to answer the question “am I missing something, or is it provable, but just really hard for the SMT solver?”. I don’t think that workflow will ever be supported in VS Code, though I think the “stay in the IDE, add proof_assert!s and lemmas” flow is overall faster.

giltho:
Hello,
Sorry for the late replay I was away end of last week.

In [Soteria](https://github.com/soteria-tools/soteria), we do something similar to [VeriFast](https://github.com/verifast/verifast). We create a [package](https://github.com/soteria-tools/soteria/blob/e11441cc8c95d8a8aa75bada7bfdaff94ee171f0/Makefile#L31-L36), which contains binaries for the tool itself and the solver, as well as all the `.dylib` files that are required, and install this package directly on the user machine. Then, the vscode plugin calls the executable with [the appropriate environment variables](https://github.com/soteria-tools/soteria/blob/e11441cc8c95d8a8aa75bada7bfdaff94ee171f0/vscode/src/install.ml#L32-L58). 
For now we only do it for macOS, and we have to build the package for intel and silicon separately. Something similar should be doable for Linux. Btw @xldenis, this is actually why I don't package the binaries directly with the extension, it's because which package you need depends on the architecture.

giltho:
Oh ok, I couldn't find the doc that explains this. When typing "architecture-specific vscode extension", the only thing I could find is [this page](https://code.visualstudio.com/api/working-with-extensions/bundling-extension), which says:

> The first reason to bundle your Visual Studio Code extension is to make sure it works for everyone using VS Code on any platform.

So I assumed extensions had to work everywhere, but I just read further and I see that bundling extensions is optional.

But still, I can't find any doc that explains how to do that hahaha. I'll just take a look at WhyCode, thanks!

xldenis:
https://code.visualstudio.com/api/working-with-extensions/publishing-extension#platformspecific-extensions

Lysxia:
Thanks for the tips! I can look into doing that for Creusot.

Issue #1517, opened 2025-05-07
Using `pearlite!` in an invariant affects what's provable (without adding `[requires(self.invariant())]`.


sarsko:
Hopefully the code below is self explanatory, but the `NoPearliteInv` doesn't wrap its invariant in `pearlite!`, which makes `this_passes` provable, wheares `PearliteInv` does wrap its invariant in `pearlite!`, resulting in `doesnt_pass` not passing unless we add `#[requires(self.invariant())] `. Also not clear to me why the coma generated is so different between the two instances.
```rust, creusot
use creusot_contracts::*;

struct NoPearliteInv {
    f1: Int,
    f2: Int,
}

impl Invariant for NoPearliteInv {
    #[predicate]
    #[open]
    fn invariant(self) -> bool {
        self.f1 == self.f2
    }
}

impl NoPearliteInv {
    #[ensures(self.f1 == self.f2)]
    fn this_passes(self) {}
}

struct PearliteInv {
    f1: Int,
    f2: Int,
}

impl Invariant for PearliteInv {
    #[predicate]
    #[open]
    fn invariant(self) -> bool {
        pearlite! {
            self.f1 == self.f2
        }
    }
}

impl PearliteInv {
    #[logic]
    #[ensures(self.f1 == self.f2)]
    fn doesnt_pass(self) {}

    #[logic]
    #[requires(self.invariant())] // Add this and it passes
    #[ensures(self.f1 == self.f2)]
    fn does_pass(self) {}
}

```

PR #1522, opened 2025-05-09
Document the behavior of type invariants in logic

arnaudgolfouse:
Following #1517, I realized we don't really document the behavior of type invariants in logic? So I whipped something up.

I also added a (somewhat shorter) explanation in the documentation of `creusot_contracts`.


jhjourdan:
Thanks!

I have to say that you are doing a lot of documentation job on Creusot, and this is very appreciable. Many thanks for all that (even if I don't explicitly say this at every PR).

PR #1531, opened 2025-05-15
Refactor generation Fn/FnMut/FnOnce specifications of closures and closure-like objects.

### LINKED ISSUE #1445: Postconditions generated for closure specifications do not verify the axiom they should verify
Postconditions of closures should verify three axioms:

```rust
    #[law]
    #[ensures(self.postcondition_mut(args, res_state, res) == (self == res_state && self.postcondition(args, res)))]
    fn fn_mut(self, args: Args, res_state: Self, res: Self::Output);

    #[law]
    #[ensures(self.postcondition_once(args, res) == (resolve(&self) && self.postcondition(args, res)))]
    fn fn_once(self, args: Args, res: Self::Output)
    where
        Self: Sized;

    #[law]
    #[trusted]
    #[ensures(self.postcondition_once(args, res) ==
              exists<res_state: Self> self.postcondition_mut(args, res_state, res) && resolve(&res_state))]
    fn fn_mut_once(self, args: Args, res: Self::Output) {}
```

I don't think that the postconditions we generate both in the "inferred" case and the "user" case verify these axioms in all cases.

In particular, there are missing calls to `resolve`, and it seems like there are some inconsistencies between the interpretation of ensures clauses in the `FnMut` and `FnOnce` cases, since variables that are not decorated with `old` correspond to the pre-state in one case and to the post-state in the other case.

### LINKED ISSUE #1452: `postcondition` and `precondition` don't work with function pointers
I currently can't do something like `iter.map(my_function)` though it should be _simpler_ in principle than a closure! All we should need to do is grab the contract clauses and stick them in the relevant definitions. Function pointers should always be `Fn` since they capture no state.

jhjourdan:
Fixes #1445, and partly fixes #1452. However true function pointers are not supported (yet?). There are some issues regarding termination that need to be addressed.&

A somewhat important improvement is that specification traits for `Default`, `From` and `IntoIterator` are no longer needed, because the functions from `FnExt` can be used instead.


jhjourdan:
Unfortunately, the coma files get quite larger, because of the laws of the `Fn`-like traits, that get pulled and pulls a lot of trivial useless things.. We need to design a better auto-loading mechanism...

creusot-contracts/src/std/option.rs
```diff
@@ -560,13 +560,26 @@ extern_spec! {
...
+    }
+
+    impl<'a, T> IntoIterator for &'a Option<T>{
+        #[ensures(*self == match result@ { None => None, Some(r) => Some(*r) })]
+        fn into_iter(self) -> Iter<'a, T>;
+    }
+
+    impl<'a, T> IntoIterator for &'a mut Option<T>{
+        #[ensures(*self == match result@ { None => None, Some(r) => Some(*r) })]
+        #[ensures(^self == match result@ { None => None, Some(r) => Some(^r) })]
```
xldenis, line 576: not the current subject either but, also would be nice to either: do some "logical reborrowing" here or more generally to be able to treat these kinds of methods as "pure" and infer their specifications. For `extern_specs` we could provide the body and do inference over it instead.

creusot-contracts/src/std/iter/map_inv.rs
```diff
@@ -35,15 +35,15 @@ impl<I: Iterator, B, F: FnMut(I::Item, Snapshot<Seq<I::Item>>) -> B> Iterator
...
     fn produces(self, visited: Seq<Self::Item>, succ: Self) -> bool {
         pearlite! {
-            self.func.unnest(succ.func)
+            self.func.hist_inv(succ.func)
             && exists<fs: Seq<&mut F>> fs.len() == visited.len()
             && exists<s : Seq<I::Item>> s.len() == visited.len() && self.iter.produces(s, succ.iter)
             && succ.produced.inner() == self.produced.concat(s)
             && (forall<i : Int> 1 <= i && i < fs.len() ==>  ^fs[i - 1] == * fs[i])
             && if visited.len() == 0 { self.func == succ.func }
                else { *fs[0] == self.func &&  ^fs[visited.len() - 1] == succ.func }
```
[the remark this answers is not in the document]
jhjourdan: What property are you speaking about?

creusot-contracts/src/std/option.rs
jhjourdan: Yeah, we should provide logical reborrowing under sums...

PR #1534, opened 2025-05-16
Add pointer specs

Lysxia:
Stuff needed to verify `swap` on slices.


creusot-contracts/src/std/ptr.rs
```diff
@@ -81,6 +81,62 @@ impl<T: ?Sized> PointerExt<T> for *mut T {
...
+    fn offset_logic(self, offset: Int) -> RawPtr<T> {
+        let _ = offset;
+        dead
+    }
+
+    // TODO: The offset in bytes, `count * size_of::<T>()`, must fit in an `isize`.
+    #[trusted]
+    #[requires(self.offset_logic(offset@) == own_offset.ptr())]
+    #[ensures(self.offset_logic(offset@) == result)]
+    unsafe fn add_own(self, offset: usize, own_offset: Ghost<&PtrOwn<T>>) -> Self {
```
Lysxia, line 117: The idea here, which may be wrong, is that the only way to prove `self.offset_logic(offset) == own_offset.ptr()` is if `self` actually comes from an allocation of an array of `T`, so all the pointers between `self` and `self.add(offset)` are in fact valid.

creusot-contracts/src/std/ptr.rs
```diff
@@ -117,3 +189,78 @@ extern_spec! {
...
+    #[ensures(result == self.right())]
+    pub fn right_ghost(&self) -> &PtrOwn<T> {
+        match self {
+            DisjointOrEqual::Equal(p) => p,
+            DisjointOrEqual::Disjoint(_, p) => p,
+        }
+    }
+
+    #[pure]
+    #[ensures(result == self.left())]
```
Lysxia, line 250: This is probably wrong. I can't prove this.

creusot-contracts/src/std/ptr.rs
jhjourdan: Well, we can imagine that we could compute offsets between fields of a struct, so arrays are not the only option here.

But I basically agree with you that this requirement should be enough. Basically, this correspond to assuming that the pointer type ghostly contains the identifier of the allocation: `offset_logic` would keep this provenance information, so that `self` and `own_offsef` would have the same provenance. So `own_offset` would guarantee that both pointers have the same block, and that this block has not been deallocated.

This matches the provenance semantics that Rust tries to adopt now.

But, in any case, this needs to be documented.

creusot-contracts/src/std/ptr.rs
```diff
@@ -81,6 +84,75 @@ impl<T: ?Sized> PointerExt<T> for *mut T {
...
+    #[open(self)]
+    fn offset_logic(self, offset: Int) -> RawPtr<T> {
+        let _ = offset;
+        dead
+    }
+
+    // TODO: The offset in bytes, `count * size_of::<T>()`, must fit in an `isize`.
+    #[trusted]
+    #[requires(self.offset_logic(offset@) == own_offset.ptr())]
+    // #[ensures(result as RawPtr<T> == self.offset_logic(offset))] // TODO: cast *mut to RawPtr ?
```
jhjourdan, line 150: Is this forbidden in Pearlite? We should allow these casts, and translate them by identity (using the Coerce construct).

creusot-contracts/src/std/ptr.rs
```diff
@@ -81,6 +81,62 @@ impl<T: ?Sized> PointerExt<T> for *mut T {
...
+    /// From https://doc.rust-lang.org/std/primitive.pointer.html#method.add:
+    ///
+    /// > If any of the following conditions are violated, the result is Undefined Behavior:
+    /// > - The offset in bytes, `count * size_of::<T>()`, computed on mathematical
+    /// >   integers (without “wrapping around”), must fit in an `isize`.
+    /// > - If the computed offset is non-zero, then `self` must be derived from a
+    /// >   pointer to some allocated object, and the entire memory range between
+    /// >   `self` and the result must be in bounds of that allocated object.
+    /// >   In particular, this range must not “wrap around” the edge of the address space.
+    unsafe fn add_own(self, offset: usize, own_offset: Ghost<&PtrOwn<T>>) -> Self;
```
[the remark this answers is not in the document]
Lysxia: Adding a `requires` errors with:

```
error: expected curly braces
   --> creusot-contracts/src/std/ptr.rs:105:82
    |
105 |     unsafe fn add_own(self, offset: usize, own_offset: Ghost<&PtrOwn<T>>) -> Self;
```

The problem is there is no body where to store the contract.

I think I'm going to just not have a spec there, and rely on having only specialized call sites.

creusot-contracts/src/logic/seq.rs
```diff
@@ -579,6 +579,20 @@ impl<T> Seq<T> {
...
+    /// Get mutable references to multiple elements in a sequence simultaneously.
+    #[trusted]
+    #[pure]
+    #[requires(forall<i: Int> 0 <= i && i < indices.len() ==> 0 <= indices[i] && indices[i] < self.len())]
+    #[requires(forall<i: Int, j: Int> 0 <= i && i < indices.len() && 0 <= j && j < indices.len() && i != j ==> indices[i] != indices[j])]
+    #[ensures(result.len() == indices.len())]
+    #[ensures(self.len() == (^self).len())]
+    #[ensures(forall<i: Int> 0 <= i && i < indices.len() ==> (*self)[indices[i]] == *result[i] && (^self)[indices[i]] == ^result[i])]
+    #[ensures(forall<i: Int> !indices.contains(i) ==> (*self).get(i) == (^self).get(i))]
+    pub fn get_mut_slice_ghost(&mut self, indices: Seq<Int>) -> Seq<&mut T> {
```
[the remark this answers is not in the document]
Lysxia: I ended up not needing this function so I will postpone adding it.

Lysxia:
TODO:

- [x] Handle `size_of`
- [x] Allow conversions in logic between `*mut` and `*const`
- [x] Finish the pointer specs and document their rationale

creusot-contracts/src/std/ptr.rs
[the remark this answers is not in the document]
Lysxia: nvm. We need to add a precondition so impls can refine it.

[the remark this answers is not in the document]
jhjourdan: That's definitely a bug, we do support specs on unimplemented trait methods. `bdd.rs` has an example of that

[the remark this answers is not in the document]
jhjourdan: Do I understand correctly that there is no issue finally?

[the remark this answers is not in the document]
Lysxia: That's right.

- There's a bug in the dummy macros that causes the `expected curly braces` error. It was not caught by the test suite because we don't run the test suite in dummy mode.
- I was working around that bug using an `extern_spec!`, but then I ran into another bug https://github.com/creusot-rs/creusot/pull/1536
- We can just fix that first bug. So there will be no issue.

creusot-contracts/src/std/ptr.rs
```diff
@@ -81,6 +84,75 @@ impl<T: ?Sized> PointerExt<T> for *mut T {
...
+impl<T> SizedPointerExt<T> for *const T {
+    #[trusted]
+    #[logic]
+    #[open(self)]
+    fn offset_logic(self, offset: Int) -> RawPtr<T> {
+        let _ = offset;
+        dead
+    }
+
+    // TODO: The offset in bytes, `count * size_of::<T>()`, must fit in an `isize`.
```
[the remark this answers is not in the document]
Lysxia: In the end this condition might not be necessary. It should be a consequence from having the `PtrOwn`. The `add` doc says:

> Allocated objects can never be larger than `isize::MAX` bytes, so if the computed offset stays in bounds of the allocated object, it is guaranteed to satisfy the first requirement. [= `count * size_of::<T>()`, must fit in an `isize`]

creusot-contracts/src/std/ptr.rs
Lysxia: One more thing: we want to allow one-past-the-end pointers. [`<*const T>::add` says this:](https://doc.rust-lang.org/std/primitive.pointer.html#method.add)

> `vec.as_ptr().add(vec.len())` (for `vec: Vec<T>`) is always safe.

I just ran into this while looking at `split_at_unchecked`.

We need a weaker form of evidence than `PtrOwn` to model this.

creusot-contracts/src/std/ptr.rs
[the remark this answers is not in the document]
jhjourdan: You're right. This spec is very weirdly written.

creusot-contracts/src/std/ptr.rs
jhjourdan: `#[requires(self.offset_logic(offset@) == own_offset.ptr() || self.offset_logic(offset@-1) == own_offset.ptr())]` ?

Lysxia: That is still too restrictive because you might not have access to the `PtrOwn` of the last element. In `split_at_unchecked` the `add` happens after the first slice is created, consuming all of its `PtrOwn`.

creusot-contracts/src/std/ptr.rs
```diff
@@ -81,6 +84,66 @@ impl<T: ?Sized> PointerExt<T> for *mut T {
...
+    unsafe fn add_own(self, offset: usize, own_offset: Ghost<&PtrOwn<T>>) -> Self {
+        self.add(offset)
+    }
+}
+
+impl<T> SizedPointerExt<T> for *mut T {
+    #[trusted]
+    #[logic]
+    #[open(self)]
+    #[ensures(self.addr_logic()@ + offset < usize::MAX@ ==> result.addr_logic()@ == self.addr_logic()@ + offset)]
```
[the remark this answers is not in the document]
jhjourdan: In addition, we should have an associativity lemma stating `p.offset_logic(x).offset_logic(y) = p.offset_logic(x+y)`.

creusot-contracts/src/std/ptr.rs
Lysxia: I'm experimenting with a `SliceOwn` token, a variant of `PtrOwn`, that contains information about allowed offsets. In particular we can create zero-sized `SliceOwn` that point one-past-the-end of an allocated slice.

I think that adding such information in the pointers themselves is not a solution because the range of possible offsets depends on if the allocated object has been freed:

> it’s always UB to offset a pointer derived from something that is now deallocated, except if the offset is 0. --- https://doc.rust-lang.org/std/ptr/index.html#provenance

Does it makes sense to be able to convert `&SliceOwn` into `&PtrOwn` of each element, and conversely, to convert `&PtrOwn` into a singleton `&SliceOwn`? (And of course, without the `&` this won't be allowed.) It's certainly convenient to reuse `PtrOwn`'s method to dereference pointers, but I wonder if there may be some extra information in a `&PtrOwn` that can't come from `&SliceOwn`.

jhjourdan: How having a `SliceOwn` for a type `T` would be different to having a `PtrOwn` for `[T]`?

It really seems bad to me to have two different token types for owning a single slot and owning a slice. For example, how would we offset into an which is stored in a struct, of which we only have a `PtrOwn` (I assume that in your proposal, it is impossible to split these ownership tokens, because that would essentially revert to the original proposal)? 

To synthesize my current view of things:

On the one hand, we need to be able to split ownership, both because it is much easier to specify reading/writing from a raw pointer, and because we can very much imagine that a large block of memory is split into smaller thunks, themselves used in very different part of a program (different threads...). Being able to split tokens is crucial for proving things like `Bumpalo` (I'm not saying that you should prove it, be not having the tools to do it seems like a red flag to me). 

On the other hand, in order to guarantee that some offsets are allowed, and to guarantee that we never partially deallocate a block by giving back only some of its tokens, we need another token that says "this block is allocated". This other token could not be split and would not provide read/write access to anything, but a shared borrow of it would be needed to offset pointers and full ownership of it would be needed to deallocate memory.

Lysxia:
With this batch of methods I've reached the milestone of verifying 7/10 of the unsafe slice methods (the "easy" ones) from the verify-rust-std challenge.

creusot-contracts/src/ghost/ptr_own.rs
```diff
@@ -234,17 +228,174 @@ impl<T: ?Sized> PtrOwn<T> {
...
+    }
+
+    /// Split a `&PtrOwn<[T]>` into two subslices of lengths `index` and `self.len() - index`.
+    #[trusted]
+    #[check(ghost)]
+    #[requires(0 <= index && index <= self.len())]
+    #[ensures(self.ptr().thin() == result.0.ptr().thin() && self.ptr().thin().offset_logic(index) == result.1.ptr().thin())]
+    #[ensures(index == result.0.len() && self.len() - index == result.1.len())]
+    #[ensures(forall<k: Int> 0 <= k && k < index ==> self.val()@[k] == result.0.val()@[k])]
+    #[ensures(forall<k: Int> index <= k && k < self.len() ==> self.val()@[k] == result.1.val()@[k - index])]
```
jhjourdan: These lines are too long.

jhjourdan:
(I have put reviewing this on my TODO list, but I am rather busy these days...)

creusot-contracts/src/ghost/ptr_own.rs
```diff
@@ -234,17 +228,174 @@ impl<T: ?Sized> PtrOwn<T> {
...
+
+    /// Split a `&mut PtrOwn<[T]>` into two subslices of lengths `index` and `self.len() - index`.
+    #[trusted]
+    #[check(ghost)]
+    #[requires(0 <= index && index <= self.len())]
+    #[ensures(self.ptr().thin() == result.0.ptr().thin() && self.ptr().thin().offset_logic(index) == result.1.ptr().thin())]
+    #[ensures(index == result.0.len() && self.len() - index == result.1.len())]
+    #[ensures(index == (^result.0).len() && self.len() - index == (^result.1).len())]
+    #[ensures(forall<k: Int> 0 <= k && k < index ==> self.val()@[k] == result.0.val()@[k] && (^self).val()@[k] == (^result.0).val()@[k])]
+    #[ensures(forall<k: Int> index <= k && k < self.len() ==> self.val()@[k] == result.1.val()@[k - index] && (^self).val()@[k] == (^result.1).val()@[k - index])]
```
jhjourdan: Idem

creusot-contracts/src/ghost/ptr_own.rs
jhjourdan: For these specs, it's probably better to state them in terms of `Seq::subsequence`. Support of sequences by Why3 is not ideal, but it may improve in the future, and it does anyway contain such quantified specs.

creusot-contracts/src/ghost/ptr_own.rs
jhjourdan: All the post-conditions about the prophecies are unsound, because one can write a different `PtrOwn<[T]>` into the returned borrow. You should guard these post-conditions by a condition making sure that the returned `ptr()` have not changed.

creusot-contracts/src/ghost/ptr_own.rs
```diff
@@ -234,17 +228,174 @@ impl<T: ?Sized> PtrOwn<T> {
...
+
+    /// Split `&mut PtrOwn<[T]>` into a sequence of `&mut PtrOwn<T>` for each element.
+    #[trusted]
+    #[check(ghost)]
+    #[ensures(result.len() == self.len())]
+    #[ensures(forall<i> 0 <= i && i < self.len()
+        ==> result[i].ptr() == self.ptr().thin().offset_logic(i)
+        && *result[i].val() == self.val()@[i]
+        && (^self).ptr() == self.ptr()
+        && *(^result[i]).val() == (^self).val()@[i])]
```
jhjourdan:
```suggestion
    #[ensures(forall<i> 0 <= i && i < self.len()
        ==> result[i].ptr() == self.ptr().thin().offset_logic(i)
        && *result[i].val() == self.val()@[i]]
    #[ensures(forall<i> 0 <= i && i < self.len()
        && (^self).ptr() == self.ptr()
        ==> *(^result[i]).val() == (^self).val()@[i])]
```

Otherwise, it's unsound (one can write into one of the returned borrow.

creusot-contracts/src/ghost/ptr_own.rs
```diff
@@ -234,17 +228,174 @@ impl<T: ?Sized> PtrOwn<T> {
...
+
+    /// Split `&PtrOwn<[T]>` into a sequence of `&PtrOwn<T>` for each element.
+    #[trusted]
+    #[check(ghost)]
+    #[ensures(result.len() == self.len())]
+    #[ensures(forall<i> 0 <= i && i < self.len()
+        ==> result[i].ptr() == self.ptr().thin().offset_logic(i)
+        && *result[i].val() == self.val()@[i])]
+    pub fn elements(&self) -> Seq<&PtrOwn<T>> {
+        panic!("called ghost function in normal code")
```
jhjourdan: It seems to me that we should be able to easily implement this function by using `split_at_ghost` and doing a recursive call.

creusot-contracts/src/ghost/ptr_own.rs
```diff
@@ -234,17 +228,174 @@ impl<T: ?Sized> PtrOwn<T> {
...
+    #[trusted]
+    #[check(ghost)]
+    #[ensures(result.len() == self.len())]
+    #[ensures(forall<i> 0 <= i && i < self.len()
+        ==> result[i].ptr() == self.ptr().thin().offset_logic(i)
+        && *result[i].val() == self.val()@[i]
+        && (^self).ptr() == self.ptr()
+        && *(^result[i]).val() == (^self).val()@[i])]
+    pub fn elements_mut(&mut self) -> Seq<&mut PtrOwn<T>> {
+        panic!("called ghost function in normal code")
```
jhjourdan: Idem, we should be able to easily implement this instead of trusting it.

creusot-contracts/src/ghost/ptr_own.rs
```diff
@@ -234,17 +228,174 @@ impl<T: ?Sized> PtrOwn<T> {
...
+        r.split_off_ghost(index).pop_front_ghost().unwrap()
+    }
+
+    /// Convert a `&mut PtrOwn<[T]>` for a non-empty slice into a `&mut PtrOwn<T>` for the element at the given index.
+    #[check(ghost)]
+    #[requires(0 <= index && index < self.len())]
+    #[ensures(result.ptr() == self.ptr().thin().offset_logic(index))]
+    #[ensures(*result.val() == self.val()@[index])]
+    #[ensures(*(^result).val() == (^self).val()@[index])]
+    #[ensures((^self).ptr() == self.ptr())]
```
jhjourdan:
```suggestion
    #[ensures((^self).ptr() == self.ptr() ==> *(^result).val() == (^self).val()@[index])]
```

creusot-contracts/src/ghost/ptr_own.rs
```diff
@@ -234,17 +228,174 @@ impl<T: ?Sized> PtrOwn<T> {
...
+    #[ensures(forall<k: Int> k != index ==> (^self).val()@.get(k) == self.val()@.get(k))]
+    pub fn index_ptr_own_mut_ghost(&mut self, index: Int) -> &mut PtrOwn<T> {
+        let mut r = self.elements_mut();
+        r.split_off_ghost(index).pop_front_ghost().unwrap()
+    }
+
+    /// Convert `&PtrOwn<T>` to `&PtrLive`
```
jhjourdan:
```suggestion
    /// Convert `&PtrOwn<[T]>` to `&PtrLive<T>`
```

creusot-contracts/src/ghost/ptr_own.rs
```diff
@@ -234,17 +228,174 @@ impl<T: ?Sized> PtrOwn<T> {
...
+///
+/// This evidence enables taking pointer offsets (see [`add_own`](crate::std::intrinsics::add_own) and others)
+/// without ownership (via [`PtrOwn`]) of that range of memory.
+///
+/// It only ever exists as `&PtrLive<T>`: this resource can be duplicated freely,
+/// but its lifetime is bounded by some `&PtrOwn<[T]>` (via `PtrOwn::live`)
+/// or `&mut PtrOwn<[T]>` (via `PtrOwn::live_mut`).
+#[allow(dead_code)]
+#[opaque]
+pub struct PtrLive<T>(PhantomData<T>);
```
jhjourdan: Instead of only allowing this type to exist under a `&'a`, why not making it `Copy` and annotated with a lifetime?


```suggestion
/// Its lifetime is bounded by some `&PtrOwn<[T]>` (via `PtrOwn::live`)
/// or `&mut PtrOwn<[T]>` (via `PtrOwn::live_mut`).
#[allow(dead_code)]
#[opaque]
#[derive(Clone, Copy)]
pub struct PtrLive<'a, T>(PhantomData<&'a T>);
```

creusot-contracts/src/ghost/ptr_own.rs
```diff
@@ -234,17 +228,174 @@ impl<T: ?Sized> PtrOwn<T> {
...
+    pub fn len(self) -> usize {
+        dead
+    }
+
+    /// The live range contains `ptr`
+    #[logic(open, inline)]
+    pub fn contains(self, ptr: *const T) -> bool {
+        pearlite! {
+            ptr == self.ptr().offset_logic(ptr.sub_logic(self.ptr()))
+            && 0 <= ptr.sub_logic(self.ptr()) && ptr.sub_logic(self.ptr()) <= self.len()@
```
jhjourdan:
```suggestion
            let offset = ptr.sub_logic(self.ptr());
            ptr == self.ptr().offset_logic(offset)
            && 0 <= offset && offset <= self.len()@
```

creusot-contracts/src/ghost.rs
```diff
@@ -26,7 +26,7 @@ pub mod local_invariant;
 mod ptr_own;
 pub mod resource;
 pub use fn_ghost::{FnGhost, FnGhostWrapper};
-pub use ptr_own::PtrOwn;
+pub use ptr_own::{PtrLive, PtrOwn};
```
jhjourdan: Used?

creusot-contracts/src/std/intrinsics.rs
```diff
@@ -1,4 +1,103 @@
...
+///
+/// # Safety
+///
+/// Source: <https://doc.rust-lang.org/core/intrinsics/fn.offset.html>
+///
+/// > If the computed offset is non-zero, then both the starting and resulting pointer must be either in bounds or at the end of an allocation.
+/// > If either pointer is out of bounds or arithmetic overflow occurs then this operation is undefined behavior.
+///
+/// - We ignore the non-zero condition for simplicity (so the precondition is a little stronger than necessary).
+/// - The preconditions ensure that the ownership `own` contains the range between `dst` and `dst + offset`,
```
jhjourdan:
```suggestion
/// - The preconditions ensure that the `live` witness contains the range between `dst` and `dst + offset`,
```

creusot-contracts/src/std/intrinsics.rs
```diff
@@ -1,4 +1,103 @@
...
+///   which also implies that arithmetic overflow does not occur.
+#[cfg(feature = "nightly")]
+#[trusted]
+#[erasure(core::intrinsics::offset)]
+#[requires(live.contains(dst))]
+#[requires(live.contains(dst.offset_logic(offset@)))]
+#[ensures(result == dst.offset_logic(offset@))]
+pub unsafe fn add_own<T>(dst: *const T, offset: usize, live: Ghost<&PtrLive<T>>) -> *const T {
```
jhjourdan:
```suggestion
pub unsafe fn add_live<T>(dst: *const T, offset: usize, live: Ghost<&PtrLive<T>>) -> *const T {
```

creusot-contracts/src/std/intrinsics.rs
```diff
@@ -1,4 +1,103 @@
...
+/// - [`<*mut T>::offset`](https://doc.rust-lang.org/core/primitive.pointer.html#method.offset-1)
+///
+/// # Safety
+///
+/// Source: <https://doc.rust-lang.org/core/intrinsics/fn.offset.html>
+///
+/// > If the computed offset is non-zero, then both the starting and resulting pointer must be either in bounds or at the end of an allocation.
+/// > If either pointer is out of bounds or arithmetic overflow occurs then this operation is undefined behavior.
+///
+/// - We ignore the non-zero condition for simplicity (so the precondition is a little stronger than necessary).
```
jhjourdan: IMO, that's a niche optimization and you don't gain anything from ignoring that.

creusot-contracts/src/std/intrinsics.rs
```diff
@@ -1,4 +1,103 @@
...
+/// - The preconditions ensure that the ownership `own` contains the range between `dst` and `dst + offset`,
+///   which also implies that arithmetic overflow does not occur.
+#[cfg(feature = "nightly")]
+#[trusted]
+#[erasure(core::intrinsics::offset)]
+#[requires(live.contains(dst))]
+#[requires(live.contains(dst.offset_logic(offset@)))]
```
jhjourdan:
```suggestion
#[requires(offset@ == 0 || live.contains(dst))]
#[requires(offset@ == 0 || live.contains(dst.offset_logic(offset@)))]
```

creusot-contracts/src/std/intrinsics.rs
```diff
@@ -1,4 +1,103 @@
...
+/// [`<*const T>::offset`](https://doc.rust-lang.org/core/primitive.pointer.html#method.offset) with `PtrOwn` permissions.
+///
+/// See also [`add_own`]. This function is the variant of `add_own` with `isize` offsets.
+#[cfg(feature = "nightly")]
+#[trusted]
+#[erasure(core::intrinsics::offset)]
+#[requires(live.contains(dst))]
+#[requires(live.contains(dst.offset_logic(offset@)))]
+#[ensures(result == dst.offset_logic(offset@))]
+pub unsafe fn offset_own<T>(dst: *const T, offset: isize, live: Ghost<&PtrLive<T>>) -> *const T {
```
jhjourdan: Same remarks.

creusot-contracts/src/std/intrinsics.rs
```diff
@@ -1,4 +1,103 @@
...
+/// [`<*mut T>::add`](https://doc.rust-lang.org/core/primitive.pointer.html#method.add-1) with `PtrOwn` permissions.
+///
+/// See also [`add_own`]. This function is the variant of `add_own` with `*mut` pointers.
+#[cfg(feature = "nightly")]
+#[trusted]
+#[erasure(core::intrinsics::offset)]
+#[requires(live.contains(dst))]
+#[requires(live.contains(dst.offset_logic(offset@)))]
+#[ensures(result as *const T == (dst as *const T).offset_logic(offset@))]
+pub unsafe fn add_own_mut<T>(dst: *mut T, offset: usize, live: Ghost<&PtrLive<T>>) -> *mut T {
```
jhjourdan: Same remarks.

creusot-contracts/src/std/intrinsics.rs
```diff
@@ -1,4 +1,103 @@
...
+/// [`<*mut T>::offset`](https://doc.rust-lang.org/core/primitive.pointer.html#method.offset-1) with `PtrOwn` permissions.
+///
+/// See also [`add_own`]. This function is the variant of `add_own` with `*mut` pointers and `isize` offsets.
+#[cfg(feature = "nightly")]
+#[trusted]
+#[erasure(core::intrinsics::offset)]
+#[requires(live.contains(dst))]
+#[requires(live.contains(dst.offset_logic(offset@)))]
+#[ensures(result as *const T == (dst as *const T).offset_logic(offset@))]
+pub unsafe fn offset_own_mut<T>(dst: *mut T, offset: isize, live: Ghost<&PtrLive<T>>) -> *mut T {
```
jhjourdan: Same remarks.

creusot-contracts/src/ghost/ptr_own.rs
Lysxia: Not quite, because there is no other function from `PtrOwn<[T]>` to `PtrOwn<T>`. Given `split_at_ghost`, the following are equivalent:

1. `&PtrOwn<[T]> -> Seq<&PtrOwn<T>>` (this `elements`)
2. `&PtrOwn<[T]> -> &PtrOwn<T>` (that gets only the first element)

It seems arbitrary whichever comes first, so I chose the one that avoids writing a recursive function for the other one.

jhjourdan: Yeah, ok, fair enough.

creusot-contracts/src/ghost/ptr_own.rs
Lysxia: Good catch. ~~I assume the last implication is meant to be `(^result[i]).ptr() == result[i].ptr() ==> *(^result[i]).val() == (^self).val()@[i]`~~

Lysxia: Ah no my bad. I'm getting confused.

jhjourdan: Me too. It seems like I'm missing something. We should not request anything about `^self`.

jhjourdan:
```suggestion
    #[ensures(forall<i> 0 <= i && i < self.len()
        ==> result[i].ptr() == self.ptr().thin().offset_logic(i)
        && *result[i].val() == self.val()@[i])]
    #[ensures((forall<i> 0 <= i && i < self.len() && result[i].ptr() == (^result[i]).ptr())
        ==> (^self).ptr() == self.ptr()
        && forall<i> 0 <= i && i < self.len() *(^result[i]).val() == (^self).val()@[i])]
```

(Surprisingly subtle...)

jhjourdan:
```suggestion
    #[ensures(forall<i> 0 <= i && i < self.len()
        ==> result[i].ptr() == self.ptr().thin().offset_logic(i)
        && *result[i].val() == self.val()@[i])]
    #[ensures((^self).ptr() == self.ptr())]
    #[ensures((forall<i> 0 <= i && i < self.len()
        && (^result[i]).ptr() == result[i].ptr() 
        ==> *(^result[i]).val() == (^self).val()@[i])]
```

Actually, the lifetime of the returned borrow should be enough to ensure that `#[ensures((^self).ptr() == self.ptr())]` is sound.

creusot-contracts/src/ghost/ptr_own.rs
jhjourdan: I did the same mistake here....


```suggestion
    #[ensures((^self).ptr() == self.ptr())]
    #[ensures((^result).ptr() == result.ptr() ==> *(^result).val() == (^self).val()@[index])]
```

creusot-contracts/src/ghost.rs
Lysxia: Reexported (`ptr_own` is private).

creusot-contracts/src/ghost/ptr_own.rs
```diff
@@ -234,17 +229,194 @@ impl<T: ?Sized> PtrOwn<T> {
...
+    /// it can be used even during the lifetime of the `&PtrLive`.
+    #[trusted]
+    #[check(ghost)]
+    #[ensures(*result.0 == *self)]
+    #[ensures(result.1.ptr() == self.ptr().thin())]
+    #[ensures(result.1.len()@ == self.len())]
+    #[ensures((^self).ptr() == self.ptr())]
+    // NOTE: If the caller swaps `result.0` with another token, there are no guarantees about the final value.
+    #[ensures((^result.0).ptr() == result.0.ptr() ==> (^result.0).val() == (^self).val())]
+    pub fn live_mut(&mut self) -> (&mut Self, &PtrLive<'_, T>) {
```
jhjourdan:
```suggestion
    pub fn live_mut(&mut self) -> (&mut Self, PtrLive<T>) {
```

No need to use borrows of `PtrLive`, and (I think that) `'_, ` is implicit.

creusot-contracts/src/ghost/ptr_own.rs
```diff
@@ -234,17 +229,194 @@ impl<T: ?Sized> PtrOwn<T> {
...
+        let mut r = self.elements_mut();
+        r.split_off_ghost(index).pop_front_ghost().unwrap()
+    }
+
+    /// Convert `&PtrOwn<[T]>` to `&PtrLive<T>`
+    #[trusted]
+    #[check(ghost)]
+    #[ensures(result.ptr() == self.ptr().thin())]
+    #[ensures(result.len()@ == self.len())]
+    pub fn live(&self) -> &PtrLive<'_, T> {
```
jhjourdan:
```suggestion
    pub fn live(&self) -> PtrLive<T> {
```

creusot-contracts/src/ghost/ptr_own.rs
```diff
@@ -234,17 +229,194 @@ impl<T: ?Sized> PtrOwn<T> {
...
     }
+}
+
+/// # Permissions for slice pointers
+///
+/// Core methods:
+///
+/// - To split a `&PtrOwn<[T]>`: [`split_at_ghost`](PtrOwn::split_at_ghost), [`split_at_mut_ghost`](PtrOwn::split_at_mut_ghost).
+/// - To index a `&PtrOwn<[T]>` into `&PtrOwn<T>`: [`elements`](PtrOwn::elements), [`elements_mut`](PtrOwn::elements_mut).
+/// - To extract a [`&PtrLive<T>`][PtrLive] (evidence used by pointer arithmetic): [`live`](PtrOwn::live), [`live_mut`](PtrOwn::live_mut).
```
jhjourdan:
```suggestion
/// - To extract a [`PtrLive<T>`][PtrLive] (evidence used by pointer arithmetic): [`live`](PtrOwn::live), [`live_mut`](PtrOwn::live_mut).
```

creusot-contracts/src/std/intrinsics.rs
```diff
@@ -1,4 +1,102 @@
...
+///
+/// The preconditions ensure that the `live` witness contains the range between `dst` and `dst + offset`,
+/// which prevents out-of-bounds access and overflow.
+#[cfg(feature = "nightly")]
+#[trusted]
+#[erasure(core::intrinsics::offset)]
+#[requires(offset@ == 0 || live.contains(dst))]
+#[requires(offset@ == 0 || live.contains(dst.offset_logic(offset@)))]
+#[ensures(result == dst.offset_logic(offset@))]
+pub unsafe fn add_live<T>(dst: *const T, offset: usize, live: Ghost<&PtrLive<T>>) -> *const T {
```
jhjourdan:
```suggestion
pub unsafe fn add_live<T>(dst: *const T, offset: usize, live: Ghost<PtrLive<T>>) -> *const T {
```

creusot-contracts/src/ghost/ptr_own.rs
Lysxia: The problem goes deeper: you can `std::mem::swap` an `&mut PtrOwn<T>`  with another token that you own (e.g., with `from_box`), to get full ownership of the token previously under `&mut`. It looks like `&mut PtrOwn<T>` should be replaced with yet another type `PtrMut<'a, T>` to ensure that the token does not outlive the mutable borrow it comes from.

jhjourdan: Actually, that looks like a job for `Pin`, doesn't it?

Lysxia: Looks like it! Then we can also remove the "still the same pointer" check, right?

jhjourdan: Yes. I have to say that this is an idea that I just had I have not analyzed all the implications. If that works, then it should be used for resource algebra's split_mut, where we essentially have the same problem. Cc @arnaudgolfouse.

jhjourdan: Wait. There is [`Pin::set`](https://doc.rust-lang.org/std/pin/struct.Pin.html#method.set)...

creusot-contracts/src/ghost/ptr_own.rs
Lysxia: There's a warning that it needs to be `'_` when it also occurs implicitly in other `&`.

creusot-contracts/src/ghost/ptr_own.rs
```diff
@@ -235,16 +210,316 @@ impl<T: ?Sized> PtrOwn<T> {
...
+
+/// Permission token for mutable access to a pointer.
+///
+/// This is used by functions that return mutable borrows of tokens (e.g., `from_mut`, `split_at_mut`, `elements_mut`).
+/// If we returned `&mut PtrOwn<T>`, that would allow the caller to move the `PtrOwn<T>` out of the borrow
+/// and keep it even after the end of the borrow.
+/// Instead, we return `&'a mut PtrMut<'a, T>` so that the token cannot outlive the borrow.
+#[allow(dead_code)]
+#[opaque]
+pub struct PtrMut<'a, T: ?Sized>(PhantomData<&'a T>);
```
jhjourdan: I don't like the idea of having two different tokens here.

Could we make `PtrOwn<T>` an alias for `PtrMut<'static, T>`?

jhjourdan: (Then, `PtrMut` may not be the best name for this...)

Lysxia:
~~To merge after #1849~~ Reverted

creusot-contracts/src/ghost/ptr_own.rs
```diff
@@ -89,7 +88,8 @@ impl<T> PtrOwn<T> {
     #[check(ghost)]
     #[requires(size_of_logic::<T>() != 0)]
     #[ensures(own1.ptr().addr_logic() != own2.ptr().addr_logic())]
-    #[ensures(*own1 == ^own1)]
+    #[ensures(own1.ptr() == (^own1).ptr())]
+    #[ensures(own1.val() == (^own1).val())]
```
jhjourdan: Why this change?

Lysxia: I'll revert it. It's a remainder of refactoring back and forth.

Lysxia:
I think this has reached a pretty stable and mergeable state.

I'm quite confident about these specs because of how much time we've spent refining them, though I still wonder if there's some way to verify or test them, short of getting into a big Rocq project like RustHornBelt.

jhjourdan:
Does this mean that we can make a submission on Amazon's challenge?

creusot-std/src/std/intrinsics.rs
```diff
@@ -1,4 +1,98 @@
...
+}
+
+/// [`<*mut T>::add`](https://doc.rust-lang.org/core/primitive.pointer.html#method.add-1) with `PtrLive` permissions.
+///
+/// See also [`add_live`]. This function is the variant of `add_live` with `*mut` pointers.
+#[cfg(feature = "nightly")]
+#[trusted]
+#[erasure(core::intrinsics::offset)]
+#[requires(offset@ == 0 || live.contains(dst) && live.contains(dst.offset_logic(offset@)))]
+#[ensures(result as *const T == (dst as *const T).offset_logic(offset@))]
```
jhjourdan: May we have a transparent, inline wrapper of `offset_logic` for mut pointers?

creusot-std/src/std/intrinsics.rs
```diff
@@ -1,4 +1,98 @@
...
+/// > If either pointer is out of bounds or arithmetic overflow occurs then this operation is undefined behavior.
+///
+/// The preconditions ensure that the `live` witness contains the range between `dst` and `dst + offset`,
+/// which prevents out-of-bounds access and overflow.
+#[cfg(feature = "nightly")]
+#[trusted]
+#[erasure(core::intrinsics::offset)]
+#[requires(offset@ == 0 || live.contains(dst) && live.contains(dst.offset_logic(offset@)))]
+#[ensures(result == dst.offset_logic(offset@))]
+pub unsafe fn add_live<T>(dst: *const T, offset: usize, live: Ghost<PtrLive<T>>) -> *const T {
```
jhjourdan: Why doesn't Rustc complain about `live` being not used?

Lysxia:
The challenge isn't complete yet. Among what's left, I need to deal with `MaybeUninit` and transmutation, which I'm planning to do in separate PR(s) (my plan is to come up with some high-level axioms to get it done and then see if they ask us to break them down).

On Amazon's side, we're also still stuck on the first step of getting Creusot accepted as a tool. When I made the submission I asked for clarification about the licensing requirements and they said they'd ask their lawyers and I haven't heard back from them. I'll try to ping them again.

creusot-std/src/std/intrinsics.rs
```diff
@@ -18,6 +112,10 @@ extern_spec! {
                     unsafe { core::intrinsics::unreachable() }
                 }
             }
+
+            // `core::intrinsics::arith_offset` has no safety requirements.
+            #[ensures(result == dst.offset_logic(offset@))]
+            unsafe fn arith_offset<T>(dst: *const T, offset: isize) -> *const T;
```
jhjourdan: There is something weird happening with overflows and `offset_logic`. The current situation might be unsound, I am not sure.

What is the semantics of `offset_logic` when there is an overflow?
- Either it has a wraparound semantics, and then the spec of `add_live` and friends need to be fixed.
- Or it return an "overflowing pointer", which cannot exist at runtime, and does not belong to any live set.

The spec of the `arith_offset` function seems to imply that we chose the wraparound semantics. This is problematic, because the precondition of `add_live` and friends seem to chose the "overflowing" semantics.

creusot-std/src/std/intrinsics.rs
```diff
@@ -1,4 +1,98 @@
...
+///
+/// > If the computed offset is non-zero, then both the starting and resulting pointer must be either in bounds or at the end of an allocation.
+/// > If either pointer is out of bounds or arithmetic overflow occurs then this operation is undefined behavior.
+///
+/// The preconditions ensure that the `live` witness contains the range between `dst` and `dst + offset`,
+/// which prevents out-of-bounds access and overflow.
+#[cfg(feature = "nightly")]
+#[trusted]
+#[erasure(core::intrinsics::offset)]
+#[requires(offset@ == 0 || live.contains(dst) && live.contains(dst.offset_logic(offset@)))]
```
jhjourdan: Does this mean we don't support off-by-one pointers?

creusot-std/src/std/ptr.rs
```diff
@@ -621,17 +615,185 @@ impl<T: ?Sized> Perm<*const T> {
...
+    pub fn index_mut(&mut self, index: Int) -> &mut Perm<*const T> {
+        let mut r = self.elements_mut();
+        proof_assert! { forall<k> index < k && k < r.len() ==> r[k].val() == r[index..].tail()[k-index-1].val() };
+        let r0 = snapshot! { r };
+        let result = r.split_off_ghost(index).pop_front_ghost().unwrap();
+        proof_assert! { forall<i> 0 <= i && i < index ==> r[i] == r0[i] }; // Unfolding of ensures of split_off_ghost r == r0[..index]
+        result
+    }
+
+    /// Extract `PtrLive<'a, T>` from `&Perm<'a, [T]>`.
```
jhjourdan:
```suggestion
    /// Extract `PtrLive<'a, T>` from `&'a Perm<[T]>`.
```

creusot-std/src/std/ptr.rs
```diff
@@ -621,17 +615,185 @@ impl<T: ?Sized> Perm<*const T> {
...
-    /// Opaque wrapper around [`std::ptr::is_aligned_logic`].
-    /// We use this to hide alignment logic by default in `invariant` because it confuses SMT solvers sometimes.
-    /// The underlying property is exposed by [`Perm::ptr_is_aligned_lemma`].
-    #[logic(open(self))]
-    pub fn ptr_is_aligned_opaque(self) -> bool {
-        self.ward().is_aligned_logic()
+    #[ensures(result.0 == self)]
+    #[ensures(result.1.ward() == self.ward().thin())]
+    #[ensures(result.1.len()@ == self.len())]
+    pub fn live_mut(&mut self) -> (&mut Self, PtrLive<'_, T>) {
```
jhjourdan:
```suggestion
    pub fn live_mut<'a, 'b>(&'a &'b mut self) -> PtrLive<'b, T> {
```

The fact that we are using a mutable borrow under a shared one seems a bit weird, but this really seems to be what we mean I think.

creusot-std/src/std/intrinsics.rs
Lysxia: `contains` includes the off-by-one pointer at the end of the range.

creusot-std/src/std/ptr.rs
Lysxia: 😲

creusot-std/src/std/intrinsics.rs
Lysxia: I was thinking "overflowing pointer" because I don't want to think about wraparound for `add_live`, but you may be right that then this spec looks dubious.

creusot-std/src/std/intrinsics.rs
Lysxia: ~~`ensures`/`requires` add `allow(unused)` on each argument.~~ (does not apply in rustc mode)

~~Actually I'm not sure why.~~

I added a blanket `allow(unused)` in `lib.rs` at some point...

Fixed in #1893

creusot-std/src/std/intrinsics.rs
Lysxia: In the end I don't even need this so I can remove it. I just added it because I thought its spec would be easy but that seems not to be the case.

jhjourdan: > I was thinking "overflowing pointer" because I don't want to think about wraparound for `add_live`, but you may be right that then this spec looks dubious.

Ok, then perhpas we could remove the requires clause from the spec of `offset_logic`, and simply state as a postcondition that the returned address is the sum of machine integers.

jhjourdan: > In the end I don't even need this so I can remove it. I just added it because I thought its spec would be easy but that seems not to be the case.

I'm not against this. That said, it does not seem to be so difficult to specify that the addition on addresses does not overflow.

creusot-std/src/std/intrinsics.rs
jhjourdan: Oh, interesting.

creusot-std/src/std/intrinsics.rs
```diff
@@ -1,4 +1,102 @@
...
```
Lysxia: I think we can actually omit the `offset@ == 0` condition, because when you use this you need a `PtrLive` anyway, and in practice, it is going to be valid.

For the `offset@ == 0` condition to be effective, this function would need to either accept `Ghost<Option<PtrLive>>` or we should export a dummy `Ghost<PtrLive>`. Either way this complicates the API and we currently don't have a use case to evaluate which of these choices is better.

creusot-std/src/std/intrinsics.rs
```diff
@@ -1,4 +1,102 @@
...
+/// Source: <https://doc.rust-lang.org/core/intrinsics/fn.offset.html>
+///
+/// > If the computed offset is non-zero, then both the starting and resulting pointer must be either in bounds or at the end of an allocation.
+/// > If either pointer is out of bounds or arithmetic overflow occurs then this operation is undefined behavior.
+///
+/// The preconditions ensure that the `live` witness contains the range between `dst` and `dst + offset`,
+/// which prevents out-of-bounds access and overflow.
+#[cfg(feature = "nightly")]
+#[trusted]
+#[erasure(core::intrinsics::offset)]
```
Lysxia: Maybe the erasure of this should be `<*const T>::add`, i.e., the stable version of this intrinsic, which is therefore more useful for most users.

`core::intrinsics::offset`'s main (only?) use is to implement `<*const T>::add` and other stable methods. That makes this unstable intrinsic only relevant to the verify-rust-std project.

creusot-std/src/std/ptr.rs
```diff
@@ -623,17 +670,277 @@ impl<T: ?Sized> Perm<*const T> {
...
+    #[ensures(result.len()@ == self.len())]
+    pub fn live(&self) -> PtrLive<'_, T> {
+        panic!("called ghost function in normal code")
+    }
+}
+
+/// Extension trait with the `live_mut` method
+///
+/// Naively, we wanted `live_mut(&'b &'a mut self) -> PtrLive<'a, T>` where `Self = Perm<*const [T]>`,
+/// but `self` being a double borrow is illegal.
```
jhjourdan: I think it is legal. It just has a peculiar syntax: 


```rust
    fn live_mut<'b>(self: &'b &'a mut Self) -> PtrLive<'a, T>;
```

creusot-std/src/std/ptr.rs
```diff
@@ -623,17 +670,277 @@ impl<T: ?Sized> Perm<*const T> {
...
+    #[trusted]
+    #[check(ghost)]
+    #[ensures(result.ward() == self.ward().thin())]
+    #[ensures(result.len()@ == self.len())]
+    pub fn live(&self) -> PtrLive<'_, T> {
+        panic!("called ghost function in normal code")
+    }
+}
+
+/// Extension trait with the `live_mut` method
```
jhjourdan: Why do we need an extension trait for this? Since `Perm` is defined in the same crate, we should be able to use a simple `impl` block.

creusot-std/src/std/ptr.rs
Lysxia: Oh, fancy syntax.

jhjourdan:
:tada: :tada: :tada:

PR #1535, opened 2025-05-21
Avoid unnecessarily resolving dependencies

Lysxia:
I was having problems with dependencies having wrong names because they are normalized at expansion time by `update_graph`. This moves them to dependency insertion time in `dependency`.

~~I'm probably still missing something because this change introduces duplicate declarations.~~ Trait laws were added to the expansion queue directly and needed an ad hoc call to normalization.

### DIFF
--- a/creusot/src/backend/clone_map/elaborator.rs
+++ b/creusot/src/backend/clone_map/elaborator.rs
@@ -78,14 +78,17 @@ impl<'tcx> Namer<'tcx> for ExpansionProxy<'_, 'tcx> {
         self.namer.normalize(ctx, ty)
     }
 
-    fn dependency(&self, dep: Dependency<'tcx>) -> &Kind {
-        let dep = dep.erase_regions(self.tcx());
+    fn raw_dependency(&self, dep: Dependency<'tcx>) -> &Kind {
         self.expansion_queue.borrow_mut().push_back((self.source, Strength::Strong, dep));
-        self.namer.dependency(dep)
+        self.namer.raw_dependency(dep)
     }
 
     fn tcx(&self) -> TyCtxt<'tcx> {
         self.namer.tcx()
+    }
+
+    fn typing_env(&self) -> TypingEnv<'tcx> {
+        self.namer.typing_env()
     }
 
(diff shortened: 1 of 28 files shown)


jhjourdan:
Okay, I looked into that.

I agree the current situation is not good, because there is no clear invariant on whether unresolved items are allowed to enter the expansion queue, and I agree that only allowing resolved items seem to be a good idea.

That being said, if we take apart the case of laws (which I consider being a bug you solve in this PR), we already only insert resolved items to the queue, and the `dependency` function is already only called (I think) with resolved items. So I actually don't think doing additional resolutions in `dependency` is a good idea, because this is redundant : it may mask bugs where resolutions are missing in some piece of code.

You can always perform resolution before calling `Namer::item`, or even add a method `Namer::resolved_item`.

jhjourdan:
Apart from that, I am rather worried by the fact that the order of declarations get changed a lot byt this PR, which does not seem to have an impact on the dependency graph. If we want to merge this, I think this would need to be investigated.

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -512,8 +507,9 @@ fn expand_laws<'tcx>(
     }
 
     for law in ctx.laws(item_container) {
+        let dep = elab.namer(dep).normalize_dependency(Dependency::Item(*law, item_subst));
         // We add a weak dep from `dep` to make sure it appears close to the triggering item
-        elab.expansion_queue.push_back((dep, Strength::Weak, Dependency::Item(*law, item_subst)));
+        elab.expansion_queue.push_back((dep, Strength::Weak, dep));
```
Lysxia, line 512: Oops. This mistake was causing all the shuffling around.

Lysxia:
I removed the resolution that is done for every dependency and instead only added it to the laws. There are also  redundant calls to `erase_regions` which I simplified. Removing them altogether seems to be a bigger task.

jhjourdan:
I added a commit to refactor a bit the resolution function of trait.rs.

I think we can now merge this, as soon as you are ready.

PR #1536, opened 2025-05-21
Specialize extern implementation of extern traits

arnaudgolfouse:
A small thing that I found: if a trait is specified using `extern_spec!`, and an impl is declared in a dependency (eg `creusot_contracts`), the spec for the _trait_ (and not the impl) was used.


jhjourdan:
LGTM.

PR #1539, opened 2025-05-21
Support for const

Lysxia:
This depends on #1535 See test cases in `lang/const.rs` for what is implemented (these new tests all pass now)

- [ ] Add support for user-defined data types

### DIFF
--- a/tests/should_succeed/lang/const.coma
+++ b/tests/should_succeed/lang/const.coma
@@ -4,12 +4,14 @@ module M_const__foo [#"const.rs" 10 0 10 27]
   
   use creusot.int.UInt64
   use creusot.prelude.Any
-
+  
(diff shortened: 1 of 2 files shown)


Halbaroth:
The following code is present in `IndexMap`:
```rust 
const MAX_ENTRIES_CAPACITY: usize = (isize::MAX as usize) / mem::size_of::<Bucket<K, V>>();
``` 

It would be nice to support it!

jhjourdan:
@Lysxia :

Could you please add the following test to this PR:

```rust
fn some_fn_pointer() -> i32 {
    2
}

fn main() {
    let x = &some_fn_pointer;
    let y = x();
}
```

It should be accepted once consts work.

Lysxia:
Redone in #1665, including the test cases above.

PR #1540, opened 2025-05-21
Fix dummy requires and ensures to also parse trait method signatures

Lysxia:


jhjourdan:
Could you elaborate on what was the bug?

Lysxia:
The dummy macros parsed a `syn::ImplItemFn` (which must have a body) whereas the creusot macros parse `ContractSubject`.

PR #1541, opened 2025-05-22
Add seq! macro in Pearlite

Lysxia:


xldenis:
I think we might even be able to encode this directly using the epsilon operator in why3.

jhjourdan:
Yeah, that's probably a better idea, actually. builmding sequences by successive `cons`es is probably terrible for automation.

xldenis:
> How do you use epsilon in why3?

you can use function literals in why3 `[| a; b; c |]` I think you might need to provide a fallback case as well? Don't quite remember anymore.

xldenis:
> Another solution, as @xldenis sketches, is to use seq.Seq.create, which creates a sequence from a function, which you would define as a series of if-then-else statement. Not particularly elegant, and I have no idea whether SMTs like this.

I thought why3's function literal syntax uses epsilon inside?

jhjourdan:
What do you pass to `Seq.create`, concretely?

xldenis:
`Seq.create N [| 1; 2; 4; 5 |]` ?

jhjourdan:
IS this valid Why3 syntax ??

xldenis:
>IS this valid Why3 syntax ??

I mean maybe I made a mistake with the exact function literal syntax but surely the principle is valid? I thought function literals gave you a function `int -> 'a` which is exactly what `Seq.create` needs?

jhjourdan:
Oh, you're right!

(That being said, I'm not convinced that this is not translated to a cascade of ifs. But this is much prettier.)

Lysxia:
OK I'll try doing this with function literals.

xldenis:
>(That being said, I'm not convinced that this is not translated to a cascade of ifs. But this is much prettier.)

I was hoping it would get translated to an uninterpreted function + axioms for each case ie:

```
(declare-fun f (Int) (T))

(assert (= (f 0) 1)
...
```
but of course that's just wishful thinking and not backed by evidence!

PR #1542, opened 2025-05-23
Resource algebras

arnaudgolfouse:
This defines resource algebras, a powerful way to use ghost state from Iris.


jhjourdan:
Please do not commit the `rustc-ice` file.

creusot-contracts/src/logic/fmap.rs
```diff
@@ -80,6 +80,40 @@ impl<K, V: ?Sized> FMap<K, V> {
...
+    #[logic]
+    #[ensures(
+        forall<k: K>
+          result.get(k) ==
+          match self.get(k) {
+              None => None,
+              Some(x) => f[(k, x)],
+          }
+    )]
+    pub fn fmapi<U>(self, f: Mapping<(K, V), Option<U>>) -> FMap<K, U>
```
jhjourdan, line 110: That's the same thing as `filter_map`, right?

creusot-contracts/src/logic/id.rs
```diff
@@ -0,0 +1,59 @@
...
+    fn deep_model(self) -> Self::DeepModelTy {
+        self
+    }
+}
+
+impl Id {
+    /// Generate a fresh id
+    #[logic]
+    #[open(self)]
+    pub fn fresh() -> Self {
```
jhjourdan, line 56:
```suggestion
    pub fn fresh() -> Ghost<Self> {
```

Otherwise, you don't have the guarantee that `Id` object do not exist at runtime.

And anyway, since you do not have any freshness guarantee at the postcondition (and you can't have once), this function is useless.

creusot-contracts/src/logic/ra/auth.rs
```diff
@@ -0,0 +1,116 @@
...
+    #[ensures(Self::rel(a, f2))]
+    fn rel_mono(a: Self::Auth, f1: Self::Frag, f2: Self::Frag) {}
+}
+
+/// The 'authority' Resource Algebra.
+///
+/// This is a specialisation of [`View`], where:
+/// - both `Auth` and `Frag` are the same type
+/// - the relation between the two is that `Frag` must always be included in `Auth`
+pub struct Auth<T: RA>(View<AuthViewRel<T>>);
```
jhjourdan, line 37: Move this higher, because the comment is helpful to understand the definitions above.

creusot-contracts/src/logic/ra.rs
```diff
@@ -0,0 +1,161 @@
...
+    /// For every element, there is a maximal (in the sense of [`Self::incl`]) part
+    /// of `self` that is [`Self::idemp`].
+    #[logic]
+    #[requires(self.valid())]
+    #[ensures(
+        (forall<b: Self> ! (b.incl(self) && b.idemp())) ||
+        (exists<b: Self> b.incl(self) && b.idemp() &&
+           forall<c: Self> c.incl(self) && c.idemp() ==> c.incl(b))
+    )]
+    fn maximal_idemp(self);
```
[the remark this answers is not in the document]
Armael: I would prefer to retain the name `maximal_idemp`, since this definition of RA follows the work on maximal idempotents whose goal is precisely to get rid of the `core` function. I think calling this function `core` would simply be confusing.

The validity assumption is present in the iris development of the maximal idempotents paper… (cf the artifact of the paper). So I'm a bit suspicious of removing it.

creusot-contracts/src/logic/ra/auth.rs
```diff
@@ -0,0 +1,116 @@
...
+        self.0.incl(other.0)
+    }
+
+    #[logic]
+    #[open]
+    #[ensures(result == (self.op(self) == self))]
+    fn idemp(self) -> bool {
+        self.0.idemp()
+    }
+}
```
jhjourdan, line 116: We *really* need a way of not repeating all these specs...

creusot-contracts/src/resource.rs
```diff
@@ -0,0 +1,226 @@
...
+    #[pure]
+    #[requires(self@ == a.op(*b))]
+    #[ensures(result.0.id() == self.id() && result.1.id() == self.id())]
+    #[ensures(result.0@ == *a)]
+    #[ensures(result.1@ == *b)]
+    pub fn split(self, a: Snapshot<R>, b: Snapshot<R>) -> (Self, Self) {
+        panic!("ghost code only")
+    }
+
+    /// Split a resource into to, and join it again once the mutable borrows are dropped.
```
jhjourdan, line 123:
```suggestion
    /// Split a resource into two, and join it again once the mutable borrows are dropped.
```

creusot-contracts/src/resource.rs
```diff
@@ -0,0 +1,226 @@
...
+    ///
+    /// # Corresponding reasoning
+    ///
+    /// `⌜valid(value)⌝ ⊢ ∃γ, Own(value, γ)`
+    #[trusted]
+    #[requires(r.valid())]
+    #[ensures(result@ == *r)]
+    pub fn alloc(r: Snapshot<R>) -> Ghost<Self> {
+        Ghost::conjure()
+    }
```
jhjourdan, line 94:
```suggestion
    #[trusted]
    #[pure]
    #[requires(r.valid())]
    #[ensures(result@ == *r)]
    pub fn alloc(r: Snapshot<R>) -> Ghost<Self> {
        Ghost::conjure()
    }
```

Uh oh, the purity checker should have spotted this. Is there a bug in the purity checker?

creusot-contracts/src/resource.rs
```diff
@@ -0,0 +1,226 @@
...
+    #[requires(self.id() == other.id())]
+    #[ensures(result.id() == self.id())]
+    #[ensures(result@ == self@.op(other@))]
+    pub fn join(self, other: Self) -> Self {
+        panic!("ghost code only")
+    }
+
+    /// Same as [`Self::join`], but put the result into `self`.
+    #[pure]
+    #[trusted]
```
jhjourdan, line 167:
```suggestion
    #[trusted]
    #[pure]
```

creusot-contracts/src/resource/fmap_view.rs
```diff
@@ -0,0 +1,195 @@
...
+        match self.0.val().auth {
+            None => FMap::empty(),
+            Some(Excl::Bot) => FMap::empty(),
+            Some(Excl::Excl(auth)) => auth,
+        }
+    }
+
+    /// Create a new, empty authoritative map.
+    #[ensures(result@ == FMap::empty())]
+    pub fn new() -> Ghost<Self> {
```
jhjourdan, line 123:
```suggestion
    #[pure]
    #[ensures(result@ == FMap::empty())]
    pub fn new() -> Ghost<Self> {
```

creusot-contracts/src/resource/fmap_view.rs
```diff
@@ -0,0 +1,195 @@
...
+/// Wrapper around a [`Resource`], that allows to agree on the values of a [`FMap`].
+///
+/// This is the fragment version
+pub struct Fragment<K, V>(FMapView<K, V>);
+
+impl<K, V> Invariant for Authority<K, V> {
+    #[predicate]
+    #[open(self)]
+    fn invariant(self) -> bool {
+        pearlite! { self.0@.auth != None }
```
jhjourdan, line 69:
```suggestion
        pearlite! { self.0@.auth != None && self.0@.frag == None }
```

creusot-contracts/src/resource/fmap_view.rs
```diff
@@ -0,0 +1,195 @@
...
+    #[ensures((^self).id() == self.id())]
+    #[ensures(result@ == (*k, *v))]
+    #[ensures(result.id() == self.id())]
+    #[pure]
+    #[allow(unused_variables)]
+    pub fn insert(&mut self, k: Snapshot<K>, v: Snapshot<V>) -> Fragment<K, V> {
+        let new_auth = snapshot!(View::mkauth(self.auth().insert(*k, *v)));
+        let new_total = snapshot!(match self.0@.frag {
+            None => *new_auth,
+            Some(frag) => new_auth.op(View::mkfrag(frag)),
```
jhjourdan, line 141: Isn't this dead code, thanks to the type invariant I added above?

creusot-contracts/src/resource/fmap_view.rs
```diff
@@ -0,0 +1,195 @@
...
+    /// Get the fragment of the map represented by this resource.
+    ///
+    /// This can also be accessed using the [`view`](crate::View::view) operator `@`.
+    #[logic]
+    #[open(self)]
+    pub fn frag(self) -> (K, V) {
+        let frag_agree = match self.0.val().frag {
+            None => FMap::empty(),
+            Some(frag) => frag,
+        };
```
jhjourdan, line 184: We need something like `unwrap_logic` for this kind of code. This is too verbose.

creusot-contracts/src/resource.rs
jhjourdan: So we never ever call this function in ghost context? Including in your persistant array code?

creusot-contracts/src/logic/ra/auth.rs
dianegolfouse: Yep, that's better

creusot-contracts/src/logic/ra/auth.rs
dianegolfouse: The type alias above does the trick!

creusot-contracts/src/resource.rs
dianegolfouse: Indeed. Honestly I don't think
```rust
let resource = Resource::alloc(...);
```
vs
```rust
let resource = ghost!(Resource::alloc(...).into_inner());
```
will be compiled differently, inlining `alloc` into nothingness is trivial.

creusot-contracts/src/resource.rs
```diff
@@ -0,0 +1,226 @@
...
+        dead
+    }
+
+    /// Get the id for this resource.
+    ///
+    /// This is the same as [`Self::id`], but for ghost code.
+    #[pure]
+    #[trusted]
+    #[ensures(result == self.id())]
+    pub fn id_ghost(&self) -> Id {
```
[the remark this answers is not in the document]
dianegolfouse: - That's not the design we chose for ghost code! If `alloc` can return `Resource<T>` then it is usable in program (which we don't want).
- This I agree with. I can see the `Snapshot<T> -> Ghost<T>` supplanting this at a later time however.
- > `let _z = x.join_shared(y);` should be enough.
  
  But there is a precondition on `join_shared`, stating that `x.id()` must equal `y.id()`! I really don't see how to do the `excl` test without the dynamic test on `id`.

creusot-contracts/src/logic/ra.rs
```diff
@@ -0,0 +1,161 @@
...
+    /// Some(x).incl(Some(y)) == (x == y || x.incl(y))
+    /// ```
+    ///
+    /// Note that the paper on the maximal idempotent axiom (GMB) uses the
+    /// reflexive definition of `incl` on paper, but not in its accompanying
+    /// Iris formalization, where it uses the non-reflexive definition (as
+    /// we do here).
+    #[logic]
+    #[ensures(result == exists<c: Self> self.op(c) == other)]
+    fn incl(self, other: Self) -> bool;
```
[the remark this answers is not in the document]
dianegolfouse: I did that, and it works. The main "issue" is that the spec for `maximal_idemp` is now very ugly 😅

creusot-contracts/src/resource.rs
jhjourdan: Except if `alloc` is not inlined... Which may happen if optimizations are not enabled.

jhjourdan:
I just push a simpler version of some definitions for the FMap RA.

dianegolfouse:
> I just push a simpler version of some definitions for the FMap RA.

This is very nice!

jhjourdan:
> This is very nice!

Actually, `incl` and `maximal_idemp` should be made open for Fmap.

creusot-contracts/src/logic/ra.rs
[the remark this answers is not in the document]
jhjourdan: Yeah, well, we could define the obvious Boolean wrapper. Let's see later if this becomes untractable.

creusot-contracts/src/resource.rs
jhjourdan: Alright, we discussed, I'm now somewhat convinced that it's not too bad to rely o nthe optimizer for this kind of optimizations.

creusot-contracts/src/resource.rs
```diff
@@ -0,0 +1,226 @@
...
+        self.val()
+    }
+}
+
+impl<R: RA> Invariant for Resource<R> {
+    #[predicate]
+    #[open]
+    fn invariant(self) -> bool {
+        self.val().valid()
+    }
```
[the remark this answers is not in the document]
jhjourdan: Done.

jhjourdan:
(I just rebased wrt. master.)

jhjourdan:
So... I redefined resource algebra without the `valid` predicate, and using a subset type for views. This uses sealed methods for simplifying a lot of specifications, which is very nice.

In the meantime, I fixed a big unsoundness bug, where the `Resource` type was not annotated with `#[trusted]`, resulting in it being a singleton, which essentially made all the RA elements which are values of resources equal... I am a bit afraid that proofs in persistent arrays will get more complex because of this.

I decided to generalize the notion of view relation, making it take an option of the authority, so that we can use it in the case there is no authority. This gets us the best of both worlds, where the view relation can be used to describe invariants on valid fragments (as described in https://gitlab.mpi-sws.org/iris/iris/-/merge_requests/543#note_58219), while still giving us the opertinity to make this trivial when there is no view, so that in practice fragments are easy to build.

I think the result is mostly satisfying, except that working with views is rather painful. The reason is that functions like `Resource::split_off` requires proving that elements of RAs are *equal*, which is made difficult by the subset type (a specific lemma about extensional equality has to be used for this). Also, updates have a tendency to be rather difficult to prove by solvers, but I don't think this is because of this change.

To solve these issues, the next improvements I can see are the following:

1- In the `Resource` type, do never use equality over RA elements. Instead, I think we can replace the `fn val(self) -> R` getter with a monotonic `fn has(self, R) -> bool` predicate. Then, everything should be specified using `has` and `incl`. The good news is that `incl` is defined extensionally, which should make it much more reliable to prove by solvers.

Incidentally, this change in philosophy makes it natural to specify a function `#[ensures(forall<x> self.has(x) ==> result.has(x.maximal_idemp()))] fn core(&self) -> Resource`, which would play the role of the `Clone` instance, except that it works even if the resource is not duplicable (but you do not get the same resource in the end if it is not duplicable).

This also makes it possible to simplify the spec of functions like `split_off`, by using `factor`: you pass the resource you want to extract, and it changes the resource in the borrow into whatever `factor` tells it remains. It is just as expressive, thanks to the monotony of `has`.

2- I would add an `Updater` trait, whose instance describe one kind of update you can do with an RA. The trait would contain methods describing the final state, and the conditions that need to hold to do the update, and a property stating that the update is valid. Then, `update` would take an `impl Updater` instead of an RA element as a parameter. By adding a generic instance `RA` -> `Updater`, we can get the same behavior as now for the cases where the solvers can prove the update easily, but for the more complex cases, one can "prove" the validity of the update by passing another type as parameter.

What do you think of these ideas?

jhjourdan:
(BTW, we need to squash the commits of this PR. The history makes no sense.)

jhjourdan:
> 1- In the `Resource` type, do never use equality over RA elements. Instead, I think we can replace the `fn val(self) -> R` getter with a monotonic `fn has(self, R) -> bool` predicate. Then, everything should be specified using `has` and `incl`. The good news is that `incl` is defined extensionally, which should make it much more reliable to prove by solvers.
> 
> Incidentally, this change in philosophy makes it natural to specify a function `#[ensures(forall<x> self.has(x) ==> result.has(x.maximal_idemp()))] fn core(&self) -> Resource`, which would play the role of the `Clone` instance, except that it works even if the resource is not duplicable (but you do not get the same resource in the end if it is not duplicable).
> 
> This also makes it possible to simplify the spec of functions like `split_off`, by using `factor`: you pass the resource you want to extract, and it changes the resource in the borrow into whatever `factor` tells it remains. It is just as expressive, thanks to the monotony of `has`.

I tried this, and mostly failed. The reason is that when we do that, the context contains plenty of `incl` relationships, which the solver try to unfold and combine, and they get confused.

Instead, I specified `FmapView` in terms of inclusion, which help quite a bit with the proofs.

jhjourdan:
> 2- I would add an `Updater` trait, whose instance describe one kind of update you can do with an RA. The trait would contain methods describing the final state, and the conditions that need to hold to do the update, and a property stating that the update is valid. Then, `update` would take an `impl Updater` instead of an RA element as a parameter. By adding a generic instance `RA` -> `Updater`, we can get the same behavior as now for the cases where the solvers can prove the update easily, but for the more complex cases, one can "prove" the validity of the update by passing another type as parameter.

I did this, and I think it is a good thing. Of course, there is not as much updates and local updates declared as in Iris, and we will have to complete the list in the future.

jhjourdan:
As far as I'm concerned, I'm happy with the current state, and I'm ready to merge (if the commits are squashed).

dianegolfouse:
> In the meantime, I fixed a big unsoundness bug, where the Resource type was not annotated with `#[trusted]`, resulting in it being a singleton, which essentially made all the RA elements which are values of resources equal... I am a bit afraid that proofs in persistent arrays will get more complex because of this.

It still works 🙂

Issue #1544, opened 2025-06-01
Internal compiler panic when using Rust's native `PartialEq` derive macro


MathisWellmann:
I was trying to use `creusot`, built from the newest version at commit `879cd335441a1a126380c007d38cb499faa316dc`,
and using it on the following function:
```rust
/// Return the minimum of two values
#[inline(always)]
#[allow(unused)]
#[creusot_contracts::ensures(result == v0 || result == v1)]
pub(crate) fn min(v0: i32, v1: i32) -> i32 {
    if v0 < v1 { v0 } else { v1 }
}
```

Now running `cargo creusot prove` results in the following error:
```
thread 'rustc' panicked at creusot/src/validate/purity.rs:190:30:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: core::option::unwrap_failed
   4: <creusot::validate::purity::PurityVisitor as rustc_middle::thir::visit::Visitor>::visit_expr
   5: <creusot::validate::purity::PurityVisitor as rustc_middle::thir::visit::Visitor>::visit_expr
   6: rustc_middle::thir::visit::walk_stmt
   7: rustc_middle::thir::visit::walk_expr
   8: <creusot::validate::purity::PurityVisitor as rustc_middle::thir::visit::Visitor>::visit_expr
   9: creusot::validate::purity::validate_purity
  10: creusot::translation::before_analysis
  11: <creusot::callbacks::ToWhy as rustc_driver_impl::Callbacks>::after_expansion
  12: rustc_interface::passes::create_and_enter_global_ctxt::<core::option::Option<rustc_interface::queries::Linker>, rustc_driver_impl::run_compiler::{closure#0}::{closure#2}>::{closure#2}::{closure#0}
  13: rustc_interface::interface::run_compiler::<(), rustc_driver_impl::run_compiler::{closure#0}>::{closure#1}
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

error: the compiler unexpectedly panicked. this is a bug.

note: using internal features is not supported and expected to cause internal compiler errors when used incorrectly

note: please attach the file at `/home/magewe/MathisWellmann/lfest-rs/rustc-ice-2025-06-01T19_59_00-1691872.txt` to your bug report

note: compiler flags: --crate-type lib -C embed-bitcode=no -C debuginfo=2

[...]

PR #1546, opened 2025-06-02
Crash purity traits

arnaudgolfouse:
Fixes 2 crashes related to #1544


creusot/src/validate/purity.rs
```diff
@@ -184,11 +184,13 @@ impl<'a, 'tcx> thir::visit::Visitor<'a, 'tcx> for PurityVisitor<'a, 'tcx> {
...
                     // try to specialize the called function if it is a trait method.
                     let subst = self.ctx.erase_regions(subst);
-                    let func_did =
+                    let Some((func_did, _)) =
                         TraitResolved::resolve_item(self.ctx.tcx, self.typing_env, func_did, subst)
                             .to_opt(func_did, subst)
-                            .unwrap()
-                            .0;
+                    else {
+                        self.thir_failed = true;
```
jhjourdan: I'm a bit stressed about this: we do never check that `thir_failed` is indeed `false` at the end of all checks.

Can we at least add an assertion in `before_analysis`?

jhjourdan:
Could you please add a test?

creusot/src/validate/purity.rs
dianegolfouse: I reworked the way we propagate `CannotFetchThir` errors: in the worst case they result in a crash (if they are dropped without being handled), and right now they seem all handled (I hope).
They are always created from rustc's `ErrorGuaranteed`, so an error will always be reported when they arise.

dianegolfouse:
Also I found 2 more similar crashes

jhjourdan:
OK, LGTM, merging.

PR #1549, opened 2025-06-05
Small changes to avoid RA errors

Lysxia:


.vscode/settings.json
```diff
@@ -1,6 +1,5 @@
 {
     "rust-analyzer.rustc.source": "discover",
-    "rust-analyzer.cargo.features": ["creusot"],
```
jhjourdan, line 3: Doesn't that mean that rust analyzer will run in non-creusot mode when editing creusot-contracts ? This seems inapropriate.

Lysxia: This feature doesn't change that. We already don't run rust analyzer in non-creusot mode. It's probably not even possible to make rust-analyzer run creusot for only creusot-contracts.

The `creusot` feature only affects the `-proc` crate, enabling type-checking of the creusot versions of the macros, without actually switching the crate to use them. So we lose that, but in exchange RA doesn't crash as soon as I open a file from `cargo-creusot`.

jhjourdan: Ok, fair enough.

PR #1550, opened 2025-06-05
Move Creusot logic into after_analysis

### LINKED ISSUE #1547: Clone THIR and do every creusot analysis in `after_analysis`
We should stop doing analyses is `before_analysis`: type checking is not finished here, and that causes us a lot of pains.
What to do:
- Clone all the THIR in `before_analysis`, store it in the translation context
- Do all the analyses (purity, termination, ...) in `after_analysis`, using the THIR we stored when needed
- Presumably, this would eliminate the need for `CannotFetchThir`

This is probably fine performance-wise.

Lysxia:
Fix #1547 

- The core change is to only clone the THIR in before_analysis so we are guaranteed to have access to it in all other passes afterwards.
- Simplify error handling: if we don't find a THIR body, there must be a pending type error, so we are free to abort now (in pearlite()) or skip work silently (in validation functions) with the guarantee that the pending error will be raised later.
- No more need to "pre-load" Pearlite terms (calls to `term()` that discard their result).
- Move all the validation into its own function.

### DIFF
--- a/creusot/src/backend.rs
+++ b/creusot/src/backend.rs
@@ -6,7 +6,6 @@ use why3::declaration::Attribute;
 use crate::{
     contracts_items::{is_resolve_function, is_spec, is_trusted},
     ctx::{ItemType, TranslatedItem, TranslationCtx},
-    error::CannotFetchThir,
     naming::ModulePath,
     options::SpanMode,
     run_why3::SpanMap,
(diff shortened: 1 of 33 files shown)


jhjourdan:
That's a big simplification, at the cost of a (hopefully) small performance loss. Thanks !

PR #1555, opened 2025-06-12
Fix handling of Pearlite closures in `validate_purity` and `validate_terminates`

Lysxia:
Closures for Pearlite quantifiers and mappings were mistaken for program closures, because they are only marked with `no_translate`. That attribute helped mitigate the mistake, but not completely because I am running into crashes in `validate_purity` and `validate_terminates` when they call `sig` believing such closures to be program closures.

- The first fix is to have `validate_purity` skip closures altogether. Closures will be visited while traversing their parents. This also ensures that each expression is only traversed once, which was previously not the case.
- The second fix is to mark Pearlite closures (= quantifiers and mappings) with another attribute `#[creusot::logic_closure]` to be included in the `is_pearlite` check. This avoids the crash in `validate_terminates`.
- I couldn't isolate a regression test case; when I tried to copy the culprit from my project into a standalone test case the crash went away. The crash is that in `creusot/src/backend/closures.rs`, in the function `ClosSubst::pre`, the `assert!(cap.place.projections.is_empty())` failed. This might be a symptom of yet another bug.

### DIFF
--- a/creusot-contracts-proc/src/creusot/pretyping.rs
+++ b/creusot-contracts-proc/src/creusot/pretyping.rs
@@ -294,6 +294,7 @@ pub fn encode_term(term: &RT) -> Result<TokenStream, EncodeError> {
             ts = quote_spanned! {sp=>
                 ::creusot_contracts::__stubs::#quant_token(
                     #[creusot::no_translate]
+                    #[creusot::logic_closure]
                     |#args| { #ts }
                 )
             };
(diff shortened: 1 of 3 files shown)


jhjourdan:
I guess the reason why you could not reproduce the bug is because you were using a too recent edition of Rust. There is a known bug where closures of recent editions are not handled correctly, because they capture places instead of capturing locals, and Creusot does not handle this currently.

PR #1558, opened 2025-06-13
Translate `size_of`

Lysxia:
- Add `size_of_logic`, the logic and `Int`-valued counterpart of `size_of`.
- Add extern spec to `size_of`. Note that `size_of` can crash at compile-time when specialized to a too big type.
- This required fixing the `extern_spec` macro to add type arguments to function calls that it generates for type checking.
- `size_of_logic` is handled specially to give it a definition in Why3. Types that have no specified size will be left abstract.

Future work: define the size of `repr(C)` types and of arrays with non-literal lengths.


creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -333,6 +337,108 @@ fn expand_ty_inv_axiom<'tcx>(
...
+        Adt(_, _) => None, // TODO
+        Array(t, n) => size_of_array(names, ctx, def_id, t, n),
+        _ => None,
+    };
+    let type_ = int_ty(ctx, names);
+    Decl::ConstantDecl(Constant { name, type_, body })
+}
+
+/// `Box<T>`, `Option<&T>`, or `Option<Box<T>>` with `T: Sized`
+fn is_pointer_adt<'tcx>(
```
jhjourdan: Isn't there anything in rustc API for this?

It seems to be something that should be handled by rustc. Also, `Option` is only one of the ADTs that can leverage this optimization.

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -252,6 +254,8 @@ impl DepElab for LogicElab {
         let sig = lower_logic_sig(ctx, &names, name, pre_sig, def_id);
         if !opaque && let Some(term) = term(ctx, typing_env, &bound, dep) {
             lower_logical_defn(ctx, &names, sig, kind, term)
+        } else if let Some(decl) = try_expand_size_of_logic(&names, ctx, name, def_id, subst) {
+            vec![decl]
```
jhjourdan: The usual convention for special functions is to generate Pearlite code, and then handle them in `term`. Why do we need yet another mechanism here?

Lysxia: idk. (1) the `else` branch below here seems to be doing something similar. (2) why the extra indirection, generate Pearlite then translate to Why3, instead of directly generating Why3?

jhjourdan: 2) It is often easier to write Pearlite terms than WhyML terms. In your case, this is probably equivalent, though

1) The `else` branch is not doing something similar, it is adding some axioms that add specifications to an already existing definition. Admitedly, this could be done somewhere else, but clearly not in `term`.

creusot/src/backend/clone_map/elaborator.rs
Lysxia: I couldn't find anything relevant in the rustc API. In terms of implementation I guess this is handled by the niche optimization, so they wouldn't have special logic for `Option`.

In terms of specification, the Rust Reference does not mention the niche optimization nor any special layout rules for `Option`. It makes no guarantees about the default `repr(Rust)`, while requiring `repr(C)` enums with fields like `Option` to separate the tag and the fields so the niche optimization isn't allowed.

The only mention of the layout of `Option` I found is the `std` doc for `size_of`. It only specifies the size of a few types beyond the Rust Reference, and the comments above are copied from there.

At the moment I think that the size of all other `repr(Rust)` ADTs should be left unspecified in our logic.

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -1012,6 +1009,99 @@ fn fn_mut_hist_inv_term<'tcx>(
...
+        _ => return None,
+    };
+    if n == 0 {
+        Some(Term::int(int_ty, 0))
+    } else {
+        let subst = ty::GenericArgs::for_item(ctx.tcx, def_id, |_, _| ty.clone().into());
+        let item_size = Term {
+            ty: int_ty,
+            span: DUMMY_SP,
+            kind: TermKind::Call { id: def_id, subst, args: [].into() },
```
jhjourdan: Cannot you use `Term::call` instead? I'm trying to force us using these smart constructors, because they are so much less verbose.

creusot/src/backend/clone_map/elaborator.rs
jhjourdan: At least, you can use: https://doc.rust-lang.org/beta/nightly-rustc/rustc_middle/ty/struct.TyCtxt.html#method.layout_of

If this fails, then there isn't much we can do, but this should work at least for pointers, and options of pointers.

Lysxia: I worry that this will result in `size_of_logic` being defined for all ground `repr(Rust)` types with values that depend on the compiler version.

jhjourdan: Ok, I see your problem.

I would have assumed that this is fine, just as `const` code in Rust depends on the compiler version.

It's fine if the proof depends on unstable details of the internal representation of types, as long as the developper commits to update it with new versions of Rust.

This is the same as for updating pieces of *code* which depend on size of types.

Lysxia: OK. I'm not a fan of this but we can think of a better solution later.

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -1012,6 +1008,50 @@ fn fn_mut_hist_inv_term<'tcx>(
...
+    ty: Ty<'tcx>,
+    n: &Const<'tcx>,
+    int_ty: Ty<'tcx>,
+) -> Option<Term<'tcx>> {
+    let n = match n.kind() {
+        ConstKind::Value(_, ty::ValTree::Leaf(scalar)) => scalar.to_target_usize(ctx.tcx) as i128,
+        // TODO: ConstKind::Param
+        _ => return None,
+    };
+    let subst = ty::GenericArgs::for_item(ctx.tcx, def_id, |_, _| ty.clone().into());
```
jhjourdan:
```suggestion
    let subst = ctx.mk_args(&[ty.clone().into()]);
```

jhjourdan:
Apart from the nitpick above, LGTM. You can merge.

PR #1559, opened 2025-06-14
Sealed default methods.

jhjourdan:
The termination check fails on the only test I added. I don't remember how it is supposed to work with trait methods. 

@arnaudgolfouse any idea?


dianegolfouse:
Did you fix the issue? I can't see anything wrong with either `sealed.rs` or `19_sealed.rs`

jhjourdan:
No, I did not

`19_sealed` is rejected by the termination checker, which is bad because this is typically the kind of things we want to do with sealed defautls.

dianegolfouse:
Ah I see: I was testing on master, without the `(sealed)`, and it works fine there.

So the issue is that the trait search always returns an instance for sealed functions. But, in `terminates.rs:383`, we consider that calling a function with a trait bounds call all the functions of that trait, specialized with the types of the call.
So when visiting
```rust
#[logic(sealed)]
fn g(self, x: Int) -> Int {
    self.f(x) + 1
}
```
It says: ah, I'm visiting `self.f`, which has a trait bound `Tr`. So it may call `g`, and it indeed founds a specific implementation for `g` since it is sealed: hence the error.

We could, I _think_, safely relax this specific check in the following case:
- The function we are calling _from_ is a default impl in a trait
- The specialized function we just found, arising from this trait's bound, is a sealed function

This is conservative, because it mimics the existing behavior of non-sealed functions, specifically in the case where we are exploring the functions called by a default implementation.

jhjourdan:
I'm a bit suspicious. How would that prevent `g` recursively calling itself?

jhjourdan:
@arnaudgolfouse This contains rather heavy changes to the termination checker. Could you have a look?

dianegolfouse:
> @arnaudgolfouse This contains rather heavy changes to the termination checker. Could you have a look?

We discussed this, this is still incomplete but simpler, and I am convinced the new implementation has less risks of accidental unsoundness.

jhjourdan:
Didn't you have changes de propose?

If not, please merge ;)

PR #1566, opened 2025-06-18
Basic subset types based on Invariant

jhjourdan:


jhjourdan:
@Armael @arnaudgolfouse Is this what we need for simpler resource algebra?

jhjourdan:
Ok, I merged, and then we'll try to update #1542.

PR #1560, opened 2025-06-16
Allow casts between `*const` and `*mut` in Pearlite

Lysxia:

PR #1569, opened 2025-06-19
Support casts between *mut and *const in programs 

Lysxia:
I thought that was already supported when I did #1560 but it seems not.


jhjourdan:
Thanks.

PR #1579, opened 2025-06-24
Handle all pointer-to-pointer casts

Lysxia:
According to [the Rust Reference](https://doc.rust-lang.org/reference/expressions/operator-expr.html#r-expr.as.pointer) casts are either noops (between same types with same Sized-ness) or a simple projection. I'm just not entirely sure whether `is_sized` is the right way to check sizedness...


creusot/src/backend/program.rs
```diff
@@ -1225,9 +1241,47 @@ fn func_call_to_why3<'tcx, N: Namer<'tcx>>(
...
+    // TODO: is using `is_sized` correct? Its doc says "it can be an overapproximation in generic contexts".
+    // https://doc.rust-lang.org/beta/nightly-rustc/rustc_middle/ty/struct.Ty.html#method.is_sized
+    // We really want it to be `true` only if the type is `Sized` so its pointers are known to be thin.
+    let sized1 = ty1.is_sized(tcx, typing_env);
+    let sized2 = ty2.is_sized(tcx, typing_env);
+    // If ty2 is unsized, then ty1 is also unsized.
+    // Otherwise this cast wouldn't have type checked and the compiler would have failed earlier.
+    if sized1 && sized2 || is_unsized(ty2) || ty1 == ty2 {
+        PtrCastKind::Id
+    } else if sized2 {
```
jhjourdan, line 1279:
```suggestion
    } else if sized2 && is_unsized(ty1) {
```

Otherwise `ty1` may be sized after instantiation.

prelude/prelude.coma
```diff
@@ -67,6 +67,10 @@ module Opaque
   type ptr
   let fresh_ptr (ret (result : ptr) {true}) = any
 
+  function thin (p : ptr) : ptr
+  let thin (p : ptr) (ret (result : ptr) { result = thin p }) = any
```
jhjourdan: This `let` declaration is unnecessary and should go away. Logical functions can be called in Coma code, and this ends up in much shorter Coma code.

tests/should_succeed/lang/cast_ptr.coma
```diff
@@ -1,28 +1,26 @@
...
   
   use creusot.prelude.Opaque
   use creusot.prelude.Any
   
   meta "compute_max_steps" 1000000
   
-  let rec g[#"cast_ptr.rs" 10 0 10 42] (p:Opaque.ptr) (return'  (x:Opaque.ptr))= (! bb0
-    [ bb0 = s0 [ s0 =  [ &_0 <- p'0 ] s1 | s1 = return''0 {_0} ]  ]
+  let rec thin[#"cast_ptr.rs" 9 0 9 41] (p:Opaque.ptr) (return'  (x:Opaque.ptr))= (! bb0
+    [ bb0 = s0 [ s0 = Opaque.thin {p'0} (fun (_res:Opaque.ptr) ->  [ &_0 <- _res ] s1) | s1 = return''0 {_0} ]  ]
```
jhjourdan:
```suggestion
    [ bb0 = s0 [ s0 = [ &_0 <- Opaque.thin p'0 ] s1 | s1 = return''0 {_0} ]  ]
```

creusot/src/backend/program.rs
Lysxia: Casts from unknown to sized pointers are accepted by rustc:
```
pub fn f<T: ?Sized>(p: *mut T) -> *mut usize {
    p as *mut usize
}
```
and we can just let `thin` be the identity on thin pointers.

jhjourdan: I see. It's unclear how we will do that, though. Could you add a comment in the prelude about this?

Lysxia: I added this in the prelude:
```
  (* If p is a fat pointer, strip its metadata.
     If p is a thin pointer, return it unchanged. *)
  function thin (p : ptr) : ptr
```

jhjourdan: :+1:

PR #1580, opened 2025-06-24
Remove creusot-rustc's ability to run Why3

Lysxia:
- This module has been unused and rotting since we started using `why3find`
- It caused `cargo-creusot` to make a redundant call to `setup::creusot_paths` which meant you needed solvers to be installed to be able to run `cargo creusot` (which doesn't use Why3 or the solvers) and our test suite `--test ui`. We could refactor less invasively to avoid this call, but it seems time to remove the culprit instead of increasing the complexity of our code.


jhjourdan:
LGTM, thanks.

PR #1581, opened 2025-06-25
Panic a bit less

dianegolfouse:
Remove some panics in multiple places. The ones in the elaborator were fishy but not hit in practice, but the one in trait resolution actually happened.

### DIFF
--- a/creusot/src/translation/traits.rs
+++ b/creusot/src/translation/traits.rs
@@ -184,7 +184,12 @@ pub(crate) fn evaluate_additional_predicates<'tcx>(
 pub(crate) enum TraitResolved<'tcx> {
     NotATraitItem,
     /// An instance (like `impl Clone for i32 { ... }`) exists for the given type parameters.
-    Instance(Instance<'tcx>),
+    Instance {
+        /// The id and substitution of the specific item found (e.g. the `clone` function in `impl Clone for i32`).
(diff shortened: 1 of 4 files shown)


creusot/src/translation/traits.rs
```diff
@@ -343,23 +342,28 @@ impl<'tcx> TraitResolved<'tcx> {
...
-                        def: (closure_def_id, closure_substs),
-                        impl_: None,
-                    })
+                    TraitResolved::Instance { def: (closure_def_id, closure_substs), impl_: None }
                 }
-                _ => unimplemented!(),
+                rustc_middle::ty::Dynamic(_, _, _) => TraitResolved::Instance {
+                    def: (trait_item_def_id, trait_ref.args),
+                    impl_: None,
+                },
```
jhjourdan, line 350: We don't support `dyn`, why do we bother returning something here?

Moreover, I don't think returning the trait item did is a good idea here. Why would it be what we want?

dianegolfouse: Well, this is caught by a better error later on. I return the trait item did because this is what appears in the Thir/MIR.

jhjourdan: But that's the whole point of this function: translate something abstract that appears in Thir/MIR into a concrete instance. We should not have non-body defids here.

If you really want to return something here, then it's better to return an `UnknownFound`.

PR #1583, opened 2025-06-25
Pearlite parser fixes

jhjourdan:
- Prevent warning when using a parenthized quantifier in a  condition
- Types of binders of quantifiers can now be omitted.


xldenis:
>Types of binders of quantifiers can now be omitted.

finally!

PR #1590, opened 2025-06-28
Specify Clone for arrays.

jhjourdan:


creusot-contracts/src/std/array.rs
```diff
@@ -83,4 +83,10 @@ extern_spec! {
         #[ensures(self@ == result@)]
         fn into_iter(self) -> std::array::IntoIter<T, N>;
     }
+
+    impl<T: Clone, const N: usize> Clone for [T; N] {
+        #[ensures(forall<i> 0 <= i && i < self@.len() ==>
+            T::clone.postcondition((&self@[i],), result@[i]))]
```
xldenis, line 89: We could specify `Copy` as implying a trivial postcondition (since it's meant to capture "just a memcpy")

jhjourdan: Ideally, we would want to add this as a law of `Copy`, but we don't have an easy way to do this....

jhjourdan: Actually, we could add it as an additional VC when proving the refinement for a `Copy` impl, and automatically add this in the context when we refer to the clone spec of something which is  `Copy`. One problem which I anticipate is that our translation is lifetime agnostic, and therefore we cannot determine reliably whether a type in copy.

xldenis: I don't think it's necessary in practice ofc, but admissible.

PR #1597, opened 2025-07-01
Use spans more when crashing

arnaudgolfouse:
Track spans better, and use them to indicate where something went wrong when we crash.


creusot/src/backend/term.rs
```diff
@@ -500,7 +507,7 @@ pub(crate) fn binop_to_binop(op: BinOp) -> WBinOp {
         BinOp::BitXor => WBinOp::BitXor,
         BinOp::Shl => WBinOp::Shl,
         BinOp::Shr => WBinOp::Shr,
-        BinOp::Div => todo!("Refactor binop_to_binop to support Div"),
-        BinOp::Rem => todo!("Refactor binop_to_binop to support Rem"),
+        BinOp::Div => WBinOp::Div,
+        BinOp::Rem => WBinOp::Mod,
```
jhjourdan: I think we should panic in this case, because we are actually following a different control flow for Div and Rem.

dianegolfouse: Ahh, alright

Lysxia (approved):

PR #1599, opened 2025-07-03
Local invariants

dianegolfouse:
Add local invariants, with a notion of namespaces.


creusot-contracts/src/local_invariant.rs
```diff
@@ -0,0 +1,172 @@
+//! Define local invariants
+//!
+//! [Local invariants](LocalInvariant) are not the same as [invariants](Invariant): they
+//! allow the use of a shared piece of data to be used in the invariant (see
+//! [`LocalInvariantSpec`]), but in return they impose a much more restricted access to
+//! the underlying data, as well as the use of [`Namespaces`].
```
xldenis: Can you expand on their use a bit more in the docs? this seems _very_ cool but I can't quite grasp what it would be good for / where it would be limited.

creusot-contracts/src/local_invariant.rs
```diff
@@ -0,0 +1,172 @@
+//! Define local invariants
+//!
+//! [Local invariants](LocalInvariant) are not the same as [invariants](Invariant): they
```
Lysxia:
```suggestion
//! [Local invariants](LocalInvariant) are not the same as [type invariants](Invariant): they
```

I propose to always say "type invariant" when talking about `Invariant`.

creusot-contracts/src/local_invariant.rs
dianegolfouse: Yeah, I added an example in the docs and the test suite. I also kind of want to showcase how to use this with ghost resources, but I fear this will get complex quickly (and there is already the 'Persistent arrays' case study in this case).

xldenis: Ahh I see these are like iris invariants

creusot-contracts/src/local_invariant.rs
```diff
@@ -0,0 +1,224 @@
...
+//! # };
+//!
+//! declare_namespace! { PCELL }
+//!
+//! /// A cell that simply asserts its content's type invariant.
+//! pub struct CellInv<T: Invariant + 'static> {
+//!     data: PCell<T>,
+//!     permission: Ghost<&'static LocalInvariant<PCellLocalInv<T>>>,
+//! }
+//! impl<T: Invariant> Invariant for CellInv<T> {
```
jhjourdan, line 25: Why do we need `T: Invariant`?

creusot-contracts/src/local_invariant.rs
```diff
@@ -0,0 +1,224 @@
...
+//! #     pcell::{PCell, PCellOwn},
+//! #     *,
+//! # };
+//!
+//! declare_namespace! { PCELL }
+//!
+//! /// A cell that simply asserts its content's type invariant.
+//! pub struct CellInv<T: Invariant + 'static> {
+//!     data: PCell<T>,
+//!     permission: Ghost<&'static LocalInvariant<PCellLocalInv<T>>>,
```
jhjourdan:
```suggestion
//!     permission: Ghost<LocalInvariant<PCellLocalInv<T>>>,
```

Why do you store `&'static LocalInvariant<PCellLocalInv<T>>` rather than `LocalInvariant<PCellLocalInv<T>>` directly?

The shared borrow will come from the outside (i.e., the *client* creates a borrow of `CellInv`, which gives rise to a borrow of the `LocalInvariant`. There is no need to create a borrow ourselves.

Also, this borrow requires the constraint `T: 'static`, which seems very artificial to me.

creusot-contracts/src/local_invariant.rs
```diff
@@ -0,0 +1,224 @@
...
+/// fn foo(ns: Ghost<&mut Set<Namespace>>) { /* ... */ }
+/// ```
+pub use base_macros::declare_namespace;
+
+/// The type of _namespaces_ associated with local invariants.
+///
+/// Can be declared with the [`declare_namespace`] macro, and then attached to a local
+/// invariant when creating it with [`LocalInvariant::new`].
+#[cfg_attr(creusot, rustc_diagnostic_item = "namespace_ty")]
+pub struct Namespace(());
```
jhjourdan, line 79:
```suggestion
#[trusted]
pub struct Namespace(());
```

This attribute is *important*, otherwise the type is not opaque, it is isomorphic to `()`, and therefore has only one element: this implies that there exists only one namespace.

creusot-contracts/src/local_invariant.rs
```diff
@@ -0,0 +1,224 @@
...
+
+impl Namespace {
+    /// Used by [`declare_namespace`].
+    #[logic]
+    #[open(self)]
+    #[requires(false)]
+    #[doc(hidden)]
+    pub fn new() -> Self {
+        Self(())
+    }
```
jhjourdan:
```suggestion
    #[logic]
    #[requires(false)]
    #[trusted]
    #[doc(hidden)]
    pub fn new() -> Self {
        Self(())
    }
```

creusot-contracts-proc/src/creusot.rs
```diff
@@ -268,6 +268,21 @@ pub fn snapshot(assertion: TS1) -> TS1 {
...
+    let ident = parse_macro_input!(namespace as Ident);
+    quote! {
+        #[logic]
+        #[open(self)]
+        #[creusot::no_translate]
+        #[creusot::decl::new_namespace]
+        #[allow(nonstandard_style)]
+        pub fn #ident() -> ::creusot_contracts::local_invariant::Namespace {
+            ::creusot_contracts::local_invariant::Namespace::new()
+        }
```
jhjourdan:
```suggestion
        #[logic]
        #[trusted]
        #[creusot::decl::new_namespace]
        #[allow(nonstandard_style)]
        pub fn #ident() -> ::creusot_contracts::local_invariant::Namespace {
            dead
        }
```

Then you don't need to declare `Namespace::new()`.

creusot-contracts/src/local_invariant.rs
jhjourdan: Also, I don't think this hack is actually needed (see comment above).

creusot-contracts/src/local_invariant.rs
```diff
@@ -0,0 +1,224 @@
...
+}
+
+/// A collection of namespaces.
+///
+/// This is given at the start of the program, and must be passed along to [LocalInvariant::open].
+pub struct Namespaces<'a>(::std::marker::PhantomData<&'a mut Vec<Namespace>>);
```
jhjourdan:
```suggestion
#[trusted]
pub struct Namespaces<'a>(::std::marker::PhantomData<&'a ()>);
```

jhjourdan: (No need to refer to `Vec` or `Namespace` in the `PhantomData`.)

creusot-contracts/src/local_invariant.rs
```diff
@@ -0,0 +1,224 @@
...
+
+impl Namespaces<'_> {
+    /// Get the underlying set of namespaces.
+    ///
+    /// Also accessible via the [`view`](View::view) (`@`) operator.
+    #[logic]
+    #[open(self)]
```
jhjourdan: This attribute is useless.

creusot-contracts/src/local_invariant.rs
jhjourdan: But the fact that this type is annotated with a lifetime needs to be explained.

creusot-contracts/src/local_invariant.rs
```diff
@@ -0,0 +1,224 @@
...
+
+/// A ghost structure, that holds a piece of data (`T`) together with an
+/// [invariant](LocalInvariantSpec).
+pub struct LocalInvariant<T: LocalInvariantSpec> {
+    value: UnsafeCell<T>,
+    #[cfg_attr(not(creusot), allow(unused))]
+    public: Snapshot<T::Public>,
+    #[cfg_attr(not(creusot), allow(unused))]
+    namespace: Snapshot<Namespace>,
+}
```
jhjourdan: This really is a opaque (trusted) type, so it should be annotated as such, and therefore the fields `public` and `namespace` are not needed.

```suggestion
#[trusted]
pub struct LocalInvariant<T: LocalInvariantSpec> {
    value: UnsafeCell<T>,
}
```

creusot-contracts/src/local_invariant.rs
```diff
@@ -0,0 +1,224 @@
...
+    #[trusted]
+    #[requires(namespaces@.contains(this.namespace()))]
+    #[requires(forall<t: Ghost<&mut T>> (**t).invariant_with_data(this.public()) && inv(t) ==>
+        f.precondition((t,)) &&
+        // f must restore the invariant
+        (forall<res: A> f.postcondition_once((t,), res) ==> (^t.inner_logic()).invariant_with_data(this.public())))]
+    #[ensures(exists<t: Ghost<&mut T>> (**t).invariant_with_data(this.public()) && f.postcondition_once((t,), result))]
+    #[pure]
+    pub fn open<'a, A>(
+        this: Ghost<&'a Self>,
```
jhjourdan, line 246: Have you tried using this https://doc.rust-lang.org/beta/unstable-book/language-features/arbitrary-self-types.html ?

creusot/src/backend/logic.rs
```diff
@@ -22,6 +23,9 @@ pub(crate) fn translate_logic(ctx: &Why3Generator, def_id: DefId) -> Option<File
     let mut names = Dependencies::new(ctx, def_id);
     let pre_sig = ctx.sig(def_id).clone().normalize(ctx.tcx, ctx.typing_env(def_id));
 
+    let namespace_ty =
+        names.def_ty_no_dependency(get_namespace_ty(ctx.ctx.tcx), GenericArgsRef::default());
```
jhjourdan, line 27: Why don't you want the dependency? This seems a bit fishy.

tests/should_succeed/local_invariant_cellinv.coma
```diff
@@ -0,0 +1,722 @@
...
+    | & _4: MutBorrow.t t_T = Any.any_l ()
+    | & _6:  (MutBorrow.t t_PCellOwn) = Any.any_l ()
+    | & _7: MutBorrow.t t_PCellOwn = Any.any_l ()
+    | & _8: MutBorrow.t t_PCellOwn = Any.any_l ()
+    | & _9: MutBorrow.t t_PCellOwn = Any.any_l ()
+    | & _10: MutBorrow.t t_PCellLocalInv = Any.any_l () ]
+     [ return''0 (result:())-> return' {result} ] 
+  
+  type t_Namespaces  =
+    { t_Namespaces__0: () }
```
jhjourdan: /!\ UNSOUND !

creusot-contracts/src/local_invariant.rs
dianegolfouse: Isn't it necessary to clone the `CellInv` ?

dianegolfouse: Ah right, that's not how it works

creusot-contracts/src/local_invariant.rs
dianegolfouse: There is a hack in `creusot/src/backend/clone_map/elaborator.rs` to handle the translation of `Namespace`.
The reason it is done like this is to favor composability of namespaces: a given function using namespaces `N1`, `N2` and `N3` will get the type
```ocaml
type namespace = N1 | N2 | N3 | other of int
```
You are perfectly right about `Namespaces` (plural) though !

jhjourdan: Ok, I see your point now. That being said, I would still prefer if this could be `#[trusted]`, both for documentation purposes and because we never know if some static analysis on type will be added later in Creusot that would look inside the definitions of types.

creusot-contracts/src/local_invariant.rs
dianegolfouse: Ehhh... I don't like using unstable features 😅 Also this function will get called in normal code : don't we have a goal that code compiled in non-creusot should work on stable? (I'm not sure about this @Lysxia?)

I'll try defining an extension trait though.

creusot-contracts/src/local_invariant.rs
dianegolfouse: Yep, thanks!

creusot/src/backend/logic.rs
dianegolfouse: Adding the dependency means that we will try to generate the type right here. This is wrong, because we have not yet analyzed the function to get all the namespaces that it uses.

creusot-contracts/src/local_invariant.rs
dianegolfouse: Technically its useless, but this is an example, that `T`'s invariant will in fact get asserted on `read/write`ing the cell.

creusot-contracts/src/local_invariant.rs
jhjourdan: Ok, this makes sense.

That being said, I think that instead of an extension trait, you should be able to use an `impl<'a, T: LocalInvariantSpec> Ghost<&'a LocalInvariant<T>>`, because luckily, we are in the crate defining `Ghost`.

creusot/src/backend/logic.rs
```diff
@@ -108,6 +112,12 @@ pub(crate) fn translate_logic(ctx: &Why3Generator, def_id: DefId) -> Option<File
     decls.extend(common_meta_decls());
     decls.extend(body_decls);
 
+    if ctx.used_namespaces.get() {
+        let mut new_decls = ctx.generate_namespace_type(namespace_ty);
+        new_decls.extend(std::mem::take(&mut decls));
+        decls = new_decls;
+    }
```
jhjourdan: You can use `Vec::splice` for this.

creusot/src/backend/logic.rs
jhjourdan: I see. This is kind of a dirty hack, but it is difficult to do better.

Perhaps this should be done as part of the code of the expander.... Not sure what's better.

jhjourdan:
I think this is ready to merge. Feel free to merge, and implement my last remark about `splice`, or not, if I'm too picky ;).

creusot-contracts/src/local_invariant.rs
dianegolfouse: Since this is very specific to local invariants, I'm not sure I want this method to appear on the actual `Ghost` type. To me an extension trait communicate this somewhat.

PR #1601, opened 2025-07-03
Remove useless gotos

arnaudgolfouse:
Add a fmir simplification to remove cases like
```
| bb1 = bb2
| bb2 = bb3
| bb3 = bb4
...
```


tests/should_fail/bug/695.coma
```diff
@@ -114,17 +113,15 @@ module M_695__inversed_if [#"695.rs" 7 0 7 78]
     | bb4 = s0
       [ s0 =  [ &_11 <- [%#s695] false ] s1
       | s1 = call_once {branch'0} {_11} (fun (_ret:()) ->  [ &_0 <- _ret ] s2)
-      | s2 = bb7 ]
+      | s2 = bb9 ]
     
-    | bb7 = bb8
+    | bb9 = bb10
```
xldenis: seems like this doesn't quite capture all the intended cases? seems like there's an interaction with loops potentially?

creusot/src/backend/optimization/useless_goto.rs
```diff
@@ -0,0 +1,66 @@
...
+///     ...
+///     goto bb2
+/// bb2:
+///     goto bb3
+/// bb3:
+///     ...
+/// ```
+/// `bb2` is useless here.
+pub(super) fn remove_useless_gotos(body: &mut fmir::Body) {
+    let graph = node_graph(body);
```
jhjourdan, line 20: Caution: you are mutating the FMIR, but not maintaining `graph` up to date.

I think the right way to do this is to use two passes:
1- for each bb, you determine which bb it can shortcut to. That's basically the same thing as path compression in union-find. Be cautious of the no-op loops!!
2- do the substitution

tests/should_fail/bug/695.coma
jhjourdan: We need to make sure the algorithm does the right thing with no-op loops, but appart from that, I don't see why there would be a problem with loops.

In particular, I think the dominance relation should be preserved, so that the weak topological ordering should not change (I think).

xldenis: > I don't see why there would be a problem with loops.

there _shouldn't_ be one but I'm just noting what appears to still be a redundant goto in the graph.

dianegolfouse: Yes, this algorithm is single pass, so it may miss some redundant gotos (depending on the visiting order of the terminators: this is why loops are not correctly optimized). I'm going to implement @jhjourdan proposition below, which should handle those cases.

creusot/src/backend/optimization/useless_goto.rs
dianegolfouse: Like this?

Lysxia (approved):
lgtm

creusot/src/backend/optimization/useless_goto.rs
jhjourdan: That seems to work properly, yes. Thanks.

PR #1612, opened 2025-07-07
Implement SliceIndex for RangeInclusive

### LINKED ISSUE #1609: Implement `SliceIndex` for `RangeInclusive`

Lysxia:
Fix #1609


tests/should_succeed/slices/01.rs
```diff
@@ -20,3 +20,21 @@ pub fn index_mut_slice(a: &mut [u32]) {
...
+    // let s = &mut a[1..=2];
+    //
+    // Error:
+    // error[E0277]: the size for values of type `[u32]` cannot be known at compilation time
+    //   --> 01.rs:27:19
+    //    |
+    // 27 |     let s = &mut a[1..=2];
+    //    |                   ^^^^^^^ doesn't have a size known at compile-time
+    //    |
+    //    = help: the trait `std::marker::Sized` is not implemented for `[u32]`
```
Lysxia: Why do I get this error if I use line 27 instead of 37?

creusot-contracts/src/std/slice.rs
```diff
@@ -146,6 +146,29 @@ impl<T> SliceIndex<[T]> for Range<usize> {
...
+    #[logic]
+    #[open]
+    fn resolve_elswhere(self, old: Seq<T>, fin: Seq<T>) -> bool {
+        pearlite! {
+            forall<i> 0 <= i && (i < self.start_log()@ || self.end_log()@ < i) && i < old.len()
+            ==> old[i] == fin[i]
+        }
+    }
+}
+
```
jhjourdan: That's not that simple. We need to take into account `empty_log`, too!

creusot-contracts/src/std/slice.rs
```diff
@@ -146,6 +146,29 @@ impl<T> SliceIndex<[T]> for Range<usize> {
...
+    #[logic]
+    #[open]
+    fn in_bounds(self, seq: Seq<T>) -> bool {
+        pearlite! { self.start_log()@ <= self.end_log()@ && self.end_log()@ < seq.len() }
+    }
+
+    #[logic]
+    #[open]
+    fn has_value(self, seq: Seq<T>, out: [T]) -> bool {
+        pearlite! { seq.subsequence(self.start_log()@, self.end_log()@ + 1) == out@ }
```
jhjourdan:
```suggestion
        pearlite! { 
            if self.is_empty_log() { out@.is_empty() }
            else { seq.subsequence(self.start_log()@, self.end_log()@ + 1) == out@ }
        }
```

creusot-contracts/src/std/slice.rs
```diff
@@ -146,6 +146,29 @@ impl<T> SliceIndex<[T]> for Range<usize> {
     }
 }
 
+impl<T> SliceIndex<[T]> for RangeInclusive<usize> {
+    #[logic]
+    #[open]
+    fn in_bounds(self, seq: Seq<T>) -> bool {
+        pearlite! { self.start_log()@ <= self.end_log()@ && self.end_log()@ < seq.len() }
+    }
```
jhjourdan:
```suggestion
    #[logic]
    #[trusted]
    #[ensures(self.end_log()@ < seq.len() && self.start_log()@ <= self.end_log()@+1 ==> result)]
    #[ensures(self.end_log()@ >= seq.len() ==> !result)]
    #[ensures(self.end_log()@ < seq.len() && self.is_empty() && self.start_log()@ <= self.end_log()@ ==> result)]
    fn in_bounds(self, seq: Seq<T>) -> bool {
        pearlite! { self.start_log()@ <= self.end_log()@ && self.end_log()@ < seq.len() }
    }
```

Because the `exhausted` field is private, there is not much more we can say. The last `ensures` is rather useless in practice.

Also, this means we cannot prove the access is valid in the case of empty ranges of the form `x..=x-1`.

creusot-contracts/src/std/slice.rs
```diff
@@ -146,6 +146,29 @@ impl<T> SliceIndex<[T]> for Range<usize> {
...
+    fn has_value(self, seq: Seq<T>, out: [T]) -> bool {
+        pearlite! { seq.subsequence(self.start_log()@, self.end_log()@ + 1) == out@ }
+    }
+
+    #[logic]
+    #[open]
+    fn resolve_elswhere(self, old: Seq<T>, fin: Seq<T>) -> bool {
+        pearlite! {
+            forall<i> 0 <= i && (i < self.start_log()@ || self.end_log()@ < i) && i < old.len()
+            ==> old[i] == fin[i]
```
jhjourdan:
```suggestion
            forall<i> 0 <= i && (i < self.start_log()@ || self.end_log()@ < i || self.is_empty_log()) && i < old.len()
            ==> old[i] == fin[i]
```

creusot-contracts/src/std/slice.rs
```diff
@@ -249,6 +272,11 @@ extern_spec! {
         #[ensures(ix.in_bounds(self@) || result == None)]
         fn get<I: SliceIndex<[T]>>(&self, ix: I) -> Option<&<I as ::std::slice::SliceIndex<[T]>>::Output>;
 
+        #[ensures((^self)@.len() == self@.len())]
+        #[ensures(ix.in_bounds(self@) ==> exists<r> result == Some(r) && ix.has_value(self_@, *r) && ix.has_value((^self_)@, ^r) && ix.resolve_elswhere(self_@, (^self_)@))]
```
jhjourdan:
```suggestion
        #[ensures(ix.in_bounds(self@) ==> exists<r> result == Some(r) && ix.has_value(self@, *r) && ix.has_value((^self)@, ^r) && ix.resolve_elswhere(self_@, (^self)@))]
```

tests/should_succeed/slices/01.rs
jhjourdan: That's because `SliceIndex::Output` is not sized, but `has_value` takes it as a parameter. Making it take `&Self::Output` as a parameter should solve the issue.

creusot-contracts/src/std/slice.rs
Lysxia: Doesn't the first clause imply the third?

tests/should_succeed/slices/01.rs
Lysxia: The error is still happening after fixing `has_value`, and I don't see any other culprit.

creusot-contracts/src/std/slice.rs
jhjourdan: Right. The last clause is pointless.

tests/should_succeed/slices/01.rs
jhjourdan: I don't know... Maybe things like `&^self`? Could you try to replace them with `such_that(||true)` to see whether they are the problem we are searching for?

Lysxia: The extern spec for `index_mut` contains `Self: Sized` as an `additional_predicates`...

Lysxia: It comes from the desugaring of `extern_spec`:
```
fn extern_spec_3cd29f59_c659_4fc2_b4a8_19e6892003f8<Self_,
    Idx>(#[allow(unused)] self_: &mut Self_, #[allow(unused)] ix: Idx)
    -> &mut Self_::Output where Idx: ?Sized, Self_: IndexMut<Idx>
```

jhjourdan: There are two bugs here:

1- extern specs for trait methods should have `Self: ?Sized`. I'm not sure whether there are other kinds of extern specs that need this. Could you please investigate?

2- These extern specs for `IndexMut` are useless and should be removed. I'm not sure why they have been added in the first place, but removing them now does not change anything in the output.

jhjourdan: (And btw, I'm somewhat guilty here because I removed `Self: ?Sized` from the extern spec in b5800f6. This was compensating the bugs in `extern_spec` for trait methods.)

Lysxia: Adding `Self: ?Sized` to other traits causes errors when methods take a `self` value. Maybe we should just use it for traits intended to be implemented by unsized types?

jhjourdan: How can you determine this syntactically?

Which are the trait that trigger these errors? I guess they are originally declared as inheriting `Sized`, so it's not absurd to require these extern specs to have this constraint.

creusot-contracts/src/std/slice.rs
```diff
@@ -111,8 +113,8 @@ impl<T> SliceIndex<[T]> for usize {
     #[logic]
     #[open]
     #[creusot::why3_attr = "inline:trivial"]
-    fn has_value(self, seq: Seq<T>, out: T) -> bool {
-        pearlite! { seq[self@] == out }
+    fn has_value(self, seq: Seq<T>, out: &T) -> bool {
+        pearlite! { seq[self@] == *out }
```
jhjourdan: I the end, is this change necessary? If not, then I'd prefer have the one without the reference, it seems cleaner.

tests/should_succeed/slices/01.rs
Lysxia: I mean we add `?Sized` manually when we can.

jhjourdan: Since the default for traits is to not inherit `Sized`, I think we should do the same for extern specs.

creusot-contracts/src/std/slice.rs
Lysxia: Indeed it was not necessary. I've reverted it.

jhjourdan:
I guess this is ready to merge.

jhjourdan:
Thanks!

PR #1613, opened 2025-07-08
Add nth_bit

### LINKED ISSUE #1576: Feature wish: add a built-in operator for the nth-bit in pearlite
For specifying code operating at the bitwise level, it would be very convenient to have a predefined operator `nth` which gives the nth bit of a machine integer. More precisely, return a Boolean telling that the nth bit is set. For example
```
let x : usize = 0x3;
#[assert(nth(x,0))];
#[assert(!nth(x,2))];
```
the 0-th bit is the least significant one. The second parameter of nth could be a Int, or a usize. 

Notice that such an operation is supported by Why3 in the stdlib: `nth` is the one where the second parameter is an int, and `nth_bv` is the one where the second parameter is a bitvector.

Thanks !

Lysxia:
Close #1576


creusot-contracts/src/lib.rs
```diff
@@ -379,7 +379,7 @@ mod base_prelude {
     pub use crate::{
         fn_pure::FnPure,
         ghost::Ghost,
-        logic::{Int, OrdLogic, Seq, ops::IndexLogic as _},
+        logic::{Int, OrdLogic, Seq, ops::IndexLogic as _, ops::NthBitLogic as _},
```
jhjourdan, line 382: Do we really want to include this in the prelude? It does not seem so common to me.

Lysxia: :shrug: I can remove it.

PR #1619, opened 2025-07-09
Add ?Sized bound to extern_spec for traits

### LINKED ISSUE #1617: Make extern specs for traits produce a `?Sized` bound

Lysxia:
Fix #1617 

Also wrap Pearlite identifiers in `*&` to avoid issues with unsized arguments in spec closures.

### DIFF
--- a/creusot-contracts/src/pcell.rs
+++ b/creusot-contracts/src/pcell.rs
@@ -111,7 +111,7 @@ impl<T> PCell<T> {
     #[trusted]
     #[requires(self.id() == perm.id())]
     #[ensures(val == (^perm.inner_logic())@)]
-    #[ensures(resolve(&(*perm.inner_logic())@))]
+    #[ensures(resolve((*perm.inner_logic())@))]
     #[ensures(self.id() == (^perm.inner_logic()).id())]
     pub unsafe fn set(&self, perm: Ghost<&mut PCellOwn<T>>, val: T) {
         let _ = perm;
--- a/creusot/src/backend/closures.rs
+++ b/creusot/src/backend/closures.rs
@@ -282,8 +282,7 @@ pub(crate) fn closure_post<'tcx>(
     // Make sure fn_once and fn_mut_once are satisfied
     post = to_resolve.iter().fold(post, |p, r| {
         if let Some((id, subst)) = ctx.resolve(typing_env, r.ty) {
-            let r = r.clone().shr_ref(ctx.tcx);
-            p.conj(Term::call_no_normalize(ctx.tcx, id, subst, [r]))
+            p.conj(Term::call_no_normalize(ctx.tcx, id, subst, [r.clone()]))
         } else {
             p
         }
@@ -316,7 +315,7 @@ pub(crate) fn closure_resolve<'tcx>(
     let typing_env = TypingEnv::non_body_analysis(ctx.tcx, def_id);
     for (ix, ty) in csubst.upvar_tys().iter().enumerate() {
         if let Some((id, subst)) = ctx.resolve(typing_env, ty) {
-            let proj = self_.clone().proj(ix.into(), ty).shr_ref(ctx.tcx);
+            let proj = self_.clone().proj(ix.into(), ty);
             resolve = Term::call(ctx.tcx, typing_env, id, subst, [proj]).conj(resolve);
         }
     }
--- a/creusot/src/backend/logic/vcgen.rs
+++ b/creusot/src/backend/logic/vcgen.rs
@@ -482,7 +482,6 @@ impl<'tcx> VCGen<'_, 'tcx> {
                         self.build_pattern(&arm.0, &|pat| Ok((pat, self.build_wp(&arm.1, k)?)))
                     })
                     .collect::<Result<Box<[_]>, _>>()?;
-
                 Ok(scrut.match_(arms))
             }),
             // VC(let P = A in B, Q) = VC(A, |a| let P = a in VC(B, Q))
--- a/creusot/src/backend/structural_resolve.rs
+++ b/creusot/src/backend/structural_resolve.rs
@@ -1,11 +1,10 @@
 use rustc_ast::Mutability;
 use rustc_middle::ty::{Ty, TypingEnv};
-use rustc_span::DUMMY_SP;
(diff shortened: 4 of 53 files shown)


creusot-contracts/src/std/ops.rs
```diff
@@ -175,19 +175,22 @@ extern_spec! {
             trait FnOnce<Args> where Args: Tuple {
                 #[requires(self.precondition(arg))]
                 #[ensures(self.postcondition_once(arg, result))]
-                fn call_once(self, arg: Args) -> Self::Output;
+                fn call_once(self, arg: Args) -> Self::Output
+                    where Self: Sized;
```
Lysxia: This is required by the `requires`/`ensures`. Do we want `precondition` and `postcondition` to take `&self` instead?

jhjourdan: Why does the precondition require this? `FnOnceExt` does not inherit `Sized`.

jhjourdan: Oh but probably the blanket *impl* assumes this. What if you add `?Sized` to it?

Lysxia:
```
error[E0277]: the size for values of type `Self_` cannot be known at compilation time
   --> creusot-contracts/src/std/ops.rs:177:27
    |
176 |                 #[requires(self.precondition(arg))]
    |                 - this type parameter needs to be `Sized`
177 |                 #[ensures(self.postcondition_once(arg, result))]
    |                           ^^^^ doesn't have a size known at compile-time
    |
    = note: all values captured by value by a closure must have a statically known size
help: consider removing the `?Sized` bound to make the type parameter `Sized`
    |
175 -             trait FnOnce<Args> where Args: Tuple {
176 -                 #[requires(self.precondition(arg))]
175 +             trait [requires(self.precondition(arg))]
    |
```

creusot-contracts-proc/src/creusot/pretyping.rs
```diff
@@ -122,8 +122,9 @@ pub fn encode_term(term: &RT) -> Result<TokenStream, EncodeError> {
                         quote_spanned! {sp=> *::creusot_contracts::__stubs::old( #(#args),* ) },
                     );
                 }
+                // Don't wrap function calls in `*&`.
+                return Ok(quote_spanned! {sp=> #func (#(#args),*)});
             }
-
             let func = encode_term(func)?;
             Ok(quote_spanned! {sp=> #func (#(#args),*)})
```
jhjourdan: Isn't this dead code ?

creusot-contracts-proc/src/creusot/pretyping.rs
```diff
@@ -200,7 +201,12 @@ pub fn encode_term(term: &RT) -> Result<TokenStream, EncodeError> {
             });
             Ok(tokens)
         }
-        RT::Path(_) => Ok(quote_spanned! {sp=> #term }),
+        RT::Path(path) => Ok(if path.inner.path.segments.len() == 1 {
+            // Trick to avoid capturing unsized arguments in spec closures.
+            quote_spanned! {sp=> (*&#term) }
+        } else {
+            quote_spanned! {sp=> #term }
+        }),
```
jhjourdan: Why do we do that only when the path len is 1? I understand that it is unnecessary if this is a global ident, but isn't uniformity better?

jhjourdan:
That's very nice.

Out of curiosity, does this change allow us to remove the borrow in the type of `resolve`? That would be very nice!

Similarly, if we could expand `forall<x: T> P(x)` into `stubs::forall(|x: &T| let x = *x; P(x)`, that would be very nice as well...

jhjourdan:
And could we also support cleanly sequences, mappings, etc... of unsized objects?

jhjourdan:
(essentially reverting the second part of #1600, but cleanly.)

creusot-contracts-proc/src/creusot/pretyping.rs
Lysxia: The `if` above handles `ident()`, but you could have `(complicated expression)()`

jhjourdan: I don't think we support anything liek this in Pearlite.

jhjourdan:
I merged master, but I'd still like if you could answer (not necessarily address) the two concerns I have above.

creusot-contracts-proc/src/creusot/pretyping.rs
Lysxia: That's true. I can remove this branch.

creusot-contracts-proc/src/creusot/pretyping.rs
Lysxia: I'll try undoing this. I did this at first when I was hoping to only modify `creusot-contracts-proc` and not `creusot`, i.e., I wasn't explicitly removing `*&` in the THIR-to-Pearlite pass yet, and I was trying to be more conservative about where to insert `*&` without breaking creusot.

Lysxia:
> And could we also support cleanly sequences, mappings, etc... of unsized objects?

The keys of mappings (fmap, mapping) and sets can often be left unsized.

This is more difficult for `Seq` and the codomain of `Fmap` and `Mapping`. The main challenge is that functions must return sized values, and that especially affects indexing. Uses of `Option` and pairs `(K, V)` also incur a `Sized` constraint. We could maybe reduce the need for `Sized` by relying more on a membership relation (`contains`) instead of an indexing function, but I'm leaving that for another day (in particular, `Seq` currently defines `contains` using indexing, and changing that might modify the output coma a lot).

Lysxia:
> Out of curiosity, does this change allow us to remove the borrow in the type of `resolve`?

That seems to work but please double check the change!

> if we could expand `forall<x: T> P(x)` into `stubs::forall(|x: &T| let x = *x; P(x)`

Done in the meantime in https://github.com/creusot-rs/creusot/pull/1634

jhjourdan:
> This is more difficult for Seq and the codomain of Fmap and Mapping. 

Thanks for looking at this. I guess tht we can stay in the current status quo, then.

> That seems to work but please double check the change!

Done. I found a few places where changes were missing. I hope there is no longer issues any more.

Issue #754, opened 2023-04-26
Allow variants to be specified on loops


jhjourdan:
Nowadays, it says:

```
error: expected `fn`
   --> /home/jjourdan/verifythis_creusot/challenge_1.rs:111:9
    |
111 |         while l != NULL {
    |    
```

xldenis:
Fixed in #755

xldenis:
Sounds like a why3 problem, I'll investigate

xldenis:
Waiting on https://gitlab.inria.fr/why3/why3/-/merge_requests/873

xldenis:
This was easy to implement before but is now much harder in Coma where there is no simple notion of variants. We need to talk with the Coma developers and come up with a solution here.

jhjourdan:
We need to add support for variants on recursive functions, too.

jhjourdan:
This is blocked on support for variants in coma. Or we do this ourselves in Creusot.

jhjourdan:
TODO for later : automatic termination of iterators that are knwon to always terminate (specific trait ? `ExactSizeIterator` ?)

PR #1622, opened 2025-07-09
Variants

### LINKED ISSUE #561: Clarify the status of variants/Well-founded
The status of `Well-founded` and variants is unclear.

Today, it is used for `Int`, which is not right, because integers are not well-founded. But this is more-or less sound because Why3 adds the requirement that the variant need to be positive.

We need to do something close to Why3, an allow variants in the following cases:
- The type is well-founded. In the case, tell Why3 that we are using our specific order relation (I don't think we are doing that right now, do we?)
- The type is Int, and let Why3 add the requirement that the variant needs to be positive
- The type is algebraic, and not marked as well-founded. Do as in Why3: do structural recursion.
- Tuples with lexicographic ordering of the previous cases? Is this even possible to tell Why3 to use different kinds of orderings on different elements of the tuple?

arnaudgolfouse:
This allow variants in various places:
- simply recursive functions (logic and program)
- loops (this allows loops in ghost code!)

Together with a rework of the `WellFounded` trait, to allow variants on custom types.

Fixes #561, #754.


creusot-contracts/src/well_founded.rs
```diff
@@ -1,41 +1,90 @@
...
+/// Instances of this trait are types which are allowed as variants of recursive definitions.
+pub trait WellFounded: Sized {
+    /// Relation used to specify well-foundedness.
+    #[logic]
+    #[rustc_diagnostic_item = "creusot_wf_relation"]
+    fn well_founded_relation(self, other: Self) -> bool;
+
+    /// Being well-founded means that there is no infinitely decreasing sequence.
+    #[logic]
+    #[requires(forall<i: Int> i >= 0 ==> Self::well_founded_relation(s[i], s[i + 1]))]
```
jhjourdan: It's arguable whether we may want to put `i` as the return value of the function, ensuring `result >= 0` and `!Self::well_founded_relation(s[result], s[result + 1])`. And that's very easy to prove by induction on `s[0]` as soon as you have an induction principle on `Self`, so it's natural to see that we are indeed building an induction principle.

creusot-contracts/src/well_founded.rs
```diff
@@ -1,41 +1,90 @@
...
+    forall<i: Int> i >= 0 ==> r[(s[i], s[i + 1])]
+)]
+pub fn well_founded_map_to_nat<T>(r: Mapping<(T, T), bool>, m: Mapping<T, Int>) {
+    pearlite! {
+        if exists<s: Mapping<Int, T>> forall<i: Int> i >= 0 ==> r[(s[i], s[i + 1])] {
+            let s = such_that(|s: Mapping<Int, T>| forall<i: Int> i >= 0 ==> r[(s[i], s[i + 1])]);
+            Int::no_infinite_decreasing_sequence(|i: Int| m[s[i]]);
+        }
+    }
+}
```
jhjourdan: Why restricting to integers? Any increasing map to a well-founded relation builds a well-founded relation in the domain. If you restrict to integers, then you restrict to building well-founded relations of ordinal measure ω, which is very low and restrictive (no lexicographic orders, for example).

Also, to ease use and proof of this lemma, you can pass `s` as a parameter (and perhaps return `i` f you decide to change the definition of `WellFounded` too:

```suggestion
/// This provides an easy way to prove the [`WellFounded`] condition, provided
/// that you can map your type to the integers.
#[logic]
#[requires(forall<t1: T, t2 : T> r[(t1, t2)] ==> U::well_founded_relation(m[t1], m[t2]))]
#[requires(forall<i: Int> i >= 0 ==> r[(s[i], s[i + 1])])]
#[ensures(false)]
pub fn well_founded_map_to_nat<T, U:WellFounded>(r: Mapping<(T, T), bool>, m: Mapping<T, U>, s: Mapping<Int, T>) {
    pearlite! {
            U::no_infinite_decreasing_sequence(|i: Int| m[s[i]]);
    }
}
```

creusot-contracts/src/well_founded.rs
```diff
@@ -1,41 +1,90 @@
...
+
+impl WellFounded for Int {
+    #[logic]
+    #[open]
+    fn well_founded_relation(self, other: Self) -> bool {
+        self >= 0 && self > other
+    }
+
+    #[trusted] // FIXME: find the right why3 lemma
```
jhjourdan: I don't think that this lemma exists. But it's easy by induction on `s[0]`, and you have access to that in Why3 IDE.

creusot-contracts/src/well_founded.rs
```diff
@@ -1,41 +1,90 @@
...
+            fn well_founded_relation(self, other: Self) -> bool {
+                self > other
+            }
+
+            #[logic]
+            #[requires(forall<i: Int> i >= 0 ==> Self::well_founded_relation(s[i], s[i + 1]))]
+            #[ensures(false)]
+            fn no_infinite_decreasing_sequence(s: Mapping<Int, Self>) {
+                pearlite! {
+                    let map_to_nat = |x: Self| x@ + $t::MAX@ + 1;
```
jhjourdan:
```suggestion
                    let map_to_nat = |x: Self| x@ - $t::MIN@;
```

creusot-contracts/src/well_founded.rs
jhjourdan: BTW, the proof is trivial enough when seen this way that having a lemma for that is probably superfluous.

why3/src/declaration.rs
```diff
@@ -91,7 +90,7 @@ impl Condition {
 pub struct Contract {
     pub requires: Box<[Condition]>,
     pub ensures: Box<[Condition]>,
-    pub variant: Option<Exp>,
+    pub variant: Option<(Exp, Type)>,
```
jhjourdan: What's the purpose of this change? It seems like the type is not used anywhere in the `why3` crate.

creusot/src/backend/optimization/useless_goto.rs
```diff
@@ -22,11 +42,7 @@ pub(super) fn remove_useless_gotos(body: &mut fmir::Body) {
     let mut shortcuts_to = IndexMap::new();
     for bb in graph.nodes() {
         let block = &body.blocks[&bb];
-        if block.stmts.is_empty()
-            && block.variant.is_none()
-            && block.invariants.is_empty()
-            && let Terminator::Goto(to_bb) = block.terminator
-        {
+        if let Some(to_bb) = is_empty_block(block) {
```
jhjourdan, line 37: Frankly, I think I prefer the previous version.

creusot/src/backend/program.rs
```diff
@@ -947,12 +973,67 @@ impl<'tcx> Block<'tcx> {
...
+            let variant_decreases =
+                pearlite::Term::call_no_normalize(lower.ctx.tcx, wf_relation, subst, [
+                    pearlite::Term::var(v.old_name, v.term.ty),
+                    v.term.clone(),
+                ]);
+
+            let variant_assertion = lower_pure(lower.ctx, lower.names, &variant_decreases)
+                .with_attr(Attribute::Attr("expl:loop variant".to_string()));
+            for s in &mut statements {
+                assert_before(&mut s.body, &lower.block_idents[&v.loop_head], &variant_assertion);
```
jhjourdan: I don't like the idea of patching the generated code after the fact. I think this is bad style and asking for trouble because you assume the code is in the form you have generated it in. Instead, I would really prefer generating variant assertions on the fly.

Since the only part of the block that need to be considered is the terminator, why wouldn't we do that as part of translating terminators?

Another solution would be to create a new coma handler for any block containing a variant. This handler would check the variant decrease, and jump to the main handler. Any jump to this block would be translated to either a jump to the main handler (if it comes from outside the component) or to this new "variant handler" if it comes from inside the component. A somewhat hacky but probably easy to do this is the make the "variant handler" shadow the main handler within the component (i.e., use the same name for both handlers).

creusot/src/backend/signature.rs
```diff
@@ -111,6 +111,8 @@ pub(crate) fn lower_contract<'tcx, N: Namer<'tcx>>(
         contract.requires.into_iter().map(|cond| lower_condition(ctx, names, cond)).collect();
     let ensures =
         contract.ensures.into_iter().map(|cond| lower_condition(ctx, names, cond)).collect();
-    let variant = contract.variant.map(|term| lower_pure(ctx, names, &term));
+    let variant = contract
+        .variant
+        .map(|term| (lower_pure(ctx, names, &term), translate_ty(ctx, names, term.span, term.ty)));
```
jhjourdan: Since we are checking variants ourselves, I don't think we should forward variant annotations to Why3. That's both redundant, and probably incorrect becase our well foundedness notion is incompatible with Why3's.

Instead, when translating a variant-based defintiion, we should check the variant decrease using the VC gen, and generate an axiom (but I think we are already doing that). This code seems useless to me.

creusot/src/translation/fmir.rs
```diff
@@ -264,10 +317,26 @@ pub struct Invariant<'tcx> {
...
+    pub(crate) loop_head: BasicBlock,
+}
+
 #[derive(Clone, Debug)]
 pub struct Block<'tcx> {
     pub(crate) invariants: Vec<Invariant<'tcx>>,
-    pub(crate) variant: Option<Term<'tcx>>,
+    /// This is the list of variants that should be checked before `continue`ing
+    /// a loop.
+    pub(crate) variants: Vec<Variant<'tcx>>,
```
jhjourdan: Why a vec? Could we have several variants?

creusot/src/translation/function.rs
```diff
@@ -72,7 +72,22 @@ struct BodyTranslator<'a, 'tcx> {
...
+    /// The basic block is the loop entry. It then contains a series of
+    /// invariants, each with a possible description for Why3.
+    invariants: IndexMap<BasicBlock, Vec<(String, Term<'tcx>)>>,
+    /// Maps a basic block to a list that contains:
+    /// - a term that should be checked to have decreased
+    /// - the name for the old value of the term
+    /// - the loop head.
+    ///
+    /// See [`InvariantsAndVariants::variants`](crate::gather_spec_closure::InvariantsAndVariants).
+    variants: IndexMap<BasicBlock, Vec<(Term<'tcx>, Ident, BasicBlock)>>,
```
jhjourdan: Why a vec ?

creusot/src/gather_spec_closures.rs
```diff
@@ -147,34 +161,66 @@ impl<'a, 'tcx> Visitor<'tcx> for InvariantsVisitor<'a, 'tcx> {
...
+                    LoopSpecKind::Variant => {
+                        let mut has_met_before_loop = false;
+                        let preds = self.body.basic_blocks.predecessors()[target]
+                            .iter()
+                            .filter(|p| {
+                                let res = !self.before_loop.contains(*p);
+                                if !res && std::mem::replace(&mut has_met_before_loop, true) {
+                                    self.ctx.dcx().span_bug(self.body.span, "a basic block is marked as the entry for multiple loops: this should not happen.")
+                                }
+                                res
```
jhjourdan: That's *very* fragile. This is essentially assuming that a loop header is always following immediately a block annotated with `before_loop`. Thus, this rely on rustc not inserting intermediary no-op blocks, etc...

I don't think this is a good design. predecessors should be classified based on the WTO.

creusot/src/translation/fmir.rs
```diff
@@ -65,6 +64,47 @@ impl<'tcx> Place<'tcx> {
             None
         }
     }
+
+    fn into_term(self, tcx: TyCtxt<'tcx>, locals: &LocalDecls<'tcx>, span: Span) -> Term<'tcx> {
```
jhjourdan: I don't like this much, both because this means that we are introducing yet another translation for places (with maybe subtle semantic differences), and because the resulting code will have some duplication.

Instead, could we not add to FMir the possibility to annotate calls with a variant to check, before substituting the parameters ? In the backend, it should be easy to check the variant when translating the call.

jhjourdan: An alternative is to generate a separate handlers for recursive calls, with one additional precondition: the fact that the variant decreased.

creusot/src/validate/terminates.rs
```diff
@@ -68,22 +68,42 @@ use rustc_trait_selection::traits::{
...
 ///
 /// Note that for logical functions, these are relaxed: we don't check loops, nor simple
 /// recursion, because why3 will handle it for us.
-pub(crate) fn validate_terminates(ctx: &TranslationCtx) {
+///
+/// # Returns
+///
+/// For each function with a `#[variant]`, returns the called function before
+/// which the variant should have decreased.
+/// FIXME: how do we ensure that the called function uses the _same_ variant?
```
jhjourdan: I don't see your point. Variants are terms depending on values of parameters. If there is a recursive call (possibly mutual), we check that the value of this term decreases. There is no notion of "same variant".

creusot/src/validate/terminates.rs
```diff
@@ -96,23 +116,31 @@ pub(crate) fn validate_terminates(ctx: &TranslationCtx) {
...
+            }
+            recursive_calls.entry(def_id).or_default().insert(def_id);
         };
     }
 
     // detect mutual recursion
     let cycles = tarjan_scc(&call_graph);
     for cycle in cycles {
+        // FIXME: if there is a variant on all the functions in the cycle, then
+        // - Insert the recursive call in `recursive_calls`
```
jhjourdan, line 127: Somehow we eed to check that variants all have the same type. This is tricky because this involves substitutions.

creusot/src/validate/terminates.rs
```diff
@@ -450,6 +478,11 @@ impl CallGraph {
             if !(is_pearlite(ctx.tcx, def_id) || ctx.sig(def_id).contract.terminates) {
                 // Only consider functions marked with `terminates`: we already ensured
                 // that a `terminates` functions only calls other `terminates` functions.
+                if let Some(variant) = &ctx.sig(def_id).contract.variant {
+                    ctx.dcx().struct_span_warn(variant.span, "variant will be ignored")
```
jhjourdan, line 460: ignored, really ? It seems to me that the variant will be checked, but it is useless to prove the termination of the function.

jhjourdan (commented):
Great job!

All in all, I think that most of this is OK, but the way variant assertion are inserted need to be improved:

- For loops, it should really rely on the WTO, and therefore I think that this should really be done by the backend. In particular, I'm not sure that we should store sets of variants to check at "continue" blocks. Instead, when entering a WTO component whose head has a variant, create a new handler for the head of the loop that will check the variant decrease, and make all the jumps from inside the component point to this new handler.

- For recursive calls in program code, I would really prefer not translating operands to terms. This would be yet another translation from places to XXX, and I'm afraid of subtle semantic differences betwen the versions. Instead, please use a method similar to the one for loop I propose above: create a new sub-handler for recursive calls, and make it check that the variant decreases before performing the actual recursive call.

why3/src/declaration.rs
dianegolfouse: It's used in `creusot/src/backend/program.rs`, when defining the variable that holds the variant.

creusot-contracts/src/well_founded.rs
dianegolfouse: Indeed.

creusot/src/backend/optimization/useless_goto.rs
dianegolfouse: Ah, yeah I did it this way because I originally forgot to check one of the fields (resulting in an unsound optimization)... This new version would have caught it, because the pattern matching is exhaustive.

creusot/src/backend/program.rs
dianegolfouse: I implemented the second solution, which works pretty well IMO

creusot/src/translation/fmir.rs
dianegolfouse: I did your second solution, which is somewhat similar to loop variants.

creusot/src/validate/terminates.rs
dianegolfouse: It is ignored, because we don't even do the analysis in this case. Thus we don't know what recursive calls the function will make, which is needed to know where to insert the variant checks.

creusot/src/validate/terminates.rs
dianegolfouse: I was thinking about having the same type here: I added a comment in the FIXME comment below.

creusot-contracts/src/logic/well_founded.rs
```diff
@@ -1,42 +1,93 @@
...
+    ///     #[logic]
+    ///     #[ensures(!Self::well_founded_relation(s[result], s[result + 1]))]
+    ///     fn no_infinite_decreasing_sequence(s: Mapping<Int, Self>) -> Int {
+    ///         Int::no_infinite_decreasing_sequence(|i| s[i].0)
+    ///     }
+    /// }
+    /// ```
+    #[logic]
+    #[ensures(!Self::well_founded_relation(s[result], s[result + 1]))]
+    fn no_infinite_decreasing_sequence(s: Mapping<Int, Self>) -> Int;
```
jhjourdan, line 38: You need to add `#[ensures(result >= 0)]`. Otherwise, it's both useless in practice and trivial to prove for integers.

creusot-contracts/src/logic/well_founded.rs
```diff
@@ -1,42 +1,93 @@
...
+}
+
+impl WellFounded for Int {
+    #[logic]
+    #[open]
+    fn well_founded_relation(self, other: Self) -> bool {
+        self >= 0 && self > other
+    }
+
+    #[trusted] // FIXME: find the right why3 lemma
```
jhjourdan:
```suggestion
    #[trusted]
```

We don't actually care.

creusot-contracts/src/logic/well_founded.rs
```diff
@@ -1,42 +1,93 @@
...
+            }
+        }
+
+        )*
+    };
+}
+
+impl_well_founded!(u8, u16, u32, u64, u128, usize, i8, i16, i32, i64, i128, isize);
+
+impl<T: WellFounded> WellFounded for &T {
```
jhjourdan, line 84: An instance for `Box` and tuples (with lexicographic ordering), at least, would be great.

jhjourdan:
(I'm in the process of reviewing this.)

creusot/src/backend/signature.rs
jhjourdan: Still a TODO.

creusot/src/analysis.rs
```diff
@@ -626,16 +636,28 @@ impl<'a, 'tcx> Analysis<'a, 'tcx> {
...
+                let scope = self.resolver.body.source_info(bb.start_location()).scope;
+                let subst = self.analysis_env.inline_pearlite_subst(tcx, scope);
+                variant.subst(&subst);
+                self.analysis_env.check_use_in_logic(
+                    variant,
+                    tcx,
+                    self.resolver.move_data(),
+                    &bad_vars,
+                );
+            }
```
jhjourdan: You could avoid duplication by:
1- chaining the two iterators
2- peeking in the iterator to see if it's not empty
3- initialize `bad_vars` and `subst`
4- iterate on the chained iterator

creusot/src/backend/logic/vcgen.rs
```diff
@@ -717,35 +738,35 @@ impl<'tcx> VCGen<'_, 'tcx> {
...
-        variant_ty: Ty<'tcx>,
-        span: Span,
+        variant_after: Term<'tcx>,
     ) -> Result<Exp, VCError<'tcx>> {
+        let wf_relation = get_wf_relation(self.ctx.tcx);
+        let variant_subst = self.ctx.tcx.mk_args(&[variant_after.ty.into()]);
+        let (wf_relation, variant_subst) =
+            TraitResolved::resolve_item(self.ctx.tcx, self.typing_env, wf_relation, variant_subst)
+                .to_opt(wf_relation, variant_subst)
+                .unwrap_or((wf_relation, variant_subst));
```
jhjourdan: In which case should we expect a `None` here? Shouldn't we fail instead?

creusot/src/backend/program.rs
```diff
@@ -164,13 +165,103 @@ pub(crate) fn to_why<'tcx, N: Namer<'tcx>>(
...
+            let subst = ctx.tcx.mk_args(&[variant_expr.ty.into()]);
+            let wf_relation = get_wf_relation(ctx.tcx);
+            let (wf_relation, subst) = TraitResolved::resolve_item(
+                ctx.tcx,
+                ctx.typing_env(body_id.def_id),
+                wf_relation,
+                subst,
+            )
+            .to_opt(wf_relation, subst)
+            .unwrap_or((wf_relation, subst));
```
jhjourdan: Same thing here: if the return value is `None`, then we should error.

creusot/src/backend/program.rs
```diff
@@ -164,13 +165,103 @@ pub(crate) fn to_why<'tcx, N: Namer<'tcx>>(
...
+            let variant_decreases = pearlite::Term::call_no_normalize(
```
jhjourdan: I don't see why the return type of the call should be normalized. Please use `Term::call` instead.

creusot/src/backend/program.rs
```diff
@@ -249,55 +353,131 @@ pub fn why_body<'tcx, N: Namer<'tcx>>(
...
+    let variant = block.variant.clone().map(|variant| {
+        let wf_relation = get_wf_relation(lower.ctx.tcx);
+        let typing_env = lower.ctx.typing_env(lower.def_id);
+        assert_eq!(GenericArgs::identity_for_item(lower.ctx.tcx, wf_relation).len(), 1); // sanity check
+        let subst = lower.ctx.tcx.mk_args(&[variant.term.ty.into()]);
+
+        let (wf_relation, subst) =
+            TraitResolved::resolve_item(lower.ctx.tcx, typing_env, wf_relation, subst)
+                .to_opt(wf_relation, subst)
+                .unwrap_or((wf_relation, subst));
```
jhjourdan: Idem

creusot/src/backend/program.rs
```diff
@@ -249,55 +353,131 @@ pub fn why_body<'tcx, N: Namer<'tcx>>(
...
+        let variant_decreases = pearlite::Term::call_no_normalize(
```
jhjourdan, line 402: Idem

creusot/src/backend/program.rs
```diff
@@ -1105,9 +1300,24 @@ impl<'tcx> Statement<'tcx> {
...
                 let ty = lower.ty(ty);
+                if lower.ctx.should_check_variant_decreases(lower.def_id, fun_id) {
+                    if subst.iter().any(|p| {
+                        matches!(
+                            p.kind(),
+                            rustc_type_ir::GenericArgKind::Type(_)
+                                | rustc_type_ir::GenericArgKind::Const(_)
+                        )
+                    }) {
+                        lower.ctx.dcx().span_err(self.span, "Cannot make a recursive call that substitute some of its type parameters");
```
jhjourdan:
```suggestion
                        lower.ctx.dcx().span_err(self.span, "Polymorphic recursion is not supported: recursive calls should have the same type parameters.");
```

jhjourdan (changes requested):
This is great work. I have a few comments here and there, and two other concerns:

- the definition of well foundedness is broken (see my comment)

- it would be great to demonstrate this new features on some of the already existing examples : PA and union find, of course, and perhaps binary search or other examples. This should not be so difficult, all the needed ingredients are already there.

creusot-contracts/src/logic/well_founded.rs
dianegolfouse: Done!

jhjourdan:
If you feel this is ready to merge, then please press the button ;)

PR #1627, opened 2025-07-12
Add Why3's meta select_lsinst all to all goals

jhjourdan:
This forces Why3 to monomorphise more things, and therefore improves provability.

Usually, the risk is exponential blowup in the size of goals sent to SMT solvers. However, we are using very little polymorphism, so there is little risk of blowup in the case of Creusot. On the other hand, when mappings are not fully monomorphised, provers don't know what to do with them.


tests/should_succeed/bdd/proof.json
```diff
@@ -112,7 +112,7 @@
                         "tactic": "split_vc",
                         "children": [
                           { "prover": "cvc5@1.0.5", "time": 0.022 },
-                          { "prover": "cvc5@1.0.5", "time": 0.514 }
+                          { "prover": "cvc5@1.0.5", "time": 1.5 }
```
xldenis, line 115: Has this change really helped provers? As far as I can see the only impact it had was increasing the runtimes.

jhjourdan: I have examples which improve a lot, yes.

It's not very surprising that provers go slightly slower, because the goal is larger. But I don't expect them to fail where they succeed before.

jhjourdan: Also, my machine is quite slow, so it's not a good benchmark.

PR #1639, opened 2025-07-25
Lint for usage of `#[trusted]`

arnaudgolfouse:
Allow using `#![forbid(creusot::trusted_code)]` to disallow usage of `#[trusted]` in the current crate.


jhjourdan:
Good idea, thanks!

PR #1641, opened 2025-07-26
Improve caching in CI

### LINKED ISSUE #1638: CI: Is the opam cache correctly used?
It seems like something is wrong with the opam cache. For example if we look at this recent job: https://github.com/creusot-rs/creusot/actions/runs/16522748786/job/46728236044

In `Install Opam`, it seems like the steps `[restoring|saving] opam cache` failed? Which causes the steps to take half the total running time :(

jhjourdan:
Fixes #1638.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -25,19 +25,19 @@ jobs:
           name: guide
           path: |
             guide/book/
-  
+
   build-docs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-    - uses: actions/cache@v4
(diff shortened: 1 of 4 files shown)


jhjourdan:
Question: I disabled the "quiet" mode of the Why3 CI task, because I like knowing what test is currently running. The down side is that it is slightly less convenient to find out which test failed when so.

I don't know who set up this, but I'm open to reverting this if there are some strong opinions.

jhjourdan:
@arnaudgolfouse , so, any objection to this?

dianegolfouse:
Nope! (neither for the disabling of "quiet" mode, I agree with you here)

Issue #1643, opened 2025-07-30
Prove fails when importing a logic function from another module


ethanbb:
It seems like I can't import a logic function in the same crate to use in contracts. Still new to Rust, so I might be misunderstanding how this is supposed to work, but here's what I'm doing:

file: predicates.rs
```rust
use creusot_contracts::*;

#[logic]
pub fn inbounds2(a: Seq<Int>, i: Int) -> bool {
    i >= 0 && i < a.len()
}
```

file: lib.rs
```rust
mod predicates;

use creusot_contracts::*;
use crate::predicates::*;


#[logic]
pub fn inbounds1(a: Seq<Int>, i: Int) -> bool {
    i >= 0 && i < a.len()
}


#[requires(a@.len() > 0 && a@.len() < usize::MAX@)]
#[ensures(inbounds1(a.deep_model(), result@))]
#[ensures(forall<k> inbounds1(a.deep_model(), k) ==> a@[result@]@ >= a@[k]@)]
pub fn find_max1(a: &[i32]) -> usize {
    let mut maxi = 0;
    let mut i = 1;

    #[invariant(i@ > 0 && i@ <= a@.len())]
    #[invariant(maxi@ >= 0 && maxi@ < i@)]
    #[invariant(forall<k> k >= 0 && k < i@ ==> a@[maxi@]@ >= a@[k]@)]
    while i < a.len() {
        if a[i] > a[maxi] {
            maxi = i;
        }                
        i = i + 1;
    }
    maxi
}


#[requires(a@.len() > 0 && a@.len() < usize::MAX@)]
#[ensures(inbounds2(a.deep_model(), result@))]
#[ensures(forall<k> inbounds2(a.deep_model(), k) ==> a@[result@]@ >= a@[k]@)]
pub fn find_max2(a: &[i32]) -> usize {
    let mut maxi = 0;
    let mut i = 1;

    #[invariant(i@ > 0 && i@ <= a@.len())]
    #[invariant(maxi@ >= 0 && maxi@ < i@)]
    #[invariant(forall<k> k >= 0 && k < i@ ==> a@[maxi@]@ >= a@[k]@)]
    while i < a.len() {
        if a[i] > a[maxi] {
            maxi = i;
        }                
        i = i + 1;
    }
    maxi
}
```

Output of `cargo creusot prove`:
```
    Updating crates.io index
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.28s
Warning: found dangling files and directories:
[...]

PR #1649, opened 2025-08-01
Guide: Update language about logic functions and document #[open]

ethanbb:
Following up on conversation in #1643. The first section of the "Logic functions" page was out of date, also mentioning `#[predicate]` which has been removed, plus it was a bit unclear about what is allowed where within a logic function. I rewrote that part with what I believe are the correct rules but please check. Also, I saw that #1622 would maybe change the situation re: loops? If so, that will need to be changed soon.

I added another paragraph documenting `#[open]` and saying a little about how it relates to how the logic function is used, e.g. as a lemma or as a predicate. I do think it might be useful to say more about lemmas somewhere and note how/when to use them and that actually any function can be considered a lemma if its pre/postconditions are useful for the prover but its body doesn't do anything useful for the program. I just realized this myself, maybe it is obvious if you have more experience working with automated provers though. For what it's worth, I found the [Dafny tutorial](https://dafny.org/latest/OnlineTutorial/Lemmas) on lemmas very helpful (actually I would have been totally lost trying to learn Creusot if I hadn't gone through the Dafny guides first, maybe it's worth recommending?)

### DIFF
--- a/guide/src/logic_functions.md
+++ b/guide/src/logic_functions.md
@@ -1,13 +1,11 @@
 # Logic functions
 
-We provide two new attributes on Rust functions: `logic` and `predicate`.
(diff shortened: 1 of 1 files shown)


dianegolfouse:
Thanks a lot! There are things I want to reword, but I'll merge this and do my changes on top.

RE: lemmas: I think we should spend some time explaining this, yes... I think I'll add the Dafny tutorial in a footnote for now.

PR #1652, opened 2025-08-01
Proof modes

arnaudgolfouse:
This is essentially a renaming of `#[pure]`, which was a pretty terrible name.

Now we have `#[mode(terminates)]` and `#[mode(ghost)]`... What do we think about those names? And what about adding more modes in the future (like `no_int_overflows`)?


jhjourdan:
I am in favor of these kind of simplifications, to avoid the accumulation of attributes. That being said, I don't much like `mode(ghost)` as a name, because it seems to imply that the function can only be used in ghost code.

Since this attribute is used to specify safety requirements, could we use `#[safety(ghost)]`, `#[safety(terminates)]` etc ? Intuitively, this means that the function is safe enough to be used in ghost/terminating contexts, but not that is can only be used in these contexts.

I was tempted to add a mode `bitwise`, but it is a bit different : instead of specifying a different set of VCs (i.e., a safety mode), it specifies how these VCs should be checked. So we should probably have another generic attribute for these kind of settings later.

While we are at it, I would like to make the `#[logic]` attribute more parameterized, to avoid the "attribute soup" that we have currently:
- add `open` to it: e.g., `#[logic(open)]`, `#[logic(open=crate))]`, etc... instead of `#[open]`?
- add `#[logic(open=not)]`, for opaque functions (today, we do that with `#[trusted]`, but I think `#[trusted]` should only be used to say that the VC should not be checked
- add `#[logic(law)]` instead of `#[law]`
- perhaps other things?

dianegolfouse:
> Since this attribute is used to specify safety requirements, could we use `#[safety(ghost)]`, `#[safety(terminates)]` etc ? Intuitively, this means that the function is safe enough to be used in ghost/terminating contexts, but not that is can only be used in these contexts.

Hmm, I like the idea, but `safety` makes me think of `unsafe` (especially now that there is a `safe` keyword)... It's better than `mode` though.

> I was tempted to add a mode `bitwise`, but it is a bit different : instead of specifying a different set of VCs (i.e., a safety mode), it specifies how these VCs should be checked. So we should probably have another generic attribute for these kind of settings later.

I would note that (for now?) `terminates` and `ghost` do not change VCs.

> While we are at it, I would like to make the `#[logic]` attribute more parameterized, to avoid the "attribute soup" that we have currently:
> 
>   * add `open` to it: e.g., `#[logic(open)]`, `#[logic(open=crate))]`, etc... instead of `#[open]`?
>   * add `#[logic(open=not)]`, for opaque functions (today, we do that with `#[trusted]`, but I think `#[trusted]` should only be used to say that the VC should not be checked
>   * add `#[logic(law)]` instead of `#[law]`
>   * perhaps other things?

I like `#[logic(law)]`, but I'm less convinced by `#[logic(open)]`... we kind of lose the link with `pub`'s syntax.

jhjourdan:
> > Since this attribute is used to specify safety requirements, could we use `#[safety(ghost)]`, `#[safety(terminates)]` etc ? Intuitively, this means that the function is safe enough to be used in ghost/terminating contexts, but not that is can only be used in these contexts.
> 
> Hmm, I like the idea, but `safety` makes me think of `unsafe` (especially now that there is a `safe` keyword)... It's better than `mode` though.

What about `#[check(ghost)]`, `#[check(terminates)]`, .... ?

> 
> > I was tempted to add a mode `bitwise`, but it is a bit different : instead of specifying a different set of VCs (i.e., a safety mode), it specifies how these VCs should be checked. So we should probably have another generic attribute for these kind of settings later.
> 
> I would note that (for now?) `terminates` and `ghost` do not change VCs.

Not the VCs sent to provers, but Creusot does verify more things with these attributes. So, it does verify more "verification conditions". Ok, well, the word is not well chosen here. But you get the idea.

> 
> > While we are at it, I would like to make the `#[logic]` attribute more parameterized, to avoid the "attribute soup" that we have currently:
> > 
> > * add `open` to it: e.g., `#[logic(open)]`, `#[logic(open=crate))]`, etc... instead of `#[open]`?
> > * add `#[logic(open=not)]`, for opaque functions (today, we do that with `#[trusted]`, but I think `#[trusted]` should only be used to say that the VC should not be checked
> > * add `#[logic(law)]` instead of `#[law]`
> > * perhaps other things?
> 
> I like `#[logic(law)]`, but I'm less convinced by `#[logic(open)]`... we kind of lose the link with `pub`'s syntax.

If you prefer, we could switch to  `#[logic(open)]`, `#[logic(open(crate)))]`, etc... Then I don't see how it is further than the syntax for `pub`.

dianegolfouse:
I settled on `check(...)`. It's a _bit_ vague, but I liked the alternatives less 🤷‍♀️

dianegolfouse:
Let's do the `logic(...)` things in another PR?

jhjourdan:
Sure. Let's merge this.

PR #1655, opened 2025-08-05
Merge syntax into `#[logic]`

arnaudgolfouse:
- [x] Change `#[law]` to `#[logic(law)]`
- [x] Change `#[open]` to `#[logic(open)]`
- [ ] Change some usages of `#[trusted]` to `#[logic(open = not)]`

(Original discussion in #1652)


jhjourdan:
> Change some usages of `#[trusted]` to `#[logic(open = not)]`

Rethinking about that, I wonder if `#[logic(opaque)]`/`#[opaque]` (for type definitions) wouldn't be clearer. What do you htink?

dianegolfouse:
> > Change some usages of `#[trusted]` to `#[logic(open = not)]`
> 
> Rethinking about that, I wonder if `#[logic(opaque)]`/`#[opaque]` (for type definitions) wouldn't be clearer. What do you htink?

I think this is slightly clearer what `logic(opaque)` means, yes. There would just need to be a small check to ensure that it is not used together with `open`.

jhjourdan:
As part of my last commit, I added quite a few things. Here is the commit message:

>    Add new tags to the `#[logic]` attribute:
>    
>    - `#[logic(inline)]` adds the rewrite_def meta and the inline:trivial attribute
>    - `#[logic(opaque)]` declares opaque definitions (`#[opaque]` can also be used for types). This is now a different notion from `#[trusted]` definitions, whose specifications are not proved
>    - `#[logic(open(xxx))]` replaces `#[open(xxx)]`
>    
>    I also cleaned up the attributes used to declare builtins and intrinsics. We no longer use diagnostic items, which are designed to serve an etirely different purpose. We no longer need `#[cfg(creusot, ...)] for these.

jhjourdan:
I added `compute_specified` as part of the tactics Why3find uses automatically.

PR #1659, opened 2025-08-08
Update toolchain to nightly-2025-09-04

Lysxia:


Lysxia:
`unsized_locals` was removed :(

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -798,35 +829,48 @@ fn fn_postcond_term<'tcx>(
...
-            ]))
+            Some(Term::call(
+                tcx,
+                typing_env,
+                get_fn_impl_postcond(tcx),
+                subst_postcond,
+                [
+                    Term {
+                        ty: closure_ty,
+                        kind: TermKind::Projection {
```
jhjourdan: This is ugly. We really need a smart constructor `Term::projection`.

creusot/src/backend/logic/vcgen.rs
```diff
@@ -705,6 +705,7 @@ impl<'tcx> VCGen<'_, 'tcx> {
                 }
                 ProjectionElem::OpaqueCast(ty) => def(ProjectionElem::OpaqueCast(ty)),
                 ProjectionElem::Subtype(ty) => def(ProjectionElem::Subtype(ty)),
+                ProjectionElem::UnwrapUnsafeBinder(_) => todo!(),
```
jhjourdan:
```suggestion
                ProjectionElem::UnwrapUnsafeBinder(ty) => def(ProjectionElem::UnwrapUnsafeBinder(ty)),
```

All the cases are handled here, even these that Creusot does not support.

creusot/src/backend/projections.rs
```diff
@@ -253,6 +256,7 @@ pub(crate) fn projections_to_expr<'tcx, 'a, N: Namer<'tcx>, V: Debug>(
             | ProjectionElem::Subtype(_) => {
                 ctx.dcx().span_bug(span, format!("Unsupported projection {proj:?}"))
             }
+            ProjectionElem::UnwrapUnsafeBinder(_) => todo!(),
```
jhjourdan: Move this above.

creusot/src/translation/function.rs
```diff
@@ -351,6 +351,7 @@ impl<'body, 'tcx> BodyTranslator<'body, 'tcx> {
                 mir::ProjectionElem::Downcast(s, ix) => mir::ProjectionElem::Downcast(s, ix),
                 mir::ProjectionElem::OpaqueCast(ty) => mir::ProjectionElem::OpaqueCast(ty),
                 mir::ProjectionElem::Subtype(ty) => mir::ProjectionElem::Subtype(ty),
+                mir::ProjectionElem::UnwrapUnsafeBinder(_) => todo!(),
```
jhjourdan:
```suggestion
                mir::ProjectionElem::UnwrapUnsafeBinder(ty) => mir::ProjectionElem::UnwrapUnsafeBinder(ty),
```

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -601,15 +609,21 @@ fn fn_once_postcond_term<'tcx>(
...
+            Some(Term::call(
+                tcx,
+                typing_env,
+                get_fn_impl_postcond(tcx),
+                subst_postcond,
+                [
+                    Term {
+                        ty: closure_ty,
+                        span: self_.span,
+                        kind: TermKind::Projection { lhs: Box::new(self_), idx: 0usize.into() },
```
jhjourdan: Smart constructor.

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -699,18 +715,24 @@ fn fn_mut_postcond_term<'tcx>(
...
-                            idx: 0usize.into(),
+                Term::call(
+                    tcx,
+                    typing_env,
+                    get_fn_impl_postcond(tcx),
+                    subst_postcond,
+                    [
+                        Term {
+                            ty: closure_ty,
+                            kind: TermKind::Projection {
```
jhjourdan: Smart constructor.

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -916,14 +956,23 @@ fn fn_once_precond_term<'tcx>(
...
-            ]))
+            Some(Term::call(
+                ctx.tcx,
+                typing_env,
+                get_fn_once_impl_precond(tcx),
+                subst_postcond,
+                [
+                    Term {
+                        ty: closure_ty,
+                        kind: TermKind::Projection {
```
jhjourdan: Smart constructor.

creusot-contracts/src/logic/fmap.rs
```diff
@@ -502,7 +502,7 @@ impl<K: Clone + Copy, V: Clone + Copy> Clone for FMap<K, V> {
 // Having `Copy` guarantees that the operation is pure, even if we decide to change the definition of `Clone`.
 impl<K: Clone + Copy, V: Clone + Copy> Copy for FMap<K, V> {}
 
-impl<K: ?Sized, V> Invariant for FMap<K, V> {
+impl<K, V> Invariant for FMap<K, V> {
```
jhjourdan: Why cannot you change the definition bellow instead?

creusot-contracts/src/resolve.rs
```diff
@@ -29,7 +29,7 @@ pub fn structural_resolve<T: ?Sized>(_: T) -> bool {
     dead
 }
 
-impl<T1, T2: ?Sized> Resolve for (T1, T2) {
+impl<T1, T2> Resolve for (T1, T2) {
```
jhjourdan: Why is this needed?

creusot-metadata/src/decoder.rs
```diff
@@ -94,14 +96,16 @@ impl SpanDecoder for MetadataDecoder<'_, '_> {
...
                 let symbol_index = self.read_u32();
-                Symbol::new_from_decoded(symbol_index)
+                Symbol::new(symbol_index)
             }
             _ => unreachable!(),
         }
     }
-
+    fn decode_byte_symbol(&mut self) -> rustc_span::ByteSymbol {
+        todo!()
```
jhjourdan: You could also copy-paste the `decode_symbol_or_byte_symbol` method.

jhjourdan:
> `unsized_locals` was removed :(

That's *really* a pitty. I don't see any workaround, apart from automatically wrapping every logic variable in a reference. But this is so painful...

creusot-contracts/src/resolve.rs
```diff
@@ -33,7 +33,7 @@ impl<T1, T2: ?Sized> Resolve for (T1, T2) {
     #[logic(prophetic)]
     #[open]
     fn resolve(self) -> bool {
-        resolve(self.0) && resolve(self.1)
+        resolve(self.0) && resolve(*&self.1)
```
Lysxia: ???

tests/should_fail/bug/695.coma
```diff
@@ -242,11 +244,7 @@ module M_695__valid [#"695.rs" 12 0 12 27]
   predicate postcondition_once (self: closure2) (args: bool) (result: ()) =
     [%#s695'9] let b = args in b /\ (self._0'0).final = (2: UInt32.t) \/ not b /\ (self._0'0).final = (1: UInt32.t)
   
-  predicate resolve'1 (self: MutBorrow.t UInt32.t) = [%#sresolve] self.final = self.current
-  
-  predicate resolve'2 (_0'1: MutBorrow.t UInt32.t) = resolve'1 _0'1
-  
-  predicate resolve'3 (_0'1: closure2) = resolve'2 _0'1._0'0
+  predicate resolve'1 (_0'1: closure2) = true
```
Lysxia: this is broken

tests/should_succeed/traits/18_trait_laws.rs
```diff
@@ -1,7 +1,8 @@
 extern crate creusot_contracts;
 use creusot_contracts::*;
 
-pub trait Symmetric {
+pub trait Symmetric: Sized {
+    // TODO
```
Lysxia: What can I do here? I don't understand the error I got before I added that bound.

tests/should_succeed/lang/unsized_quant.rs
```diff
@@ -1,14 +1,12 @@
-#![allow(incomplete_features)]
-#![feature(unsized_locals)]
-
+// TODO: keep this test?
```
Lysxia: Rust removed the feature `unsized_locals`. If we can't find a workaround then this test is kinda pointless.

Lysxia (commented):
Almost there.

- [x] fix some breakage about `resolve` for closures;
- [x] decide what to do about removed support for unsized locals, which we relied on to bind variables of unsized types in Pearlite quantifiers and mappings.

jhjourdan:
> fix some breakage about resolve for closures;

Could you tell a bit more about this?

> decide what to do about removed support for unsized locals, which we relied on to bind variables of unsized types in Pearlite quantifiers and mappings.

I have an idea for this, and, even thoughwe already tried many things to solve that issue, I don't think this idea has already been tried.

Here it is: each local variable used in pearlite whose type is T would have type &'a T instead (for a lifetime 'a to be introduced somewhere, if needed). Then, when compiling pearlite code, if we see an identifier, we can determine (with some simple syntactic scope analysis) if the identifier is bound in pearlite, in which case we dereference it, or if it is bound in program code, in which case we let it as is. Then, we need to change the type of identifiers at the place they are bound: this is easy if they are bound in logical closures/quantifiers. If they are bound by logical function parameters, we need to rewrite the prototype. E.g. we translate:

```rust
#[logic]
fn f((x, y): (i32, i32)) -> i32 { ... }
```

into:

```rust
#[logic]
fn f((ref x, ref y): (i32, i32)) -> i32 { ... }
```

What do you think?

Lysxia:
> Could you tell a bit more about this?

I don't have details yet, I just noticed that there's a `should_fail` test that succeeds and the main difference is that it has a `resolve` predicate for a closure that became `true` instead of resolving its captured variable, here:

https://github.com/creusot-rs/creusot/pull/1659#discussion_r2304051333

creusot-contracts/src/std/ops.rs
```diff
@@ -53,7 +53,7 @@ pub trait FnMutExt<Args: Tuple>: FnOnceExt<Args> {
 
     #[law]
     #[ensures(self.postcondition_once(args, res) ==
-              exists<res_state: Self> self.postcondition_mut(args, res_state, res) && resolve(res_state))]
+              exists<res_state: &Self> self.postcondition_mut(args, *res_state, res) && resolve(res_state))]
```
Lysxia: Oups
```suggestion
              exists<res_state: &Self> self.postcondition_mut(args, *res_state, res) && resolve(*res_state))]
```
This is the dumb cause of the breakage with closures I noticed yesterday.

jhjourdan: The kind of reason why I would really like having quantifiers of unsized types

Lysxia:
I'm going to try your idea to deal with unsized variables in Pearlite.

tests/should_succeed/traits/18_trait_laws.rs
Lysxia: The error message:
```
tests/should_succeed/traits/18_trait_laws.rs: failure
error[E0277]: the size for values of type `Self` cannot be known at compilation time
 --> 18_trait_laws.rs:9:5
  |
9 |     #[ensures(a.op(b) == b.op(a))]
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ doesn't have a size known at compile-time
  |
note: unsized values must be place expressions and cannot be put in temporaries
 --> 18_trait_laws.rs:9:5
  |
9 |     #[ensures(a.op(b) == b.op(a))]
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

The desugared term looks roughly like this:
```
        ::creusot_contracts::__stubs::equal((*&a).op(*&a), (*&a).op(*&a))
```

I think it's complaining about where to store the results of the `op` calls.

creusot-contracts/src/resolve.rs
Lysxia: There was a bug in the proc macro where field projections were wrapped in `{ ... }`, which must be `Sized`.

Desugaring of `resolve(self.1)` before:
```
resolve({ (*&self).1 })
```

I've fixed it:
```
resolve((*&self).1)
```

creusot/src/translation/external.rs
```diff
@@ -104,10 +104,12 @@ pub(crate) fn extract_extern_specs_from_item<'tcx>(
...
     let mut errors = Vec::new();
     for i in 0..outer_subst.len() {
-        match (inner_subst[i + extra_parameters].unpack(), outer_subst[i].unpack()) {
+        match (inner_subst[i + extra_parameters].kind(), outer_subst[i].kind()) {
             (GenericArgKind::Type(t1), GenericArgKind::Type(t2)) => match (t1.kind(), t2.kind()) {
                 (TyKind::Param(param1), TyKind::Param(param2))
-                    if param1.name == param2.name || param1.name.as_str().starts_with("Self") =>
+                    if true // TODO
+                        || param1.name == param2.name
+                        || param1.name.as_str().starts_with("Self") =>
```
Lysxia: Last TODO

Lysxia:
- [x] fix test `18_trait_laws.rs`
- [x] fix check of `Param` in extern specs

creusot/src/translation/external.rs
jhjourdan: What's the issue here?

(Apart from the fact that `start_with("Self")` is a very bad way of detecting `Self_`.)

tests/should_succeed/traits/18_trait_laws.rs
jhjourdan: The solution is probably to make every pearlite function return a reference, which we could dereference after each function call. I don't know how disruptive this is.

Otherwise, I think it's fair to add a `: Sized` bound here....

But, out of curiosity, why do we have these `*&` everywhere?

jhjourdan: > But, out of curiosity, why do we have these *& everywhere?

Nevermind, I remember that now.

creusot/src/translation/external.rs
Lysxia: The type of the fn argument of `is_some_and` changed from `impl FnOnce(T) -> bool` to `impl [const] FnOnce(T) -> bool + [const] Destruct` (and this type is used as the name of its own Param). So it seems that the right fix is to change the `extern_spec` of `is_some_and` to have a matching type. I haven't found the right feature yet (I think it's `const_trait_impl` but `impl [const] FnOnce(T) -> bool + [const] Destruct` still refuses to parse.

Lysxia: It's not getting parsed by `syn` which doesn't know about this new feature of course.

jhjourdan: Do we use the latest version of syn?

Lysxia: Indeed we just needed to update to the latest version from two weeks ago.

Lysxia: Adding `[const]` to arguments implies making the function itself `const`, which conflicts with our encoding of contracts.

```
error[E0493]: destructor of `std::option::Option<T>` cannot be evaluated at compile-time
  --> creusot-contracts/src/std/option.rs:67:17
```

So maybe it's time to not compare parameters by name?

jhjourdan: The sad thing is that, if I remember well, we rely on these names being equal for substitutions... So if the names do not match, extern specs will start to be randomly bogus..

jhjourdan: (I'm not so sure of this, though.)

Lysxia: I figured out a more robust implementation here. Basically, if we have an extern spec like

```
fn extern_spec_f<Self_, A, B>() {
  <Self_ as Tr<A>>::f::<B>()
}
```

we just check that the generic arguments in the `f` call are a permutation of the generic parameters of `extern_spec_f`, then we invert that substitution to instantiate the extern spec contract. Lifetimes are ignored. That way we no longer rely on using the exact same names as the original function, and this also seamlessly handles the special positioning of `Self` in the generic argument list of trait methods.

jhjourdan: So you call `f` in the body of its extern spec? That looks like an option. Perhaps @xldenis could give his opinion here, because I never touched this part of Creusot, and I don't know what are the tradeoffs.

xldenis: ugh yeah the const stuff was annoying at one point in the past already. We don't rely on the names of the arguments for correctness (normally), I just added that as an additional layer of sanity checking (it is _weird_ that you would change the names). 

The implementation @Lysxia is what external specs already should be desugaring to, unless i've forgotten something major.

Lysxia:
Also uncommented the `PredCell::update` method from https://github.com/creusot-rs/creusot/pull/1673

PR #1660, opened 2025-08-11
Avoid wrapping return handler for open program bodies

Lysxia:


jhjourdan:
I think this PR introduced a regression that need to be investigated: all the handlers of axiomatized functions add a trivial VC (the goal is `true`).

All the `proof.json` files are easily replayed, because Why3find is smart enough to skip these goals. However, when re-creating the `proof.json` files, they appear, and they appear when launching Why3ide.

Lysxia:
I don't see it, or I don't understand how to reproduce this regression.

PR #1664, opened 2025-08-12
Inline promoted bodies

Lysxia:
Refactor `to_why` to separate constructing the body from building the signature. This is part of my ongoing work on translating `const`. Inlining promoteds is not technically necessary as part of that but it makes the generated code more straightforward and similar to the source.

All the tests are affected because the lowerings of the signature and of the body of each function are done in a different order. The original order is tricky to preserve because signatures were lowered in the middle of constructing the body.


Lysxia:
(Now fixed) ~~Actually, to reduce the diff, I can try translating the signature after translating the body, because translating the body only needs the names from the signature.~~

PR #1675, opened 2025-08-29
Rework handling of unsized variables in Pearlite

Lysxia:
To not rely on deprecated feature `unsized_locals`.

Implement https://github.com/creusot-rs/creusot/pull/1659#issuecomment-3228473959 with one key difference: we can't rewrite arbitrary patterns to wrap variables under `ref` because proc macros can't distinguish variables from unqualified constructors (you want to avoid "`ref None`"). This affects `Mapping` arguments, `match` patterns, and logic function arguments. I compromise by only rewriting `Mapping`s whose argument pattern is a single identifier (`|x|` or `|x: ty|`), which is the vastly most common use case. `Mapping` with non-identifier patterns are not modified. Nor are `match` patterns and logic function arguments (but note that `#[logic] fn f(x: unsizedTy)` already works without doing anything; it's non-variable patterns that are trickier).

### DIFF
--- a/creusot-contracts-proc/Cargo.toml
+++ b/creusot-contracts-proc/Cargo.toml
@@ -23,7 +23,7 @@ creusot = ["dep:uuid", "dep:pearlite-syn", "proc-macro2/span-locations"]
 quote = "1.0"
 uuid = { version = "1.12", features = ["v4"], optional = true }
 pearlite-syn = { version = "0.6.0-dev", path = "../pearlite-syn", features = ["full"], optional = true }
-syn = { version = "2.0", features = ["full", "visit-mut"] }
+syn = { version = "2.0", features = ["full", "visit", "visit-mut"] }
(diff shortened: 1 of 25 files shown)


jhjourdan:
Good!

Would it make sense to do the same with let bindings that ony bind a variable ?

jhjourdan:
Perhaps we could have some special syntax to tell the macro that an ident is in fact a binder and that triggers this mechanism?

jhjourdan:
My problem is that if you have a function that takes `Option<unsized>`, you can't even pattern match on it...

Lysxia:
You can still manually match on `Some(ref x)`. The special syntax would just automate inserting `*` at use sites of `x`.

Lysxia:
Unless we have an immediate need for it I propose to postpone adding new Pearlite syntax for this.

PR #1680, opened 2025-09-02
Fixes to the handling of const params

jhjourdan:
@Lysxia : as discussed, I tried to use const params in e.g., the `sparse_array` example, and I discovered several bugs. With these fixes, the situation is better, but it appears that const params are never substituted, even when their value is known.

Could you please fix this? And check that associated constants are substituted correctly (in pearlite functions, in specifications)?


Lysxia:
Ready for review.

>  And check that associated constants are substituted correctly (in pearlite functions, in specifications)?

There are already tests for instantiating associated consts in `lang/const.rs`, and I've added one for what was missing here.

creusot/src/translation/constant.rs
```diff
@@ -124,36 +125,29 @@ fn try_scalar_to_literal<'tcx>(
     })
 }
 
-// This is also used in the translation of `RValue::Repeat`.
-pub(crate) fn tyconst_to_term<'tcx>(
+/// Use this when we are about to output Why3 (`lower_term` in `term.rs` and `build_wp` in `vcgen.rs`).
+/// Converts `Const::Param` to `TermKind::Item` which is ignored by `instantiate` so don't use this for terms that will be subject to substitutions
+/// (in that case you probably want `Term::const_()`).
+pub(crate) fn tyconst_to_term_final<'tcx>(
```
jhjourdan: This should be moved in the backend, I think.

creusot/src/translation/constant.rs
```diff
@@ -124,36 +125,29 @@ fn try_scalar_to_literal<'tcx>(
...
     match c.kind() {
         Value(ty, value) => valtree_to_term(value, ctx, ty, env, span),
+        Unevaluated(ty::UnevaluatedConst { def, args }) => Some(Term::item(def, args, ty)),
         Param(p) => {
-            let def_id = ctx.generics_of(caller_id).const_param(p, ctx.tcx).def_id;
-            let ty = ctx.type_of(def_id).instantiate_identity();
-            Some(Term::const_param(ctx.tcx, def_id, ty, span))
+            let tcx = ctx.tcx;
+            let def_id = tcx.generics_of(caller_id).const_param(p, tcx).def_id;
+            Some(Term::item(def_id, erased_identity_for_item(tcx, def_id), ty))
```
jhjourdan: `erased_identity_for_item(tcx, def_id)` is necessarily empty because this is a const param, right? Then I think it should be clerer to use the empty substitution here.

jhjourdan (commented):
Generally the changes looks good, and I would agree merging if the test passed.

Still, I am curious : by what magic the const param gets substituted?

creusot/src/translation/pearlite.rs
```diff
@@ -162,6 +161,8 @@ pub enum TermKind<'tcx> {
         arg: Box<Term<'tcx>>,
     },
     Item(DefId, GenericArgsRef<'tcx>),
+    /// Constants. May get substituted using `instantiate` (via `TypeFoldable`).
+    Const(Const<'tcx>),
```
jhjourdan, line 165: Ah, I see. That's the magic.

tests/should_succeed/sparse_array/proof.json
```diff
@@ -7,15 +7,34 @@
...
+    "M_sparse_array__qyi11026919801270832767__bounded_fset_len": {
+      "vc_bounded_fset_len": { "prover": "cvc5@1.0.5", "time": 0.06 }
+    },
+    "M_sparse_array__qyi11026919801270832767__get": {
+      "vc_get": { "prover": "cvc5@1.0.5", "time": 0.051 }
+    },
+    "M_sparse_array__qyi11026919801270832767__lemma_permutation": {
+      "vc_lemma_permutation": { "prover": "cvc5@1.0.5", "time": 0.064 }
+    },
+    "M_sparse_array__qyi11026919801270832767__lemma_permutation_aux": {
```
jhjourdan, line 30: Why is this duplicated in the proof.json file?

jhjourdan:
The only remaining thing to fix here is the corrct unfolding of contants cross-crates, including correct placement of resolve statements.

jhjourdan:
Could you either fix this here or open an issue? This is an unsoundness issue, so I would prefer handling this soon if possible.

tests/should_succeed/sparse_array/proof.json
Lysxia: Changing the `Sparse` type changes the hash of its `impl`s, and why3find doesn't remove old goals (this is fixed in its latest version, I should update why3find soon.)

Lysxia: I removed the proof and re-generated it to only keep fresh goals.

Lysxia:
> Could you either fix this here or open an issue? This is an unsoundness issue, so I would prefer handling this soon if possible.

OK, I opened #1681 because that is going to take a little bit longer and the fixes here are more urgent.

Issue #1681, opened 2025-09-03
Cross-crate translation of consts does not resolve mutable borrows


Lysxia:
If crate B uses a `const` from crate A, then borrows in the body of the `const` will not be resolved. This breaks soundness because this also omits type invariant assertions on the final value.

I did some refactoring in preparation for handling those resolves correctly but then I was lazy and did not implement it because a `const` that involves mutable borrows sounds so far fetched.

Lysxia:
#1690 fixes the unsoundness by failing when we are missing borrow data for resolves.

TODO: store cross-crate borrow data to actually support consts that use mutable borrows.

PR #1690, opened 2025-09-04
Fail in case of missing resolve data for mutable borrows

Lysxia:
Fix the unsoundness of #1681


Lysxia:
Currently failing because of promoted `&mut []`:

```
fn as_mut_slice(s: &mut Option<usize>) -> &mut [usize] {
    match s {
        None => &mut [],
        Some(t) => std::slice::from_mut(t),
    }
}
```

`&mut []` is apparently the only case where const-promotion involves mutable borrows, so the solution is to allow it as a special case: no resolve is necessary because the value is uniquely determined (the length of the prophecized value is set to zero by the type invariant).

jhjourdan:
Does this fix the unsoundness by emiting an error instead, or does it actually solve the issue by serializing the borrow data in the cmeta file?

Lysxia:
It fixes the unsoundness by emitting an error :)

jhjourdan:
Ok, then the issue should be left open.

PR #1691, opened 2025-09-04
Better specs for Union Find.

jhjourdan:
This also makes the ^ notation overloadable, which is convenient for Ghost and Snapshot values.


xldenis:
Oh I really like the overloaded `^` i don't know why we never did that before.

jhjourdan:
The sad thing about this overloading is that it makes * and ^ asymetric, because if s is a snapshot, then the current value is denoted by `**x`, while the final value is denoted by `^x`.

But anyway, it was already somwhat asymetric, because autoderef only works for current values of course.

PR #1696, opened 2025-09-05
Renaming: we differentiate *tokens* used to open invariants, and *namespaces* used to identify them

jhjourdan:


jhjourdan:
@arnaudgolfouse : agreed?

dianegolfouse:
Yep, that's better

PR #1704, opened 2025-09-05
Fix error when translating `dyn`

### LINKED ISSUE #1337: Debug implementation causes Creusot ICE
```
thread 'rustc' panicked at creusot/src/backend/ty_inv.rs:69:18:
not implemented: dyn [Binder { value: Trait(std::fmt::Write), bound_vars: [] }] + '?24

error: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/creusot-rs/creusot/issues/new

note: please make sure that you have updated to the latest nightly

note: rustc 1.82.0-nightly (c1a6199e9 2024-07-24) running on aarch64-apple-darwin

note: compiler flags: -Z unstable-options --crate-type bin -C embed-bitcode=no -C debuginfo=2 -C split-debuginfo=unpacked

note: some of the compiler flags provided by cargo are hidden

query stack during panic:
```

Lysxia:
Fix #1337

Fix a check to allow `dyn` to be translated to an opaque type

This allows translating `derive(Debug)`: `Debug::fmt` uses the type `std::fmt::Formatter` which contains a field of type `&dyn Write`.

### DIFF
--- a/creusot-contracts/src/std/fmt.rs
+++ b/creusot-contracts/src/std/fmt.rs
@@ -1,17 +1,132 @@
 use crate::*;
 #[cfg(creusot)]
-use ::std::fmt::{Arguments, Debug, Formatter};
+use ::std::fmt::{Arguments, Debug, Error, Formatter};
 
 extern_spec! {
     mod core {
         mod fmt {
             impl<'a> Formatter<'a> {
                 #[requires(true)]
+                fn write_str(&mut self, data: &str) -> Result<(), Error>;
+
+                #[requires(true)]
                 fn debug_struct_field1_finish<'b>(
                     &'b mut self,
                     name: &str,
                     name1: &str,
                     value1: &dyn Debug,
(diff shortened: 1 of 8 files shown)


jhjourdan:
Is this sound?

Lysxia:
There are no operations on `Opaque.dyn`, can something be broken even in spite of that?

jhjourdan:
`dyn` types implement the corresponding traits, and allow calling their associated members. If you have a member function that says as a postcondition "there is only one element in Self", then this becomes wrong once in `dyn`.

I haven't written anything concrete, and I am not sure this is indeed a paradox, but this seems very fishy. At the very least, this deserves thinking about this possible paradox, and others....

Lysxia:
OK it is a good counterexample. This is more subtle than I thought.

One solution for now is to start by forbidding the construction of trait objects. That should plug the soundness hole. From there we can figure out conditions where we can allow it.

PR #1705, opened 2025-09-06
Improve translation of dyn

Lysxia:
Plug some soundness holes

- Traits may be unimplementable, in which case `dyn` should be forbidden (see `tests/should_fail/unsound_dyn.rs`). Currently we have a hardcoded whitelist: we allow `dyn std::fmt::Debug` and `dyn std::fmt::Write` in order to not break `derive(Debug)`. (Note: there are also cases where a trait is implementable but `dyn` would still break soundness. For example, a trait with a law that `Self` is a singleton.)
- Each `dyn` type should be translated to a different Why3 type. Previously `Opaque.dyn` could implement contradictory traits.

### DIFF
--- a/creusot/src/backend/clone_map/elaborator.rs
+++ b/creusot/src/backend/clone_map/elaborator.rs
@@ -52,7 +52,7 @@ use why3::{
 
 /// Weak dependencies are allowed to form cycles in the graph, but strong ones cannot,
 /// weak dependencies are used to perform an initial stratification of the dependency graph.
-#[derive(PartialEq, PartialOrd, Copy, Clone)]
+#[derive(PartialEq, PartialOrd, Copy, Clone, Debug)]
 pub enum Strength {
     Weak,
     Strong,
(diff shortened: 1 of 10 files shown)


creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -436,10 +436,59 @@ fn expand_type<'tcx>(
...
+        ty::ExistentialPredicate::Trait(tr) => is_logically_dyn_compatible_trait(tcx, tr.def_id),
+        _ => false,
+    })
+}
+
+/// Base trait for which `dyn` is supported by Creusot
+fn is_logically_dyn_compatible_trait<'tcx>(tcx: ty::TyCtxt<'tcx>, tr: DefId) -> bool {
+    // TODO: support more traits
+    let name = tcx.def_path_str(tr);
+    ["std::fmt::Debug", "std::fmt::Write"].contains(&name.as_ref())
```
Lysxia, line 470: flex-tape-meme

PR #1706, opened 2025-09-08
Simplify formalization of Union-Find, and a new even simpler version …

jhjourdan:
This also include a new API in creusot-contracts, allowing to extract data from a snapshot to ghost code.


dianegolfouse:
Are there significant differences between `union_find.rs` and `union_find_cpp.rs`?

jhjourdan:
Yes. The simpler version does not contain the invariants related to depths, and does not contain the invariant that guarantee the unicity of addresses of elements. As a result, in this simpler version, we cannot prove termination and we cannot compare elements by address, but the proof is easier to explain.

It's not clear that we want to keep both versions in the repo. We want to keep the more involved version because, well, it is more involved. But I don't know where to put the simpler version, which we need to keep somewhere for the paper submission. Perhaps Creusot-example is a good place for this?

dianegolfouse:
> Perhaps Creusot-example is a good place for this?

If this is ready to go live (it's still private at the moment) then why not!

jhjourdan:
Oh, I did not know that this was still private! Let's postpone that after the deadline, then.

PR #1707, opened 2025-09-09
`Plain` for `logic::Id`

dianegolfouse:


tests/should_succeed/persistent_array.rs
```diff
@@ -176,9 +176,9 @@ pub mod implementation {
             let new_frac = {
                 let program_value = &program_value;
                 self.map_invariant.borrow().open(tokens, |mut tokens| {
```
jhjourdan: It's independent from this, PR, but please use a different name here. `tokens` means two different things, that's very confusing.

tests/should_succeed/persistent_array.rs
```diff
@@ -253,8 +253,8 @@ pub mod implementation {
             i: usize,
             tokens: Ghost<&'a CompleteMap<T>>,
         ) -> Option<&'a T> {
-            let id = snapshot!(inner.view().id());
-            let perm = ghost!(tokens.own_map.get_ghost(&id).unwrap());
+            let id = snapshot!(inner@.id());
+            let perm = ghost!(tokens.own_map.get_ghost(&*id.into_ghost()).unwrap());
```
jhjourdan:
```suggestion
            let perm = ghost!(tokens.own_map.get_ghost(&*inner@.id_ghost()).unwrap());
```

tests/should_succeed/persistent_array.rs
dianegolfouse: Ohhh, right

tests/should_succeed/persistent_array.rs
dianegolfouse: Thanks!

PR #1711, opened 2025-09-09
Add creusot::meta attribute

Lysxia:


tests/should_succeed/specification/logic_functions.coma
```diff
@@ -31,3 +31,21 @@ module M_logic_functions__use_logic_pearlite [#"logic_functions.rs" 19 0 19 27]
...
 end
+module M_logic_functions__use_quatorze [#"logic_functions.rs" 54 0 54 21]
+  let%span slogic_functions = "logic_functions.rs" 53 10 53 26
+  let%span slogic_functions'0 = "logic_functions.rs" 50 4 50 6
+
+  use creusot.prelude.Any
+
+  constant quatorze: int = [%#slogic_functions'0] 14
+
+  meta rewrite_def function quatorze
```
jhjourdan: I think the name of the meta should always be a string.

In any case, there is no reason to emit something different for `rewrite_def` and for `compute_max_steps`.

jhjourdan:
I added a commit that uses the new attribute. This looks good, thanks.

PR #1714, opened 2025-09-10
Reorganize creusot-contracts modules

Lysxia:
The `creusot_contracts` API is getting a bit crowded. That makes the documentation daunting to newcomers. I've kept the features that are absolutely essential at the top, and everything else is moved into an inner module.

- Create a `structures` module for more "advanced" data structures: `peano`, `pcell`, `predcell`, `ptr_own`, `resource`
  `structures` reexports all of the types so there shouldn't ever be a reason to `use` from these sub-modules anymore.
- `fn_ghost` -> `ghost::fn_ghost`
- `well_founded` -> `logic::well_founded`
- `util` -> merged into `logic`
- `num_rational` -> `logic::real`

I'm not quite familiar with `local_invariant` yet, but I've kept it at the top, as a "core feature", because it requires special handling by Creusot.


jhjourdan:
Good idea! I agree that the name space started to get a bit messy.

>     * `fn_ghost` -> `ghost::fn_ghost`
> 
>     * `well_founded` -> `logic::well_founded`
> 
>     * `util` -> merged into `logic`
> 
>     * `num_rational` -> `logic::real`

Agreed.

>     * Create a `structures` module for more "advanced" data structures: `peano`, `pcell`, `predcell`, `ptr_own`, `resource`
>       `structures` reexports all of the types so there shouldn't ever be a reason to `use` from these sub-modules anymore.
>

None of them are data strucutres. So "structure" is a rather bad name. And in general, I don't really see any coherence of all these.

Also, I don't like the idea of both exposing the modules and reexporting their contents, because this creates aliases, which are confusing. Either we reexport the whole modules and make the submodule private, or we use a submodule only.

`ptr_own`, `resource` and `local_invariant` are all about ghost code. So they could be moved to a submodule of `ghost`.

`pcell` and `predcell` are about interior mutability, so we could move them to a `cell` module.

`peano` really is a different from others, so maybe we should leave it in the root.

> 
> 
> I'm not quite familiar with `local_invariant` yet, but I've kept it at the top, as a "core feature", because it requires special handling by Creusot.

Lysxia:
OK, I've implemented your suggestions. The hierarchy looks like this, with no redundant modules:
```
ghost::PtrOwn
ghost::resource::{Resource, fmap_view::{Authority, Fragment}}
ghost::local_invariant::*
cell::{PCell, PredCell}
peano::PeanoInt
```

creusot-contracts/src/ghost/resource.rs
```diff
@@ -1,5 +1,16 @@
...
+//!
+//! This module also defines a specialization of [`Resource`] for the case where you want:
+//! - A single, authoritative version of a [`FMap`][crate::logic::FMap]
+//! - Multiple fragments of such a map, that each assert that some key-value is in the
+//!   map.
+//!
+//! These are the [`Authority`] and [`Fragment`] types respectively.
+mod fmap_view;
+
+pub use fmap_view::{Authority, Fragment};
```
jhjourdan: I'm not so sure for this change. `Authority` and `Fragment` really come together, and comes at a different level from `Ressource`.

creusot-contracts/src/lib.rs
```diff
@@ -22,6 +22,25 @@
...
+//!
+//! 2. [`logic`][mod@logic]: Logical structures used in specifications
+//!
+//! 3. [`ghost`][mod@ghost]: Ghost code
+//!
+//! 4. [`std`][mod@std]: Specifications for the `std` crate
+//!
+//! 5. [`cell`][mod@cell]: Interior mutability
+//!
+//! 6. [`prelude`][mod@prelude]: What you get from `use creusot_contracts::*;`
```
jhjourdan, line 43: While we are at it, I wonder whether `prelude` and `base_prelude` are still relevant or should be removed. The goal was to find some way to import it by default at the header of every crate loaded by Creusot, but I don't think we will be able to do that.

creusot-contracts/src/lib.rs
```diff
@@ -439,9 +446,12 @@ mod base_prelude {
...
 
     #[cfg(creusot)]
-    pub use crate::std::mem::size_of_logic;
-    #[cfg(creusot)]
-    pub use crate::util::such_that;
+    pub use crate::{
+        invariant::inv,
+        logic::such_that,
+        resolve::{resolve, structural_resolve},
+        std::mem::size_of_logic,
```
jhjourdan: I wonder if we could make this list shorter. There are certainly things here that we don't use so often.

I'm thinking of (the list is open to discussion):
- `such_that`
- `size_of_logic`
- `structural_resolve`
- `FnGhost`
- `InhabitedInvariant`
- `Subset`
- `OrdLogic`
- `Seq`
- `WellFounded`
- `Clone` (we no longer have our own)
- `Default` (we no longer have our own)
- `PartialEq` (we no longer have our own)
- `PartialOrd` (we no longer have our own)

jhjourdan: Of course, some of these are traits used in the `pearlite!` macro, so that it should still be exported, but with `as _`.

creusot-contracts/src/lib.rs
jhjourdan: I mean, they should maybe be inlined in the root. No need for a dedicated module.

Lysxia: Having all the reexports in a `prelude` module makes the generated doc a bit tidier.

The doc page currently just shows "Re-exported: `pub use prelude::*`". If we inlined `prelude`, we would suffer all of the traits re-exported `as _`.

jhjourdan: That's just a matter of adding a no-doc attribute, isn't it?

creusot-contracts/src/lib.rs
Lysxia: > - Clone (we no longer have our own)
> - Default (we no longer have our own)
> - PartialEq (we no longer have our own)

I think these re-exports are our derive macros and this is one fishy part of Rust namespacing.

Lysxia: I think it's worth keeping `OrdLogic` and `Seq` to mirror `PartialOrd` and `Vec` being in `::std::prelude`.

creusot-contracts/src/lib.rs
Lysxia: I'd still like the list of re-exports to appear somewhere.

creusot-contracts/src/ghost/resource.rs
Lysxia: OK, reverted

creusot-contracts/src/lib.rs
jhjourdan: Alright.

PR #1727, opened 2025-09-22
Allow `result` argument name for extern functions

Lysxia:
Motivation: compiling with `--test` crashes on [`assert_test_result<T: Termination>(result: T) -> Result<(), String>`](https://doc.rust-lang.org/test/fn.assert_test_result.html)

Also update an error message about unsupported casts


jhjourdan:
I would like `result` to be accessed with a special syntax which would not conflict with identifiers, like `res!` or so.

PR #1729, opened 2025-09-23
termination of `union_find`

dianegolfouse:


dianegolfouse:
At some point we will want some mechanism to say 'this function terminates if some other function terminates'

jhjourdan:
Yeah, I actually think this is not so difficult : we just need to make the `#[check(..)]` flags first class so that we can use them in specs.

creusot-contracts/src/ghost/ptr_own.rs
```diff
@@ -58,6 +59,7 @@ impl<T> PtrOwn<T> {
 impl<T: ?Sized> PtrOwn<T> {
     /// Creates a ghost `PtrOwn` and associated `*const` from an existing [`Box`].
     #[trusted]
+    #[check(ghost)]
```
jhjourdan, line 62: Unsound!

If you allow this in ghost code, then you essentially allow creating an unbounded number of pointers, which all have distinct addresses in `usize`!

PR #1731, opened 2025-09-23
Access ptr metadata in logic

Lysxia:
I'm not sure this is the right way to do this.

Motivation: I want `PtrOwn<[T]>` to have the invariant that the length in the pointer metadata matches the length of the owned value. `self.val()@.len() == metadata_logic(self.ptr())@`. But `PtrOwn<T>` already has an `Invariant`.

What's annoying with this solution is the extra constraint `... + LogicMetadata`  on the `impl Invariant for PtrOwn<T>` . Maybe `metadata_logic` should be a function with trivial bounds, like [`metadata`](https://doc.rust-lang.org/std/ptr/fn.metadata.html).


creusot-contracts/src/ghost/ptr_own.rs
```diff
@@ -26,7 +28,9 @@ pub struct PtrOwn<T: ?Sized> {
 
 impl<T: ?Sized> PtrOwn<T> {
     /// The raw pointer whose ownership is tracked by this `PtrOwn`
+    #[trusted]
     #[logic]
+    #[ensures(metadata_matches(*self.val(), metadata_logic(result)))]
```
Lysxia: Would it make sense to move the current body of `invariant` here too?

Lysxia: Conclusion: we put everything into the `Invariant`. There is no strong reason to do things one way (invariant) or ther other (postcondition of `ptr()`).

PR #1736, opened 2025-09-26
Ghost iterators

dianegolfouse:
Add iterators on `Seq<T>`, `&Seq<T>`, `FMap<K, V>` and `&FMap<K, V>`.

I did not do mutable iterators for now.


jhjourdan:
It's rather sad that we will not be able to use iterator combinators in ghost code, because of `check(ghost)` and the fact that I think it is rather optimistic to require that all the iterators are ghost safe.

dianegolfouse:
I was under the impression that making `ghost` a (somewhat) first-class property would allow this: for example something like 
```rust
impl<I: Iterator> Iterator for Enumerate<I> {
    type Item = I::Item;
    #[check(ghost if I::next is ghost)]
    fn next(&mut self) -> Option<Self::Item> { /* */ }
}
```
Although of course it does not exists today.

jhjourdan:
That's right. This is essentially the same problem we have with termination.

And I don't see why we woul not make that a plain somewhat first-class value. Then, `#[check]` is just syntactic sugar for `#[requires()]` (although we may implement it differently in trivial cases.

PR #1740, opened 2025-09-29
Open the body of `metadata_matches`

Lysxia:


creusot/src/contracts_items/intrinsics.rs
```diff
@@ -93,6 +93,7 @@ impl Intrinsic {
                 | Intrinsic::PostconditionOnce
                 | Intrinsic::HistInv
                 | Intrinsic::Precondition
+                | Intrinsic::MetadataMatches
```
jhjourdan, line 96: Is this dead code?

jhjourdan: My bad, I as confused.

Lysxia: it is `dead`

PR #1744, opened 2025-09-30
Make addr_logic a builtin

### LINKED ISSUE #1674: `addr_logic` is not preserved by casts
When casting pointers, one cannot derive equality of the logical addresses:

```rust
let p1: *const T = ...;
let p2 = p1 as *const U;
proof_assert!(p1.addr_logic() == p2.addr_logic()); // unprovable!
```
That's because for each type, a different `addr_logic` function is generated: those functions are uninterpreted, and independent from each other.

The fix could be to have a hack to generate a single `addr_logic` function is this case: other functions on pointers should then behave correctly.

Lysxia:
Close #1674


creusot-contracts/src/std/ptr.rs
```diff
@@ -1,6 +1,16 @@
 use crate::*;
 pub use ::std::ptr::*;
 
+/// Integer representation of a pointer in logic.
+///
+/// This returns `u64` because there is no support for builtins with architecture-dependent types.
+/// In any case it is enough to support both 32-bit and 64-bit architectures.
```
jhjourdan: An easy way to use `usize` instead of `u64` here is to define in the prelude three different versions of `creusot.prelude.Opaque.addr_logic`, for the three different bit widths, and then use `cfg` to chose the right one.

See, for example what we are doing for `[T]::view` in `slice.rs`.

jhjourdan:
```rust
impl<T> View for [T] {
    type ViewTy = Seq<T>;

    #[logic]
    #[cfg_attr(target_pointer_width = "16", builtin("creusot.slice.Slice16.view"))]
    #[cfg_attr(target_pointer_width = "32", builtin("creusot.slice.Slice32.view"))]
    #[cfg_attr(target_pointer_width = "64", builtin("creusot.slice.Slice64.view"))]
    fn view(self) -> Self::ViewTy {
        dead
    }
}
```

Lysxia: Thanks!

creusot-contracts/src/std/ptr.rs
```diff
@@ -87,9 +96,9 @@ pub trait PointerExt<T: ?Sized>: Sized {
 }
 
 impl<T: ?Sized> PointerExt<T> for *const T {
-    #[logic(opaque)]
+    #[logic(open)]
```
jhjourdan:
```suggestion
    #[logic]
    #[cfg_attr(target_pointer_width = "16", builtin("creusot.prelude.Opaque.addr_logic_u16"))]
    #[cfg_attr(target_pointer_width = "32", builtin("creusot.prelude.Opaque.addr_logic_u32"))]
    #[cfg_attr(target_pointer_width = "64", builtin("creusot.prelude.Opaque.addr_logic_u64"))]
```

(You no longer need the `addr_logic` proxy.)

tests/should_fail/wrong_permissions/why3session.xml
```diff
@@ -24,10 +24,10 @@
  <goal name="vc_wrong_permcell_permission">
  <transf name="split_vc" >
   <goal name="vc_wrong_permcell_permission.0" expl="borrow requires">
-  <proof prover="0"><result status="timeout" time="1.000000" steps="12020186"/></proof>
+  <proof prover="0" timelimit="5"><result status="timeout" time="5.000000" steps="89195195"/></proof>
   <proof prover="1"><result status="unknown" time="0.020698" steps="1585"/></proof>
   <proof prover="2"><result status="unknown" time="0.052441" steps="6775"/></proof>
-  <proof prover="3"><result status="unknown" time="0.010895" steps="14"/></proof>
+  <proof prover="3" timelimit="5"><result status="unknown" time="0.010895" steps="14"/></proof>
```
jhjourdan: In the failing tests, I'm trying to keep using a timeout of 1 second instead of 5, because this makes the CI faster.

tests/should_fail/wrong_permissions/why3session.xml
```diff
@@ -39,10 +39,10 @@
...
   </goal>
   <goal name="vc_unknown_ptr_own_permission.1" expl="as_ref requires">
-  <proof prover="0"><result status="timeout" time="1.000000" steps="34223284"/></proof>
-  <proof prover="1"><result status="unknown" time="0.032468" steps="3235"/></proof>
-  <proof prover="2"><result status="unknown" time="0.044182" steps="8095"/></proof>
-  <proof prover="3"><result status="unknown" time="0.014728" steps="36"/></proof>
+  <proof prover="0" timelimit="5"><result status="timeout" time="5.000000" steps="107276244"/></proof>
+  <proof prover="1" timelimit="5"><result status="unknown" time="0.032468" steps="5831"/></proof>
+  <proof prover="2" timelimit="5"><result status="unknown" time="0.044182" steps="9804"/></proof>
+  <proof prover="3" timelimit="5"><result status="timeout" time="5.000000" steps="35278"/></proof>
```
jhjourdan: Same here.

PR #1746, opened 2025-09-30
Add specs for `align_of`

### LINKED ISSUE #1750: Verify `is_aligned_to`
https://github.com/creusot-rs/creusot/blob/3040a90a40e31b6dfe9cb731ed9171899f18e7e7/creusot-contracts/src/std/ptr.rs#L188-L193

I think the issue is to get solvers to recognize the bitwise operation `d & (x - 1)` (where `x` is a power of 2) as a modulo operation.

Self-contained problem:

```rust
use creusot_contracts::*;

#[bitwise_proof]
#[requires(i < 10usize && x@ == i@.pow2())]
#[ensures(result == (d@ % x@ == 0))]
pub fn f(d: usize, x: usize, i: usize) -> bool {
    // proof_assert!{x == 1usize << i};
    d & (x - 1) == 0
}
```

Lysxia:
Updating the `PtrOwn` invariant to take alignment into account is surprisingly tricky. Note that unlike the program method `is_aligned` which is only defined for `T: Sized`, the logic function `is_aligned_logic` is defined for all `T` as an intrinsic with an ad hoc definition for `T = [U]`, because I want `PtrOwn<[T]>`  to have the invariant that the pointer is aligned.

~~I gave up trying to prove `is_aligned_to` in the `extern_spec!`, and opened #1750 to come back to it later.~~

Close #1750


creusot-contracts/src/std/mem.rs
```diff
@@ -66,3 +70,15 @@ extern_spec! {
...
     dead
 }
+
+/// [`align_of`] as a logic `Int` value.
+///
+/// [`align_of`]: https://doc.rust-lang.org/std/mem/fn.align_of.html
+#[trusted]
+#[logic(open)]
+#[intrinsic("align_of_logic")]
+#[ensures(0 <= result && exists<i: Int> result == i.pow2())]
```
jhjourdan:
```suggestion
#[ensures(0 < result && exists<i: Int> 0 <= i && result == i.pow2())]
```

Otherwise it does not say anything because `pow2` is undefined for negatives.

creusot-contracts/src/std/num.rs
```diff
@@ -184,6 +184,20 @@ macro_rules! spec_type {
...
 }
 
+macro_rules! spec_unsized {
+    ($type:ty) => {
+        spec_type!($type);
+
+        extern_spec! {
+            impl $type {
+                    #[check(ghost)]
+                    #[ensures(result == exists<i: Int> self@ == i.pow2())]
```
jhjourdan: Same here.

creusot-contracts/src/std/ptr.rs
```diff
@@ -81,9 +114,22 @@ pub trait PointerExt<T: ?Sized>: Sized {
     #[logic]
     fn addr_logic(self) -> usize;
 
+    /// `true` if the pointer is null.
+    #[logic(open)]
```
jhjourdan: sealed?

creusot-contracts/src/std/ptr.rs
```diff
@@ -100,6 +146,11 @@ impl<T: ?Sized> PointerExt<T> for *const T {
     fn is_null_logic(self) -> bool {
         self.addr_logic() == 0usize
     }
+
+    #[logic(open)]
```
jhjourdan:
```suggestion
    #[logic(open, inline)]
```

creusot-contracts/src/std/ptr.rs
```diff
@@ -116,27 +167,78 @@ impl<T: ?Sized> PointerExt<T> for *mut T {
...
```
jhjourdan:
```suggestion
    #[logic(open, inline)]
```

(This will help you through `compute_specify`, which behaves a bit like `simpl` in Rocq.)

tests/creusot-contracts/creusot-contracts/why3session.xml
```diff
@@ -6,7 +6,7 @@
 <prover id="1" name="CVC4" version="1.8" timelimit="5" steplimit="0" memlimit="1000"/>
 <prover id="2" name="CVC5" version="1.0.5" timelimit="5" steplimit="0" memlimit="1000"/>
 <prover id="3" name="Alt-Ergo" version="2.6.0" timelimit="1" steplimit="0" memlimit="1000"/>
-<file format="coma" proved="true">
+<file format="coma">
```
jhjourdan: :warning:

creusot-contracts/src/std/ptr.rs
```diff
@@ -30,14 +37,40 @@ pub fn metadata_matches<T: ?Sized>(_value: T, _metadata: <T as Pointee>::Metadat
...
+#[intrinsic("is_aligned_logic_sized")]
+pub fn is_aligned_logic_sized<T>(ptr: *const T) -> bool {
+    ptr.is_aligned_to_logic(align_of_logic::<T>())
+}
+
+#[logic(open)]
+#[intrinsic("is_aligned_logic_slice")]
+pub fn is_aligned_logic_slice<T>(ptr: *const [T]) -> bool {
+    ptr.is_aligned_to_logic(align_of_logic::<T>())
 }
```
jhjourdan, line 83: BTW, why are we not using specialization for all these functions?

Lysxia: Idk. I will look into it!

tests/creusot-contracts/creusot-contracts/why3session.xml
Lysxia: Ongoing work!

creusot-contracts/src/std/ptr.rs
Lysxia:
```
error: logical functions cannot use the `default` modifier
```

jhjourdan: errr.... OK, that's probably because we don't know how to handle specializatio in the termination checker.

PR #1749, opened 2025-09-30
When determining resolution points, do not consider shared borrows to freeze the borrowee.

### LINKED ISSUE #1743: Spurious 'borrowed or uninitialized variable' error
The following code triggers a spurious error:
```rust
use creusot_contracts::{ghost::PtrOwn, *};

pub struct S {
    perm: PtrOwn<i32>,
    ptr: *const i32,
}

impl S {
    #[requires(true)]
    pub fn minimize(&mut self) {
        let r = unsafe { PtrOwn::as_ref(self.ptr, ghost!(&self.perm)) };
        #[invariant(0 <= self.perm.val()@)]
        loop {
            let _ = *r;
        }
    }
}
```
The error:
```
error: Use of borrowed or uninitialized variable self
  --> test.rs:13:21
   |
13 |         #[invariant(0 <= self.perm.val()@)]
   |                     ^^^^^^^^^^^^^^^^^^^^^
```

Apparently `self` is considered 'frozen' by the borrow checker at this point...

jhjourdan:
This seems obviously sound. I have no idea why we did not do that earlier.

@xldenis, any catch I have not thought about?

Fixes #1743.


xldenis:
> I have no idea why we did not do that earlier.

I'm pretty sure that was done at an earlier point, but this code has been rewritten many times so its very feasible that it got dropped. 

> @xldenis, any catch I have not thought about?

Looking at the generated coma, it seems like the resolution points get flipped to occur before the last use of the borrow? 

https://github.com/creusot-rs/creusot/pull/1749/files#diff-f50fae94e05989d71fe22991c46f4721a791cb63bbb94e03aa6575f80d03e95eL26-L27

The ones I've seen are all sound but does that hold in the general case?

jhjourdan:
> Looking at the generated coma, it seems like the resolution points get flipped to occur before the last use of the borrow?

Not quite as far as I can see. They get flipped with the last use of borrows *when it's an immutable use*, which is sound.

xldenis:
> They get flipped with the last use of borrows when it's an immutable use, which is sound.

Ok that makes sense.

PR #1752, opened 2025-10-01
Update to nightly-2025-10-01

Lysxia:


creusot/src/backend/program.rs
```diff
@@ -818,8 +818,6 @@ impl<'tcx> RValue<'tcx> {
                     _ => unsupported_cast(lower.ctx, span, source, target),
                 }
             }
-            RValue::Len(op) => Exp::qvar(lower.names.in_pre(PreMod::Slice, "length"))
```
jhjourdan, line 821: Is this replaced by anything?

Lysxia: In Rust it appears to be subsumed by a `UnOp` for accessing pointer metadata, generalized to arbitrary places, if I understand this PR correctly https://github.com/rust-lang/rust/pull/146564 And indeed we already handle this in the case `RValue::UnaryOp(UnOp::PtrMetadata, op)` https://github.com/creusot-rs/creusot/blob/91030ab0275e0804bd7fc02615767048f983f5d4/creusot/src/backend/program.rs#L879

PR #1753, opened 2025-10-02
impl Resolve for Seq, FMap, FSet

Lysxia:


jhjourdan:
Shouldn't we do that for other ghost containers like `FMap` and `FSet` too?

tests/should_succeed/cc/fmap.rs
```diff
@@ -0,0 +1,24 @@
...
+        #[invariant(produced.len() <= 1)]
+        #[invariant(produced.len() == 0 || ^snap_k == 1usize && ^snap_v == 1usize)]
+        #[variant(1 - produced.len())]
+        for (k2, v2) in s {
+            *k2 = 1;
+            *v2 = 1;
+        }
+        k + v
+    }
+}
```
Lysxia, line 24: Note this very "fun" example for testing `resolve` on keys...

PR #1754, opened 2025-10-02
Make PtrOwn opaque, and specify in its invariants that the underlying value verifies its invariant.

jhjourdan:


creusot-contracts/src/ghost/ptr_own.rs
```diff
@@ -21,31 +23,30 @@ use crate::*;
...
 }
 
 impl<T: ?Sized> Invariant for PtrOwn<T> {
-    #[logic(open, inline)]
+    #[logic(open, prophetic)]
     fn invariant(self) -> bool {
         !self.ptr().is_null_logic()
             && self.ptr_is_aligned_opaque()
             && metadata_matches(*self.val(), metadata_logic(self.ptr()))
+            && inv(self.ptr())
```
Lysxia, line 49: Shouldn't this be `inv(self.val())`?

jhjourdan: Indeed. My bad.

Lysxia: Fixed in #1783

PR #1763, opened 2025-10-06
Enable `check(ghost)` even for trusted functions

### LINKED ISSUE #1762: `#[trusted]` disables `validates_purity`
This generates invalid Why3:

```rust
use creusot_contracts::*;

#[check(ghost)]
pub fn faux() -> bool { false }

#[trusted]
#[requires(faux())] // program function called in logic
pub fn f() {}
```

Lysxia:
Close #1762


jhjourdan:
One potential issue with this fix is that the purity validator may crash on code patterns that we do not support. I had a quick look, it seems good, though.

PR #1765, opened 2025-10-07
Support for edition 2024

### LINKED ISSUE #962: Closure spec causes the variable to be incorrectly captured
The following code:
```
#[requires(p.0@ == 1)]
#[ensures(result@ == 2)]
fn disjoint_captures(mut p: (i32, i32)) -> i32 {
    let p0 = &p.0;
    let mut clos = #[ensures(p.1@ == 1)] || {
        p.1 = 1;
    };
    clos();
    *p0 + p.1
}
```
When using edition 2021, should at least compile (it does without the specifications), but the closure's spec causes a borrow error:
```
error[E0502]: cannot borrow `p` as mutable because it is also borrowed as immutable
  --> /.../test.rs:5:42
   |
4  |     let p0 = &p.0;
   |              ---- immutable borrow occurs here
5  |     let mut clos = #[ensures(p.1@ == 1)] || {
   |                              ---         ^^ mutable borrow occurs here
   |                              |
   |                              second borrow occurs due to use of `p` in closure
6  |         p.1 = 1;
   |         --- capture is mutable because of use here
...
10 |     *p0 + p.1
   |     --- immutable borrow later used here
```

### LINKED ISSUE #1538: Struct member projection not supported within quantified expressions
For example, this code:

```rust
use creusot_contracts::*;

struct Foo {
    bar: (),
}

#[predicate]
fn baz(foo: Foo) -> bool {
    pearlite! {
        forall<x: ()> foo.bar == ()
    }
}
```

makes creusot panic like so:

```text
thread 'rustc' panicked at creusot/src/backend/closures.rs:406:17:
assertion failed: cap.place.projections.is_empty()
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
   4: creusot::backend::closures::ClosSubst::pre
   5: creusot::translation::specification::pre_sig_of
   6: once_map::unsync::OnceMap<K,V,S>::get_or_try_insert
   7: creusot::validate::purity::validate_purity
   8: creusot::translation::before_analysis
   9: <creusot::callbacks::ToWhy as rustc_driver_impl::Callbacks>::after_expansion
  10: rustc_interface::passes::create_and_enter_global_ctxt::<core::option::Option<rustc_interface::queries::Linker>, rustc_driver_impl::run_compiler::{closure#0}::{closure#2}>::{closure#2}::{closure#0}
  11: rustc_interface::interface::run_compiler::<(), rustc_driver_impl::run_compiler::{closure#0}>::{closure#1}
```

jhjourdan:
- Add support for partial closures, both in specs and code
- Use stubs for IndexLogic and View to leverage auto-deref without importing public traits to the context.

Fixes #1538.
Fixes #962.

### DIFF
--- a/tests/should_succeed/persistent_array/proof.json
+++ b/tests/should_succeed/persistent_array/proof.json
@@ -265,7 +265,10 @@
       "vc_open": { "prover": "cvc5@1.3.1", "time": 0.018 },
       "vc_open'0": { "prover": "cvc5@1.3.1", "time": 0.029 },
       "vc_open'1": { "prover": "cvc5@1.3.1", "time": 0.032 },
-      "vc_set": { "prover": "z3@4.15.3", "time": 1.8 },
+      "vc_set": {
+        "tactic": "compute_specified",
+        "children": [ { "prover": "z3@4.15.3", "time": 0.165 } ]
+      },
       "vc_v_Some": { "prover": "cvc5@1.3.1", "time": 0.016 }
     },
     "M_persistent_array__testing": {


Lysxia:
Fancy!

Issue #1396, opened 2025-03-02
The old keyword seems to re-enables borrow checking


jhjourdan:
Consider the following code:

```rust
extern crate creusot_contracts;
use creusot_contracts::*;

#[requires(f.precondition(()))]
#[ensures(f.postcondition_once((), ()))]
fn apply_once<F: FnOnce()>(f: F) {
    f()
}

#[requires(f.precondition(()))]
#[ensures(exists<g: F> f.postcondition_mut((), g, ()) && resolve(&g))]
fn apply_mut<F: FnMut()>(mut f: F) {
    f()
}

fn foo() {
    let mut x = 13;
    let r = &mut x;
    let snap_r = snapshot!{ r };
    #[invariant((*r)@ == 13 + produced.len())]
    #[invariant(^*snap_r == ^r)]
    for _ in 0 .. 42 - 13 {
        // BEGIN loop body
        apply_mut(
            #[requires((*r)@ < 42)]
            #[ensures((*old(r))@ == (*r)@ + 1)]
            || *r += 1,
        );
        // END loop body
    }
    proof_assert!(x@ == 42);
}
```

This is rejected with the following error message:

```
   |
18 |     let r = &mut x;
   |         - move occurs because `r` has type `&mut i32`, which does not implement the `Copy` trait
...
22 |     for _ in 0 .. 42 - 13 {
   |     --- inside of this loop
...
26 |             #[ensures((*old(r))@ == (*r)@ + 1)]
   |                             - use occurs due to use in closure
27 |             || *r += 1,
   |             ^^ value moved into closure here, in previous iteration of loop
   |
help: consider creating a fresh reborrow of `r` here
   |
26 |             #[ensures((*old(&mut *r))@ == (*r)@ + 1)]
   |                             ++++++
```

Lysxia:
This counterexample is currently accepted (but why3find doesn't find a proof).

jhjourdan:
What if you fix the spec by replacing `#[ensures((*old(r))@ == (*r)@ + 1)]` with `#[ensures((*old(r))@ + 1 == (*r)@)]` ?

PR #1769, opened 2025-10-07
Add test for 1396

jhjourdan:


jhjourdan:
I needed to fix the example from #1396 in an interesting way: either we use `*old(x)` in the spec of the closure, and then `x` itself is captured, in which case we need to state that its prophecy does not change (i.e., `^old(x) == ^x`), or we use `old(*x)` (a new feature of #1765), and then only `*x` is captured, in which case we do not need to say that the prophecy of `x` does not change, because it is not captured.

Partial captures get more useful than I thought.

PR #1770, opened 2025-10-08
A bit of cleanup in generated names

jhjourdan:
- Fields of tuples are named f0....fn instead of _0, ..., _n. This avoids conflicts with locals names
- Same thing for names of closures fields: c0...cn
- Names for locals representing parameters shadow parameter names
- In the Why3 printer, use shadowing to print shadowing, which avoids generating new names.
- Use "return" instead of "return'" as the main continuation of every fonction. This avoid conflicts with other identifiers, since this is a keyword, and this is more idiomatic in Coma.


jhjourdan:
@Lysxia Could you please have a look? I think I have uncovered two bugs in the Why3 crate.

why3/src/coma.rs
```diff
@@ -241,19 +241,21 @@ impl Expr {
...
+                            let in_params = d
+                                .prototype
+                                .params
+                                .iter()
+                                .filter_map(|p| {
+                                    if let Param::Cont(n, _, _) = &*p { Some(n) } else { None }
+                                })
+                                .any(|n| n == cont);
+                            !in_params && d.body.occurs(cont)
+                        }))
```
Lysxia, line 258: Good catch.

why3/src/printer.rs
```diff
@@ -66,30 +66,9 @@ struct Scope {
...
-    /// constant x: usize
-    /// function f (x: usize): ()    (* x is shadowed here (also unused) *)
-    /// goal vc_f: ... POST(f(x-1)) -> POST(...)
-    /// ```
-    ///
-    /// I don't know if it's the only place to fix; we can probably avoid shadowing if we really want to.
-    undo: Vec<Vec<(Ident, Option<Symbol>)>>,
+    /// Remark 2: In cas of shadowing, we do not do anything, and keep the same symbold for the given
+    /// identifier
+    undo: Vec<Vec<Ident>>,
```
Lysxia, line 71: I can't remember what I thought went wrong before, but this looks OK to me now.

why3/src/coma.rs
```diff
@@ -455,11 +435,27 @@ impl Print for Expr {
...
+                                } else {
+                                    alloc.nil()
+                                },
+                                l.0.pretty_value_name(alloc, scope),
+                                ": ",
+                                l.1.pretty(alloc, scope),
+                                " = ",
+                                b
+                            ]
+                        }),
```
Lysxia, line 458: OK yeah you want to print the bodies first.

Lysxia (approved):

PR #1781, opened 2025-10-10
Ignore Fake borrows

Lysxia:


jhjourdan:
Why was the old situation bad ?

Lysxia:
It seems we never ran into this case before, since the tests case are not affected by this change. But I finally ran into [a counterexample](https://github.com/creusot-rs/creusot/blob/3e4c539ce105c67d4a49cfa4d422b803027be673/tests/should_succeed/bug/fake_ref.rs) that compiled to absolute nonsense: in each branch there are fake borrows that read from a different constructor than the matched one.

PR #1783, opened 2025-10-14
Add size_of_val_logic

Lysxia:


creusot-contracts/src/ghost/ptr_own.rs
```diff
@@ -43,10 +46,17 @@ impl<T: ?Sized> PtrOwn<T> {
...
-        !self.ptr().is_null_logic()
-            && self.ptr_is_aligned_opaque()
-            && metadata_matches(*self.val(), metadata_logic(self.ptr()))
-            && inv(self.ptr())
+        pearlite! {
+            !self.ptr().is_null_logic()
+                && self.ptr_is_aligned_opaque()
+                && metadata_matches(*self.val(), metadata_logic(self.ptr()))
+                // Allocations can never be larger than `isize` (source: https://doc.rust-lang.org/std/ptr/index.html#allocation)
+                && size_of_val_logic(self.val()) <= isize::MAX@
```
jhjourdan, line 69: Cannot we make this part of the post-condition of `size_of_val`?

creusot-contracts/src/std/mem.rs
```diff
@@ -74,6 +74,32 @@ pub fn size_of_logic<T>() -> Int {
     dead
 }
 
+#[trusted]
+#[logic(open)]
```
jhjourdan:
```suggestion
#[logic(open, inline)]
```

creusot-contracts/src/std/mem.rs
```diff
@@ -74,6 +74,32 @@ pub fn size_of_logic<T>() -> Int {
...
+#[intrinsic("size_of_val_logic")]
+#[ensures(0 <= result)]
+pub fn size_of_val_logic<T: ?Sized>(val: &T) -> Int {
+    dead
+}
+
+#[logic(open)]
+#[intrinsic("size_of_val_logic_sized")]
+pub fn size_of_val_logic_sized<T>(_val: &T) -> Int {
```
jhjourdan: Shouldn't this functions be made private? They seem to be implementation details to me (same for metadata, etc...).

creusot-contracts/src/ghost/ptr_own.rs
Lysxia: No because you may not have a call to `size_of_val` where you need this, and even if you tried it's quite inconvenient to call given a PtrOwn: you have to turn the pointer into a reference.

creusot-contracts/src/std/mem.rs
Lysxia: fair enough.

creusot-contracts/src/ghost/ptr_own.rs
Lysxia: I guess it could still be a postcondition to `size_of_val` as an addition independent from this `PtrOwn` change.

Lysxia: Actually, now that I think about it, this could be exploited due to the fact that Rust checks the sizes of constants quite late, a mix of code gen time and run time. If `size_of_val` has `#[ensures(result@ <= isize::MAX@)]`, the following proves `false`:

```
#[ensures(false)]
pub fn main() {
    ghost! {
        let x = [0usize; usize::MAX];
        let _ = ::std::mem::size_of_val(&x);
    };
}
```

For `PtrOwn` we mitigate this by making its constructor methods `check(terminates)` instead of `check(ghost)`. This is already the case for `from_box` but `from_ref` and `from_mut` should be fixed too.

jhjourdan: Hum, that's a good point. Could a copy-past this explanation to a comment?

Lysxia: OK I've added comments about it on `PtrOwn` and on `size_of_val`.

creusot-contracts/src/std/mem.rs
```diff
@@ -67,18 +81,48 @@ extern_spec! {
...
 pub fn size_of_logic<T>() -> Int {
     dead
 }
 
+/// [`size_of_val`] as a logic `Int` value.
+#[trusted]
+#[logic(open, inline)]
+#[intrinsic("size_of_val_logic")]
+#[ensures(0 <= result)]
+pub fn size_of_val_logic<T: ?Sized>(val: &T) -> Int {
```
jhjourdan, line 96: Another nit: why does this take a reference? Unsized params are allowed!

Lysxia: Oh right!

jhjourdan: And yet another thing: `is_aligned_logic_str` and friends should be private too.

Lysxia: Already done!

PR #1784, opened 2025-10-14
Change erasure to consider `*const` and `*mut` as distinct

Lysxia:
Originally I added this hack because `PtrOwn::from_box` was already using `*const` and I wasn't sure if it was worth changing. And the hack was only limited to argument types. Now that I'm looking at it again it seems more idiomatic to use `*mut` in these places, which simplifies the erasure check.

### DIFF
--- a/tests/should_fail/wrong_permissions.coma
+++ b/tests/should_fail/wrong_permissions.coma
@@ -291,8 +291,8 @@ module M_wrong_permissions__wrong_ptr_own_permission
     | bb1 = s0
       [ s0 = [ &ptr'0 <- _2.f0 ] s1 | s1 = new {(1: Int32.t)} (fun (_ret: tuple) -> [ &_4 <- _ret ] s2) | s2 = bb2 ]
     | bb2 = s0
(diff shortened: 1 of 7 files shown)


creusot-contracts/src/ghost/ptr_own.rs
```diff
@@ -80,7 +80,7 @@ impl<T> PtrOwn<T> {
     /// cell initialized with `v`.
     #[check(terminates)] // can overflow the number of available pointer adresses
     #[ensures(result.1.ptr() == result.0 && *result.1.val() == v)]
-    pub fn new(v: T) -> (*const T, Ghost<PtrOwn<T>>) {
+    pub fn new(v: T) -> (*mut T, Ghost<PtrOwn<T>>) {
```
Lysxia, line 83: @jhjourdan  Does this look reasonable to you, or am I missing a reason to use `*const` here?

jhjourdan: The main real difference is variance, if I understand well. We don't care about variance, because values are in fact carried by `PtrOwn`. So I guess this has no impact...

PR #1787, opened 2025-10-15
Index `Seq` with a pair of integers

dianegolfouse:
Ultimately we may want to index with tuples or even sequences of integers. For now this is all I need.


creusot-contracts/src/logic/seq.rs
```diff
@@ -539,6 +539,24 @@ impl<T> Seq<T> {
         panic!()
     }
 
+    #[check(ghost)]
+    #[requires(0 <= index.0 && index.0 < self.len() && 0 <= index.1 && index.1 < self.len())]
+    #[ensures(*result.0 == self[index.0] && *result.1 == self[index.1])]
+    pub fn multi_index(&self, index: (Int, Int)) -> (&T, &T) {
+        (&self[index.0], &self[index.1])
+    }
```
xldenis: any reason why you wouldn't implement `Index/IndexMut` instead?

dianegolfouse: The type is not the same, `&(T, T)` vs `(&T, &T)`. I wasn't comfortable returning a reference to a pair if there isn't "really" one.

jhjourdan: I think I agree with @xldenis, using `IndexLogic` would be much cleaner.

We already have borrows with are not actual borrows (see `FMap::split_mut_ghost`).

dianegolfouse: Alright, let's do that

creusot-contracts/src/logic/seq.rs
```diff
@@ -692,6 +692,34 @@ impl<T> ::std::ops::IndexMut<Int> for Seq<T> {
...
+}
+
+impl<T> ::std::ops::IndexMut<(Int, Int)> for Seq<T> {
+    #[trusted]
+    #[check(ghost)]
+    #[requires(0 <= index.0 && index.0 < self.len() && 0 <= index.1 && index.1 < self.len())]
+    #[requires(index.0 != index.1)]
+    #[ensures((*result).0 == (*self)[index.0] && (*result).1 == (*self)[index.1]
+           && (^result).0 == (^self)[index.0] && (^result).1 == (^self)[index.1])]
+    #[ensures(forall<i> 0 <= i && i < self.len() && i != index.0 && i != index.1 ==> (*self)[i] == (^self)[i])]
```
Lysxia:
```suggestion
    #[ensures(forall<i> i != index.0 && i != index.1 ==> (*self).get(i) == (^self).get(i))]
```

jhjourdan:
At some point, we will need to do that for `FMap`, too.

PR #1789, opened 2025-10-15
Add ptr and slice methods

Lysxia:
- Pointer arithmetic in logic: `offset_logic`, `sub_logic`
- Pointer extensionality (currently for slice pointers): a fat pointer is entirely described by the pair of its thin pointer and its metadata
- Extern specs for `<*const T>::len` (`metadata_logic` is its logic version), `slice_from_raw_parts` (constructing a slice pointer from a thin pointer and a length).
- `as_mut_ptr_own` (`&mut [T]` to a pair of pointer and `PtrOwn`)


creusot-contracts/src/std/ptr.rs
```diff
@@ -180,6 +180,116 @@ impl<T: ?Sized> PointerExt<T> for *mut T {
     }
 }
 
+/// Extension methods for `*const T` where `T: Sized`.
```
jhjourdan, line 183: It's not forbidden to move that in `PointerExt` and add `where` contraints to logical methods.

creusot-contracts/src/std/ptr.rs
```diff
@@ -180,6 +180,116 @@ impl<T: ?Sized> PointerExt<T> for *mut T {
...
+
+    #[logic(law)]
+    #[ensures(self.offset_logic(offset).sub_logic(self) == offset)]
+    fn sub_offset_logic(self, offset: Int);
+}
+
+impl<T> SizedPointerExt<T> for *const T {
+    #[trusted]
+    #[logic(opaque)]
+    #[ensures(self.addr_logic()@ + offset * size_of_logic::<T>() < usize::MAX@ ==> result.addr_logic()@ == self.addr_logic()@ + offset * size_of_logic::<T>())]
```
jhjourdan: Any reason why not using a pre/post pair instead of an implication?

creusot-contracts/src/std/ptr.rs
Lysxia: Ah: `error: Laws cannot have additional generic parameters or trait constraints`

creusot-contracts/src/std/slice.rs
```diff
@@ -102,6 +105,19 @@ impl<T> SliceExt<T> for [T] {
         let (ptr, own) = PtrOwn::from_ref(self);
         (ptr as *const T, own)
     }
+
+    /// Convert `&mut [T]` to `*mut T` and a mutable ownership token.
+    #[trusted]
```
jhjourdan: This should not be trusted.

creusot-contracts/src/std/ptr.rs
jhjourdan: Ok, fair enough, we need to do something with this law mechanism, it's really broken.

creusot-contracts/src/std/ptr.rs
Lysxia: No reason. Fixed.

PR #1790, opened 2025-10-15
Restrict `PtrOwn::disjoint_lemma` to non-zero sized types and add axioms for non-zero sized types

### LINKED ISSUE #1263: Zero sized types and raw pointers
creating a `PtrOwn` from e.g., a `Box`, should require that the underlying value is not zero-size, because uniqueness guarantees do not hold for ZSTs.

We currently have an assert for this, but we shuould instead check this statically using a pre-condition.

Lysxia:
See counterexample in `should_fail/zst.rs`

- Remove assert in `from_box` (close #1263)
- Generate axiom when we can determine that `size_of_logic > 0` (to fix the tests (`union_find` etc.) that use `disjoint_lemma`)

This axiom may leak private details of a type: adding or removing some private fields may change its "non-zero sizedness". But this is not much worse than what leaks through `size_of` already.


Lysxia:
This breaks the `union_find` proofs because we know nothing about the size of user-defined types. In general there is very little that is guaranteed about the default `repr(Rust)` types. In this case, we could say that `Node` must be non-zero sized because it has at least one non-empty and non-zero sized variant (`Link` contains a pointer).

jhjourdan:
> it has at least one non-empty and non-zero sized variant (Link contains a pointer).

Or because it has more than one variant (which means that we need to store the discriminant).

Lysxia:
related https://github.com/creusot-rs/creusot/issues/1263

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -299,10 +299,126 @@ fn expand_logic<'tcx>(
...
+    } else {
+        false
+    }
+}
+
+/// If `true`, the type is definitely inhabited. This is a best-effort underapproximation.
+fn is_inhabited<'tcx>(tcx: TyCtxt<'tcx>, ty: Ty<'tcx>) -> bool {
+    use rustc_type_ir::TyKind::*;
+    match ty.kind() {
+        Bool | Char | Int(_) | Uint(_) | Float(_) | Ref(_, _, _) | RawPtr(_, _) => true,
```
jhjourdan: &T and &mut T are not inhabited when T isn't. Not sure rustc is able to leverage this, though.

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -299,10 +299,126 @@ fn expand_logic<'tcx>(
...
+        Adt(def, args) => def.is_box() || is_inhabited_adt(tcx, def, args),
```
jhjourdan: Same thing for Box.

creusot/src/backend/clone_map/elaborator.rs
jhjourdan: The problem being that if you recurse, the algorithm may not terminate...

PR #1794, opened 2025-10-16
Fix spec of Vec::clone

Lysxia:


jhjourdan:
So silly... well spotted!

PR #1795, opened 2025-10-17
Make private fields opaque

jhjourdan:
Also refactored generation of structural invariant/resolve, to make the VCs simpler to prove and easier to read.

Fixes #418.

Issue #1122, opened 2024-09-25
Static analysis for type invariants


jhjourdan:
We need a static analysis to determine when type invariants holds for places. This can be useful for several purposes:

- Automatically add loop invariants stating that type invariants are maintained by a loop. This would remove a lot of instances of `#[invariant(inv(...))]` in our test suite.
- Avoid emitting VC for type invariants when resolving borrows.

Note that there are many places type invariants are enforced nowadays, since each creation of a mutable borrow asserts the invariant holds for the prophecy. Additionally, we could add a macro `fold_inv!(place)`, which would assert the type invariant for `place`, and use this fact in the static analysis.

PR #1797, opened 2025-10-20
Fix unsoundnesses in #1795

jhjourdan:
By assuming that foreign opaque types had trivial resolve and trivial type invariant, I made two axioms unsound:
- structural_resolve(self) => resolve(self) (wrong as soon as structural_resolve is true and resolve non-trivial)
- post-conditions of accessors such as `Skip::iter`.

I fixed this by:
- Never making structural_resolve trivial for opaque types
- For foregin types, do not make type invariants trivial (instead, it is an uninterpreted function)

### DIFF
--- a/creusot-contracts/src/std/iter/filter_map.rs
+++ b/creusot-contracts/src/std/iter/filter_map.rs
@@ -29,7 +29,9 @@ impl<B, I: Iterator, F: FnMut(I::Item) -> Option<B>> Invariant for FilterMap<I, 
 }
 
 #[logic(prophetic)]
-pub fn private_invariant<B, I: Iterator, F: FnMut(I::Item) -> Option<B>>(f: FilterMap<I, F>) -> bool {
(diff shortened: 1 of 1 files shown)


jhjourdan:
Sadly, I had to add new instances of `#[invariant(inv(xxxx))]` in tests. We really need to address #1122 at some point.

xldenis:
> Never making structural_resolve trivial for opaque types

> For foreign types, do not make type invariants trivial (instead, it is an uninterpreted function)

I feel like it should be possible to eliminate those proof obligations in most situations no? Even if `Skip` has an invariant you can't ever break it so do we really need to lug around pointless symbols? However, i doubt this has a large impact on proofs (apart from making terms harder to read).

jhjourdan:
> I feel like it should be possible to eliminate those proof obligations in most situations no? Even if Skip has an invariant you can't ever break it so do we really need to lug around pointless symbols? However, i doubt this has a large impact on proofs (apart from making terms harder to read).

Indeed. I think this is essentially what #1122 is about.

PR #1799, opened 2025-10-20
Better error message for incorrect usage of `dead`

dianegolfouse:


creusot/src/translation/pearlite/from_thir.rs
```diff
@@ -353,11 +359,11 @@ impl<'tcx> ThirTerm<'_, 'tcx> {
                         Ok(Term { ty, span, kind: TermKind::Old { term: Box::new(term) } })
                     }
                     Intrinsic::ClosureResult => Ok(Term::unit(self.ctx.tcx).span(span)),
-                    Intrinsic::Dead => Err(Error::msg(
+                    Intrinsic::Dead => Err(self.ctx.dcx().span_err(
                         span,
-                        "The `dead` term can only be used for the body of trusted logical functions",
+                        "The `dead` term can only be used for the body of `logic(opaque)` functions",
```
jhjourdan, line 364: They should not beused either if the function has a post-condition and is not trusted.

PR #1801, opened 2025-10-20
Simplify red-black-trees, support for @ patterns in pearlite.

jhjourdan:


tests/should_succeed/red_black_tree.rs
```diff
@@ -184,9 +165,7 @@ impl<K: DeepModel, V> Resolve for Tree<K, V> {
 impl<K: DeepModel, V> Resolve for Node<K, V> {
     #[logic(prophetic)]
     fn resolve(self) -> bool {
-        pearlite! {
-            forall<k, v> self.has_mapping(k, v) ==> resolve(v)
-        }
```
xldenis, line 189: i wonder if it wouldn't make sense to move to a closure syntax for quantifiers now? I feel like we're probably in a state where we could _literally_ use closures and define `fn forall<P : ...>(p: P) -> bool { forall<x> p(x) }` that would allow us to take a huge step to fully removing this macro.

xldenis: (which would be a huge ergonomics win)

jhjourdan:
OTOH, we still have the `@` syntax, and I think our syntax for `forall` is more ergonomic than a closure-based one.

PR #1803, opened 2025-10-21
Refactor namespaces in creusot-contracts, edition 2024 in tests

jhjourdan:
- Use edition 2024 in tests
- Do no longer use shadowing in e.g., creusot_contracts::std::iter::Iterator
- Do no longer re-export std in creusot_contracts::std (this causes ambiguities in imports with edition 2024).
- Do no longer re-export prelude in creusot_contracts (this causes ambiguities, because both std and creusot_contracts::std are in scope.)
- FMap and FSet always require sizedness of type parameters (coherence, we anyway cannot provide enough interesting methods if not sized).
- Test red_black_tree : a few simplifications/optimizations


creusot-contracts/src/logic.rs
```diff
@@ -18,6 +18,7 @@ pub mod seq;
 mod set;
 mod well_founded;
 
+// TODO: do not alias these names. decide either to re-export them, or use a submodule.
```
Lysxia, line 21: Is this not how you reexport them?

jhjourdan: The `fset` module is public, and we reexport `fset::FSet`, which gives two different public paths for `fset::FSet`, which is confusing. I'd like to make a choice between these two possibilities.

If we use the `fset` module, then we should not reexport `fset::FSet` (which makes sense to me).
If, on the contrary, we do not expose the `fset` module, then we need to do something with all th lemmas there which are not part of an `impl` block. In any case, I think it would make sense to make them part of an `impl` block.

PR #1807, opened 2025-10-24
Generate better coma names

Lysxia:
- Give human-readable names to traits and impl blocks instead of hashes
  (e.g., `trait_{TraitName}`, `impl_{TyName}`, `impl_{TraitName}_for_{TyName}`).
  In case of collision (such as for impl blocks for the same type), just append a counter.
- Include whole types in their Coma name (e.g., `Option<usize>` becomes `type t_Option_usize` instead of just `t_Option`).
- Include first type argument in function names (e.g., `inv::<Vec<T>>` becomes `inv_Vec_T_Global`).
  We don't include all type arguments to avoid getting too long names.
  For methods, we include the Self type instead.
- Strip `_logic` suffix of function names to shorten goals.
- Don't prefix public field names with their type. Unnamed field are named `f{index}` and named field names are just `{name}`
  (or `f_{name}` if `{name}` starts with an uppercase letter).
- Don't prefix constructor names with `C_` unless they start with a lowercase letter.
- Use the same disambiguating suffix for fields and constructor names of the same type
  (e.g., we may generate records with field names `{f0'1; f1'1}` but not `{f0'1; f0'0}`).
- Rename constructor eliminators from `v_{Constr}` to `elim_{Constr}`.


jhjourdan:
Excellent!

I haven't read all the code, and I'm slightly worried about name conflicts if the encodings are not injective. But I guess you was cautious enough?

Why are there some hashes remaining in some names?

Lysxia:
Yeah this is handled in `naming::ComaNames`. I keep track of what strings are already assigned to a `LocalDefId` and bump a counter until I find an unused string.

>Why are there some hashes remaining in some names?

Where?

tests/should_fail/alloc_max_size.coma
```diff
@@ -1,28 +1,31 @@
...
   
-  predicate inv [@inline:trivial] (_1: Seq.seq UInt64.t) = true
+  predicate inv_Seq_usize [@inline:trivial] (_1: Seq.seq UInt64.t) = true
   
-  meta "rewrite_def" predicate inv
+  meta "rewrite_def" predicate inv_Seq_usize
   
-  predicate invariant' (self: Slice64.array UInt64.t) =
-    inv (Slice64.view self) /\ Seq.length (Slice64.view self) = UInt64.t'int (18446744073709551615: UInt64.t)
+  predicate invariant_array_usize_18446744073709551615 (self: Slice64.array UInt64.t) =
```
jhjourdan, line 12: Here.

jhjourdan: Oh but this is not a hash, it's the size of the array, `usize::MAX`!

Lysxia: That's the length of the array! (BTW I think this test is broken, it should fail but in reality it would fail at codegen time, and our tests stop before that.)

jhjourdan: No, I think this is unsound actually. Codegen would not fail, because ghost code is eliminated when Creusot mode is not activated.

I think we should simply forbid `size_of_val` in ghost code.

PR #1810, opened 2025-10-24
Move installation to XDG_DATA_HOME

### LINKED ISSUE #1798: `.config/` files should be moved to `.local/share/`
These files are arguably more part of "the Creusot installation" (= `.local/share/creusot`) than user-written configuration (= `.config/creusot`). This change would make it easier to Nixify Creusot.

So Creusot should work first without anything in `.config` and only then we may consider allowing user-config on top of that.

Lysxia:
Close #1798

- Creusot now works starting with no files in `.config/creusot` (or more precisely `$XDG_CONFIG_HOME/creusot`).
- Creusot generates `why3.conf` in `.config/creusot` the first time it is needed (for why3 or why3find). We do this late generation because the location of solvers may change before writing it in that file (cf. Nix). And we place it in `.config/creusot` because it is a "user-writable" file via Why3 IDE.
- For uniformity with solvers, `why3` and `why3find` are now expected to be in `$XDG_DATA_HOME/creusot/bin` (possibly as symlinks). For expert users, this allows easily switching between different versions of these binaries (without messing with the local opam switch if you have one).
- Add `cargo creusot version` command (replaces `status`).
- Remove the version check by default, so changing dependencies is easier. You can still check that you have the "recommended" versions of provers (= those used by our tests) by running `cargo creusot version`.
- Add `cargo creusot why3-conf` to (re)generate `why3.conf` with a custom `provers-parallelism` (mainly for CI and the test suite).
- Add Installation section in the guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,31 +63,8 @@ More examples are found in [tests/should_succeed](tests/should_succeed).
    cargo creusot --help
    ```
 
-   A regular installation consists of:
-   - the `cargo-creusot` executable in `~/.cargo/bin/`;
-   - the `creusot-rustc` executable in `{DATA_DIR}/toolchains/$TOOLCHAIN/bin`;
-   - the `why3` and `why3find` executables in `{SWITCH}/bin`;
-   - the Creusot prelude in `{SWITCH}/lib/why3find/packages/creusot`;
-   - SMT solvers (Alt-Ergo, CVC4, CVC5, Z3) in `{DATA_DIR}/bin`;
-   - configuration files in `{CONFIG_DIR}`.
-
-where the directories depend on the OS:
-
-| | Linux | MacOS |
-|-|-|-|
-| `{DATA_DIR}` | `~/.local/share/creusot` | `~/Library/Application Support/creusot.creusot` |
-| `{CONFIG_DIR}` | `~/.config/creusot` | `~/Library/Application Support/creusot.creusot` |
-| `{SWITCH}` | `~/.local/share/creusot/_opam` | `~/.creusot/_opam` |
-
-Installation options can be set in a text file `INSTALL.opts`.
-They are just space-separated command-line arguments.
-Type `./INSTALL --help` for a list of available options.
-For example:
-
-```sh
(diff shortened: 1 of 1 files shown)


jhjourdan:
This apparently broke the deploy CI script.

Lysxia:
mdbook 0.5.0 came out overnight and removed the `multilingual` field https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md#mdbook-050-beta1

jhjourdan:
And what about that:

```
error: invalid value 'why3-and-why3find' for '--external <TOOL>'
  [possible values: alt-ergo, z3, cvc4, cvc5]
```

?

Lysxia:
I broke that!

jhjourdan:
When trying to do tests for #1813, I get the following error:

```
$ cargo creusot 
Error: `cargo metadata` exited with an error: error: failed to load source for dependency `creusot-contracts`

Caused by:
  Unable to update /home/opam/creusot/creusot-contracts

Caused by:
  failed to read `/home/opam/creusot/creusot-contracts/Cargo.toml`

Caused by:
  No such file or directory (os error 2)
```

Is this caused by this PR too?

Lysxia:
No it's the `Cargo.toml` of that project that's not portable (we don't have a documented portable way of depending on a local creusot-contracts).

jhjourdan:
(NEvermind, this is unreleated.)

Honey-Be:
> When trying to do tests for #1813, I get the following error:
> 
> ```
> $ cargo creusot 
> Error: `cargo metadata` exited with an error: error: failed to load source for dependency `creusot-contracts`
> 
> Caused by:
>   Unable to update /home/opam/creusot/creusot-contracts
> 
> Caused by:
>   failed to read `/home/opam/creusot/creusot-contracts/Cargo.toml`
> 
> Caused by:
>   No such file or directory (os error 2)
> ```
> 
> Is this caused by this PR too?

You can do tests for it on the pre-configured devcontainer of the project, which the Cargo.toml is configured for.

PR #1812, opened 2025-10-27
Make size_of, size_of_val, and align_of terminates instead of ghost

Lysxia:
Context: https://github.com/creusot-rs/creusot/pull/1807#discussion_r2460173616

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -40,11 +40,7 @@ jobs:
         key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock', 'prelude-generator/**') }}
     - name: Install Creusot
       run: |
-        ./INSTALL --external why3-and-why3find --skip-extra-tools
-    - name: Dummy creusot setup
-      run: |
-        mkdir -p ~/.config/creusot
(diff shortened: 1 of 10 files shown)


tests/should_fail/alloc_max_size.stderr
```diff
@@ -0,0 +1,16 @@
+warning: unused import: `ghost::PtrOwn`
```
jhjourdan: TODO.

jhjourdan:
Why is there any risk of overflow for `align_of`?

Lysxia:
You can use `repr(align(N))` to set arbitrary values of alignment [between 1 and 2^29](https://doc.rust-lang.org/reference/type-layout.html#r-layout.repr.alignment.constraint-alignment) and we seem to want Creusot to work on 16-bit.

jhjourdan:
Hum... ok, fair enough. (It's unclear what happens if you ask for a 2^28 alignment on a 16 bits machine, though.)

xldenis:
>Hum... ok, fair enough. (It's unclear what happens if you ask for a 2^28 alignment on a 16 bits machine, though.)

Should be easy to test, but I suspect that you get a compilation error.

PR #1820, opened 2025-10-27
Update toolchain to 2025-11-13

Lysxia:
Weird: I'm getting warnings about unused variables for all logic functions. Reason: the unused variables lint now works on MIR https://github.com/rust-lang/rust/pull/142390


jhjourdan:
Are you sure adding the `allow` attribute is the right solution?

jhjourdan:
I mean, this implies that we will never leverage unused warnings in pearlite.

Lysxia:
No I'm not sure yet. I'm still just trying stuff out.

Lysxia:
My current plan is to make Pearlite go through the liveness analysis in MIR. The challenge is to dodge the borrow checker. We currently do that by removing the code of Pearlite closures, but now I need to find a way to put it back.

jhjourdan:
Could you explain the problem?

Lysxia:
The liveness analysis is what enables the `unused_variables` warning. Before, it was performed on HIR (or THIR, I'm not sure), but now it is performed on MIR, after borrow checking. Since we remove Pearlite code from MIR before borrow checking, all Pearlite variables are now considered unused. This includes parameters of all logic functions. This results in a lot of spurious warnings.

xldenis:
it seems like it should be possible to perform the lowering to mir, run the liveness analysis ourselves for error reporting but then continue to no-op the body in queries to trick borrowck.

jhjourdan:
Is that liveness analysis the same as the one used in borrowck? If so, we should be careful not to break borrowck by providing it inconsistent information.

Lysxia:
No it happens after `mir_borrowck`

Lysxia:
I've got a working fix for the unused variable warnings, but it currently relies on exposing the private definition of `Steal` with a very unsafe transmute. I've sent a PR to rust to expose the method I actually need.

creusot/src/cleanup_spec_closures.rs
```diff
@@ -1,56 +1,170 @@
...
+        RefCell::new(HashMap::new());
+}
+
+pub(crate) fn restore_mir_for_liveness_check<'tcx>(tcx: TyCtxt<'tcx>, local_id: LocalDefId) {
+    let def_id = local_id.to_def_id();
+    let (mir, _) = tcx.mir_promoted(local_id);
+    // This is a terrible hack but it shouldn't affect correctness:
+    // we've already copied MIR bodies right after borrow checking.
+    // This is only used for warnings about unused variables.
+    let mut mir = mir.risky_hack_borrow_mut();
```
Lysxia, line 33: Shiny new hack that was merged in rustc this week!

PR #1825, opened 2025-10-30
Untyped elements for union-find

dianegolfouse:
This doesn't change the proof at all, but makes the internal structures (e.g. `UFInner`) easier on the eyes.
Also I guess this makes this interface slightly more flexible.


tests/should_succeed/union_find.rs
```diff
@@ -73,7 +73,7 @@ mod implementation {
             (forall<e> /*#[trigger(self.0.domain.contains(e))]*/ self.0.domain.contains(e) ==>
                 // this invariant was not in the why3 proof: it ensures that the keys and the payloads of `perm` agree
                 self.0.perms.contains(e) &&
-                self.0.perms[e].ptr() == e.0 &&
+                self.0.perms[e].ptr() == e.0 as *const () as *const _ &&
```
jhjourdan: Two subsequent coercions necessary?

dianegolfouse: Ah, I guess `*mut T -> *const T` is automatic... nice!

Issue #1828, opened 2025-11-04
Cannot found the `HirId` of a `LocalDefId` in  `gather_intrinsics`


Halbaroth:
I tried the latest version of Creusot on my axiomatized Hashbrown crate and I ran into a bug! 
Calling `RUST_BACKTRACE=1 cargo creusot` produces the following backtrace:
```
thread 'rustc' panicked at /rustc-dev/fa3155a644dd62e865825087b403646be01d4cef/compiler/rustc_middle/src/hir/mod.rs:220:84:
index out of bounds: the len is 3397 but the index is 3397
stack backtrace:
   0:     0x7fe82d95718b - std::backtrace::Backtrace::create::hde1940a9aa55c90f
   1:     0x7fe82d9570d5 - std::backtrace::Backtrace::force_capture::h39b97da0bc6a5fb8
   2:     0x7fe82c9827d2 - std[7e1b8c565ae6a770]::panicking::update_hook::<alloc[8e0e9e80cef66960]::boxed::Box<rustc_driver_impl[75770ce4a31f8030]::install_ice_hook::{closure#1}>>::{closure#0}
   3:     0x7fe82d9580bf - std::panicking::panic_with_hook::h3136bc18e19ec6ee
   4:     0x7fe82d957e7a - std::panicking::panic_handler::{{closure}}::haa99ed2ac62a97d2
   5:     0x7fe82d951d79 - std::sys::backtrace::__rust_end_short_backtrace::he1a6c69637605395
   6:     0x7fe82d93279d - __rustc[d556568c0434a7c8]::rust_begin_unwind
   7:     0x7fe82a980bd0 - core::panicking::panic_fmt::hb9dc3f33c24f4370
   8:     0x7fe82bd1477c - core::panicking::panic_bounds_check::h2e5253339b5927a4
   9:     0x7fe82e4594b1 - rustc_query_impl[86c61af390cc4ddf]::plumbing::__rust_begin_short_backtrace::<rustc_query_impl[86c61af390cc4ddf]::query_impl::local_def_id_to_hir_id::dynamic_query::{closure#2}::{closure#0}, rustc_middle[1e514162b26a76c0]::query::erase::Erased<[u8; 8usize]>>
[...]

PR #1829, opened 2025-11-04
Attempt at fixing #1828 by limiting the kinds of DefId that we consider for intrinsics.

jhjourdan:


jhjourdan:
@Halbaroth : could you try this PR?

Halbaroth:
It works :) thanks!

PR #1831, opened 2025-11-06
Better error message when pattern matching on Int in Pearlite

### LINKED ISSUE #1827: Better error message when pattern matching on `Int` in Pearlite
See #1826. I don't think we can easily support this. So instead we should emit a proper error message.

mcoulmance:
fix #1827

### DIFF
--- a/creusot-deps.opam
+++ b/creusot-deps.opam
@@ -5,9 +5,9 @@ maintainer: "Armaël Guéneau <armael.gueneau@inria.fr>"
 authors: "the creusot authors"
 depends: [
   "ocaml" {= "5.3.0"}
-  "why3" {= "git-d5877421a"}
-  "why3-ide" {= "git-d5877421a" & !?in-creusot-ci}
-  "why3find" {= "1.2.0"} # if these versions changes don't forget to update creusot-setup/src/tools_versions_urls.rs
+  "why3" {= "git-51e0579b"}
+  "why3-ide" {= "git-51e0579b" & !?in-creusot-ci}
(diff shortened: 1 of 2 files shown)


creusot-contracts-proc/src/creusot/pretyping.rs
```diff
@@ -634,6 +634,10 @@ fn encode_arm_(arm: &TermArm, locals: &mut Locals) -> Result<TokenStream, Encode
     if arm.guard.is_some() {
         return Err(EncodeError::Unsupported(arm.span(), "match guard".to_string()));
     }
+    // See #1827 for details
+    if matches!(arm.pat, Pat::Lit(syn::ExprLit{lit: Lit::Int(_), .. })) {
+        return Err(EncodeError::Unsupported(arm.span(), "Pattern matching literals on Int are unsupported by Pearlite. Consider using if-then-else instead.".to_string()));
+    }
```
jhjourdan: This will only work if the litteral is a top-level pattern. I.e., the following will not raise the error:

```rust
match o {
  Some(0) => ...
  ...
}

jhjourdan: On possible fix is to refactor the visitor used in `pattern_bind`: define a type `PatEncoder`, with a field `&mut Locals`, and implement `Visit` for it. In the visitor, you can do the vriable binding tracking (as is currently done), and add the check for litteral patterns. If it contains a litteral pattern, the it would set a Boolean in the `PatEncoder`, whouch would then raise an error in `pattern_bind`.

jhjourdan:
LGTM, thanks !

PR #1839, opened 2025-11-26
Improve specs of bitwise operations

Lysxia:


creusot-contracts/src/std/num.rs
```diff
@@ -202,6 +202,30 @@ macro_rules! spec_unsized {
                     #[check(ghost)]
                     #[ensures(result == (self != $zero && self & (self - $one) == $zero))]
                     fn is_power_of_two(self) -> bool;
+
+                    #[check(ghost)]
+                    #[ensures(self >> $type::BITS - result == $zero)]
+                    #[ensures((result != $type::BITS) == (self >> ($type::BITS - result - 1u32) == $one))]
+                    #[ensures((result == $type::BITS) == (self == $zero))]
```
jhjourdan:
```suggestion
                    #[ensures((result != $type::BITS) == (self >> ($type::BITS - result - 1u32) == $one))]
                    #[ensures((result == $type::BITS) == (self == $zero))]
```

This clause is redundant (though it's unclear whether SMT solvers will be more comfortable with suc or such clause).

creusot-std/src/std/num.rs
```diff
@@ -202,6 +202,30 @@ macro_rules! spec_unsized {
...
+                    #[check(ghost)]
+                    #[ensures(self >> $type::BITS - result == $zero)]
+                    #[ensures((result != $type::BITS) == (self >> ($type::BITS - result - 1u32) == $one))]
+                    #[ensures((result == $type::BITS) == (self == $zero))]
+                    fn leading_zeros(self) -> u32;
+
+                    #[check(ghost)]
+                    #[ensures(self << $type::BITS - result == $zero)]
+                    #[ensures((result != $type::BITS) == ((self >> $type::BITS) & $one == $one))]
+                    #[ensures((result == $type::BITS) == (self == $zero))]
```
jhjourdan:
```suggestion
                    #[ensures((result != $type::BITS) == (self << ($type::BITS - result - 1u32) == $one << ($type::BITS - 1u32))]
                    #[ensures((result == $type::BITS) == (self == $zero))]
```

I see no reason we would have a different spec for both functions.

creusot-std/src/std/num.rs
```diff
@@ -202,6 +202,30 @@ macro_rules! spec_unsized {
...
+                    fn trailing_zeros(self) -> u32;
+
+                    #[check(ghost)]
+                    #[ensures(!self >> $type::BITS - result == $zero)]
+                    #[ensures((result != $type::BITS) == (!self >> ($type::BITS - result - 1u32) == $zero))]
+                    #[ensures((result == $type::BITS) == (self == $type::MAX))]
```
jhjourdan:
```suggestion
                    #[ensures((result != $type::BITS) == (!self >> ($type::BITS - result - 1u32) == $zero))]
                    #[ensures((result == $type::BITS) == (self == $type::MAX))]
```

creusot-std/src/std/num.rs
```diff
@@ -202,6 +202,30 @@ macro_rules! spec_unsized {
...
+                    fn leading_ones(self) -> u32;
+
+                    #[check(ghost)]
+                    #[ensures(!self << $type::BITS - result == $zero)]
+                    #[ensures((result != $type::BITS) == ((!self >> $type::BITS) & $one == $one))]
+                    #[ensures((result == $type::BITS) == (self == $type::MAX))]
```
jhjourdan:
```suggestion
                    #[ensures((result == $type::BITS) == (!self << ($type::BITS - result - 1u32) == $one << ($type::BITS - 1u32)))]
                    #[ensures((result == $type::BITS) == (self == $type::MAX))]
```

jhjourdan:
Suggestion: have logic functions for `trailing/leading_one/zeros`, specified the same way, and use them to specify the program function.

Lysxia:
@mcoulmance the main todo is to fix the translation of `<<` in logic to [`lsl_bv : t -> t -> t`](https://www.why3.org/stdlib/bv.html#lsl_bv_518) instead of `lsl : t -> int -> t`. The challenge is that `<<` allows arguments of different types, so we need to cast the second argument accordingly. This also means deciding of a way to handle overflows, which we didn't need before when the second argument was cast to `int`.

creusot/src/backend/term.rs
```diff
@@ -103,6 +103,22 @@ struct Lower<'a, 'tcx, N: Namer<'tcx>> {
     weakdep: bool,
 }
 
+pub fn recast_int<'tcx, N>(namer: &N, ty1: &TyKind, ty2: &TyKind, arg: why3::Exp) -> why3::Exp
+where
+    N: Namer<'tcx>,
+{
```
Lysxia, line 114: Only do this cast if `ty1 != ty2`.

Lysxia:
This looks good to me. Thanks @mcoulmance!

PR #1843, opened 2025-12-01
Add Nix support

lafeychine:
This MR adds a reproducible build schema for solvers (Alt-Ergo, CVC4, CVC5, Z3) and dependencies (why3, why3find).

Thanks to that, a user can now use the following commands :
 - `nix develop`: will enter inside a sub-shell, where a developer can compile and test Creusot;
 - `nix shell`: will enter inside a sub-shell, where an end-user can run Creusot (`cargo creusot new ...`, `cargo creusot prove`).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -70,11 +70,16 @@ jobs:
 
   build-nix:
     runs-on: ubuntu-latest
+    permissions:
+      actions: write # Allow to purge the cache
     steps:
     - uses: actions/checkout@v5
-    - uses: DeterminateSystems/nix-installer-action@main
-    - uses: DeterminateSystems/magic-nix-cache-action@main
-    - uses: DeterminateSystems/flake-checker-action@main
+    - uses: nixbuild/nix-quick-install-action@v34
+    - uses: nix-community/cache-nix-action@v6
+      with:
+        primary-key: nix-${{ runner.os }}-${{ hashFiles('**/*.nix', '**/flake.lock') }}
+        gc-max-store-size: 10G
+        purge: true
     - run: nix shell --impure --command bash -c "cd .. && cargo creusot new test-project && cd test-project && cargo creusot prove"
       env:
         NIXPKGS_ALLOW_UNFREE: 1


creusot-install/why3find.json
```diff
@@ -1,10 +1,10 @@
 {
   "fast": 0.2,
-  "time": 1.0,
-  "depth": 4,
-  "packages": [ "creusot" ],
-  "provers": [ "alt-ergo", "z3", "cvc4", "cvc5" ],
-  "tactics": [ "split_vc" ],
+  "time": 1,
+  "depth": 10,
+  "packages": [],
```
jhjourdan: I thought we needed to keep `creusot` here.

nix/z3.nix
jhjourdan, line 1: Is Creusot the right place for these files? Isn't there a repo of Nix packages where they could be hosted?

why3find.json
```diff
@@ -1,10 +1,16 @@
 {
   "fast": 0.2,
-  "time": 2,
+  "time": 1,
   "depth": 10,
```
jhjourdan: I now remember that I changed this value when the mechanics of `compute_specified` did not work well. Now that it seems to work well, we could probably go back to 4, which would reduce the time spent in failed proof searches.

jhjourdan (commented):
Cc @Halbaroth . Could you please review this briefly?

Generally, this looks good to me, but I am really incompetent for reviewing this. Anyway, I am ready to merge this. (Perhaps @Lysxia could also have a look if he wants.)

Halbaroth:
I'm not a Nix expert at all and Vincent seems to know Nix stuff very well. I will test it as soon as this PR is ready.

jhjourdan:
I think it ***is*** ready. The details in the why3find.json files are somewhat unrelated.

lafeychine:
The PR is indeed ready.

You may need to delete your `~/.config/creusot/` folder, in order for Creusot to set the correct solver path

nix/z3.nix
lafeychine: At the end, we could hope this folder to be upstreamed in the following respective repositories :
 - `nixpkgs`: for related package builds, and builds options (e.g. `CVC5` with `CoCoALib` for example);
 - `wh3find` and `cvc5`: for manual patches of specific features we need.

Currently, as we also wish to find a solution for packaging Cresuot's dependencies, we can make some attempts locally (i.e. on this folder) to modify these build recipes to achieve this goal

Halbaroth:
Sadly, I got a 404 error because the frama-c gitlab is still down: https://git.frama-c.com/pub/why3find

Lysxia:
I put a mirror up at https://gitlab.com/lysxia/why3find if you want to test stuff.

.github/workflows/rust.yml
```diff
@@ -67,6 +67,21 @@ jobs:
...
+    - uses: DeterminateSystems/nix-installer-action@main
+    - uses: DeterminateSystems/magic-nix-cache-action@main
+    - uses: DeterminateSystems/flake-checker-action@main
+    - run: |
+        nix shell --impure -c cargo creusot new test-project
+        cd test-project
+        nix shell --impure -c cargo creusot prove
+      env:
+        NIXPKGS_ALLOW_UNFREE: 1
+
```
jhjourdan, line 86: This job seems to be quite slow (and slower than any other). Do you think it would be possible to cache something?

jhjourdan: Apparently `magic-nix-cache-action@main` should do the job. Let's merge and see.

jhjourdan:
Thanks!

PR #1844, opened 2025-12-01
Operations on `AtomicI32`

dianegolfouse:


creusot-contracts/src/ghost/atomic_invariant.rs
```diff
@@ -0,0 +1,90 @@
...
+//! Atomic invariants are used to specify concurrent operations.
+
+use crate::{ghost::FnGhost, prelude::*};
+
+/// Creusot wrapper around [`std::sync::atomic::AtomicI32`]
+pub struct AtomicI32(::std::sync::atomic::AtomicI32);
+
+/// Does a single atomic operation, in ghost
+#[opaque]
+pub struct Committer {}
```
jhjourdan:
```suggestion
pub struct Committer;
```

creusot-contracts/src/ghost/atomic_invariant.rs
```diff
@@ -0,0 +1,90 @@
...
+
+/// Ownership for a [`AtomicI32`].
+#[opaque]
+pub struct AtomicI32Own {}
```
jhjourdan: Idem.

creusot-contracts/src/ghost/atomic_invariant.rs
```diff
@@ -0,0 +1,90 @@
...
+
+impl AtomicI32Own {
+    /// Which atomic does this holds the ownership for
+    #[logic(opaque)]
+    pub fn id(self) -> AtomicI32 {
```
jhjourdan:
```suggestion
    pub fn atomic(self) -> AtomicI32 {
```

I agree with the idea of not using `Id` as the return type here, but then the name `atomic` is perhaps more appropriate. Open for debate.

Btw, why don't we do the same for `PermCell`? That would be cleaner.

creusot-contracts/src/ghost/atomic_invariant.rs
```diff
@@ -0,0 +1,90 @@
...
+        dead
+    }
+
+    /// Value currently contained in the atomic
+    #[logic(opaque)]
+    pub fn value(self) -> i32 {
```
jhjourdan:
```suggestion
    pub fn val(self) -> i32 {
```

Coherence with e.g., `PermCell`.

creusot-contracts/src/ghost/atomic_invariant.rs
```diff
@@ -0,0 +1,90 @@
...
+    ///
+    /// This is used so that we can only use the committer with the right [`AtomicOwn`].
+    #[logic(opaque)]
+    pub fn id(self) -> AtomicI32 {
+        dead
+    }
+
+    /// Value that will be written to the [`AtomicOwn`]
+    #[logic(opaque)]
+    pub fn final_value(self) -> i32 {
```
jhjourdan: In the case of atomic updates, this should rather be a relation between the old and the new value.

creusot-contracts/src/ghost/atomic_invariant.rs
```diff
@@ -0,0 +1,90 @@
...
+    /// This does the write on the atomic in ghost code, and can only be called once.
+    #[requires(!self.shot())]
+    #[requires(self.id() == own.id())]
+    #[ensures((^self).shot())]
+    #[ensures((^own).id() == (*own).id())]
+    #[ensures((^own).value() == self.final_value())]
+    #[check(ghost)]
+    #[trusted]
+    #[allow(unused_variables)]
+    pub fn shoot(&mut self, own: &mut AtomicI32Own) {
```
jhjourdan: I propose to add post-conditions stating that the value of the committer has been updated.

The reason is that some updates return the updated value, and therefore we need to specify this return value using the final state of the committer.

creusot-contracts/src/ghost/atomic_invariant.rs
```diff
@@ -0,0 +1,90 @@
...
+    }
+}
+
+impl AtomicI32 {
+    /// Wrapper for [`std::sync::atomic::AtomicI32::store`].
+    ///
+    /// The store is always sequentially consistent.
+    #[requires(forall<c: &mut Committer> !c.shot() ==> c.id() == *self ==> c.final_value() == val ==>
+        f.precondition((c,)) &&
+        (forall<r> f.postcondition_once((c,), r) ==> c.shot())
```
jhjourdan:
```suggestion
        (forall<r> f.postcondition_once((c,), r) ==> (^c).shot())
```

creusot-std/src/ghost/invariant.rs
```diff
@@ -98,7 +98,8 @@ pub struct Namespace(());
 /// To help passing it into functions, it may be [reborrowed](Self::reborrow),
 /// similarly to a normal borrow.
 #[opaque]
-pub struct Tokens<'a>(PhantomData<&'a ()>);
+// `*mut ()` so that Tokens are neither Send nor Sync
+pub struct Tokens<'a>(PhantomData<&'a ()>, PhantomData<*mut ()>);
```
jhjourdan, line 129: I guess you did use a negative impl because it requires an unstable feature?

creusot-contracts/src/std/thread.rs
```diff
@@ -0,0 +1,47 @@
...
+    }
+}
+
+extern_spec! {
+    mod std {
+        mod thread {
+            impl<T> JoinHandle<T> {
+                #[ensures(match result {
+                    Ok(res) => self.valid_result(res),
+                    Err(_) => true,
```
jhjourdan: That's frustrating: on the one hand, we would like to use `Err(_) => false,` here, because we have proved that the new thread should not panic and because we would like to be able to `unwrap` the result.

On the other hand, this is unsound, because the `Err` case can actually happen in the case of stack overflow or OOM.

I suggest we keep this extern spec as is, and add (via an extension trait) a version that unwraps the result immediately. This will panic if the spawned thread panicked, but this "garbage in, garbage out", so I think this is fine.

jhjourdan (commented):
Plenty of other atomic types and operations are missing, we should use a macro to generate specs for all of them.

Also, missing : at least one test.

Naming: what about renamig `LocalInvariant` -> `ThreadInvariant` (or something else like `NonAtomicInvariant` : I simply don't like `LocalInvariant`).

While I understand why you used a wrapper around `AtomicI32` (and I somehow agree), I disagree with the idea of moving them to `ghost::atomic_invariant`. This has nothing to do with invariants, really. So we should either add a dedicated module or move this to `std::sync::atomic` (even though this is arguable because this is really a different type).

creusot-std/src/ghost/invariant.rs
dianegolfouse: I didn't even try, but yes it does

creusot-contracts/src/ghost/atomic_invariant.rs
dianegolfouse: > I agree with the idea of not using `Id` as the return type here, but then the name `atomic` is perhaps more appropriate.

I also thought about something like `associated_with`, that describes what this logical data "means" (and this can be the same name for e.g. `PermCell`)

creusot-std/src/ghost/invariant.rs
```diff
@@ -171,9 +180,8 @@ impl Tokens<'_> {
     #[ensures(result.0.contains(*ns))]
     #[ensures(result.1.namespaces() == self.namespaces().remove(*ns))]
     #[check(ghost)]
-    #[allow(unused_variables)]
     pub fn split<'a>(&'a mut self, ns: Snapshot<Namespace>) -> (Tokens<'a>, Tokens<'a>) {
```
jhjourdan, line 202: Could we make `Namespace` `Copy` and `Plain`, and pass `Namespace` instead of `Snapshot<Namespace>` for every ghost function?

jhjourdan: @arnaudgolfouse As far as I can remember, `Namespace` is pure data, it does not convey any ownership, so this should be safe, right?

creusot-contracts/src/ghost/invariant.rs
```diff
@@ -190,35 +198,99 @@ impl View for Tokens<'_> {
...
     type Public;
 
     #[logic(prophetic)]
     fn protocol(self, data: Self::Public) -> bool;
 }
 
-/// Define method call syntax for [`LocalInvariant::open`].
-pub trait LocalInvariantExt<'a> {
+#[opaque]
+pub struct AtomicInvariant<T: Protocol>(PhantomData<T>);
```
jhjourdan:
```suggestion
pub struct AtomicInvariant<T>(PhantomData<T>);
```

That should simplify some `impl` declarations.

creusot-contracts/src/ghost/invariant.rs
```diff
@@ -190,35 +198,99 @@ impl View for Tokens<'_> {
...
+
+unsafe impl<T: Protocol + Send> Sync for AtomicInvariant<T> {}
```
jhjourdan:
```suggestion
unsafe impl<T: Send> Sync for AtomicInvariant<T> {}
```

creusot-contracts/src/ghost/invariant.rs
```diff
@@ -190,35 +198,99 @@ impl View for Tokens<'_> {
...
+    ///
+    /// This will call the closure `f` with the inner data. You must restore the
+    /// contained [`Protocol`] before returning from the closure.
+    #[trusted]
+    #[requires(tokens.contains(self.namespace()))]
+    #[requires(forall<t: &mut T> t.protocol(self.public()) && inv(t) ==>
+        f.precondition((t,)) &&
+        // f must restore the invariant
+        (forall<res: A> f.postcondition_once((t,), res) ==> (^t).protocol(self.public())))]
+    #[ensures(exists<t: &mut T> t.protocol(self.public()) && f.postcondition_once((t,), result))]
```
jhjourdan:
```suggestion
    #[ensures(exists<t: &mut T> inv(t) && t.protocol(self.public()) && f.postcondition_once((t,), result))]
```

creusot-contracts/src/ghost/invariant.rs
```diff
@@ -190,35 +198,99 @@ impl View for Tokens<'_> {
...
+    #[check(ghost)]
+    pub fn open<A>(&self, tokens: Tokens, f: impl FnGhost + for<'a> FnOnce(&'a mut T) -> A) -> A {
+        panic!("Should not be called outside ghost code")
+    }
+
+    /// Open the invariant to get the data stored inside, immutably.
+    /// This allows reentrant access to the invariant.
```
jhjourdan: You could perhaps say that this variant is here for coherence with the non-atomic case, but should not really be useful in practice, because atomicity prevents reentrancy anyways (well, except from reentrancy within the same ghost block, but this is kind of artificial).

creusot-std/src/ghost/invariant.rs
```diff
@@ -190,35 +198,99 @@ impl View for Tokens<'_> {
...
+    /// Get the 'public' part of this invariant.
+    #[logic(opaque)]
+    pub fn public(self) -> T::Public {
+        dead
+    }
+
+    /// Open the invariant to get the data stored inside.
+    ///
+    /// This will call the closure `f` with the inner data. You must restore the
+    /// contained [`Protocol`] before returning from the closure.
```
jhjourdan, line 281:
```suggestion
    /// contained [`Protocol`] before returning from the closure.
    
    /// Note: this function can only be called from ghost code, because atomic invariants are always wrapped in `Ghost`. This guarantees atomicity.
```

creusot-contracts/src/ghost/invariant.rs
jhjourdan:
```suggestion
    #[ensures(exists<t: Ghost<&mut T>> inv(t) && t.protocol(this.public()) && f.postcondition_once((t,), result))]
```

creusot-contracts/src/std/sync.rs
```diff
@@ -55,3 +58,76 @@ impl<T: DeepModel> DeepModel for Arc<T> {
...
+
+/// Creusot wrapper around [`std::sync::atomic::AtomicI32`]
+pub struct AtomicI32(::std::sync::atomic::AtomicI32);
+
+impl AtomicI32 {
+    /// Wrapper for [`std::sync::atomic::AtomicI32::fetch_add`].
+    ///
+    /// The fetch and the store are always sequentially consistent.
+    // TODO: Handle overflow wrapping
+    #[requires(forall<c: &mut Committer> !c.shot() ==> c.tied() == *self ==> (^c).value()@ == val@ + (*c).value()@ ==>
```
jhjourdan: I think this is unsound, because it says in the precondition of `f` that `(^c).value()@ == val@ + (*c).value()@`. So, if you resolve `c` at the very beginning of `f`, you get `^c == *c`, which is contradictory, and then you can do whatever in th closure (including not committing).

jhjourdan: Instead, the committer should have two logic functions: `old_value` and `new_value` (or maybe one logic function that returns a mutable borrow, but this is artificial, because this mutable borrow never actually exists).

creusot-contracts/src/std/sync.rs
```diff
@@ -55,3 +58,76 @@ impl<T: DeepModel> DeepModel for Arc<T> {
...
+        F: FnGhost + FnOnce(&'a mut Committer) -> A,
+    {
+        self.0.fetch_add(val, std::sync::atomic::Ordering::SeqCst);
+        Ghost::conjure()
+    }
+
+    /// Wrapper for [`std::sync::atomic::AtomicI32::store`].
+    ///
+    /// The store is always sequentially consistent.
+    #[requires(forall<c: &mut Committer> !c.shot() ==> c.tied() == *self ==> (^c).value() == val ==>
```
jhjourdan: Same here.

creusot-contracts/src/std/sync.rs
```diff
@@ -55,3 +58,76 @@ impl<T: DeepModel> DeepModel for Arc<T> {
...
 }
+
+/// Creusot wrapper around [`std::sync::atomic::AtomicI32`]
+pub struct AtomicI32(::std::sync::atomic::AtomicI32);
+
+impl AtomicI32 {
+    /// Wrapper for [`std::sync::atomic::AtomicI32::fetch_add`].
+    ///
+    /// The fetch and the store are always sequentially consistent.
+    // TODO: Handle overflow wrapping
```
jhjourdan: That should be easy: simply remove the `@`. The semantics of `+` for bounded integers in the logic is wrapping.

tests/should_succeed/atomics/parallel_add.rs
```diff
@@ -0,0 +1,127 @@
...
+            let inv = &inv;
+            let true_ = Some(Excl(true));
+
+            atomic.fetch_add(
+                1,
+                ghost! {
+                    #[check(ghost)]
+                    |c: &mut Committer| {
+                        inv.open(tokens.into_inner(), #[check(ghost)] move |inv: &mut ParallelAddAtomicInv| {
+                            proof_assert!(!(frag1@.frag().unwrap_logic().0));
```
jhjourdan: Is this necessary? Does the proof fails without that?

tests/should_succeed/atomics/parallel_add.rs
```diff
@@ -0,0 +1,127 @@
...
+        }),
+        snapshot!((atomic, frag1.id(), frag2.id())),
+        snapshot!(PARALLEL_ADD()),
+    );
+
+    thread::scope(|s| {
+        let mut frag1 = frag1.borrow_mut();
+        let mut frag2 = frag2.borrow_mut();
+
+        let t1 = s.spawn(|tokens| {
```
jhjourdan:
```suggestion
        let t1 = s.spawn(move |tokens| {
```

Otherwise, ghosts are captured by references, and the closure contains unnecessary pointers at runtime (yes, this is unfortunate).

tests/should_succeed/atomics/parallel_add.rs
```diff
@@ -0,0 +1,127 @@
...
+            let inv = &inv;
+            let true_ = Some(Excl(true));
+
+            atomic.fetch_add(
+                1,
+                ghost! {
+                    #[check(ghost)]
+                    |c: &mut Committer| {
+                        inv.open(tokens.into_inner(), #[check(ghost)] move |inv: &mut ParallelAddAtomicInv| {
```
jhjourdan: Why is the `move` necessary here?

tests/should_succeed/atomics/parallel_add.rs
```diff
@@ -0,0 +1,127 @@
...
+                            proof_assert!(!(frag1@.frag().unwrap_logic().0));
+                            inv.auth1.join_in(frag1.take());
+                            inv.auth1.update(ViewUpdate(snapshot!(|()| (true_, true_))));
```
jhjourdan: Could you try `raw_update` instead here? It is probably just simpler for this specific case.

tests/should_succeed/atomics/parallel_add.rs
```diff
@@ -0,0 +1,127 @@
...
+        let _ = t2.join_unwrap();
+    });
+
+    ghost! {
+        inv.open(tokens.into_inner(), #[check(ghost)] move |inv: &mut ParallelAddAtomicInv| {
+            inv.auth1.join_in(frag1.take());
+            inv.auth2.join_in(frag2.take());
+
+            proof_assert!(inv.auth1@.auth().unwrap_logic().unwrap_logic().0);
+            proof_assert!(inv.auth2@.auth().unwrap_logic().unwrap_logic().0);
```
jhjourdan: Are these necessary?

creusot-std/src/ghost/invariant.rs
dianegolfouse: Yes, this should not be an issue.

jhjourdan: I tried this, and there is actually an issue: the namespace constants are logic functions, which cannot be used in ghost code. So if we replace `Snapshot<Namespace>` with `Namespace`, we have to do the `into_ghost` dance, which is not pleasant.

The real solution is to provide a ghost version of these constants. But let's wait for #1346 so that we can share the names.

jhjourdan:
If you agree, I think the current situation is acceptable.

Lysxia (approved):

PR #1846, opened 2025-12-02
Release 0.8.0

Lysxia:
It would be nice if why3find came back online soon...


xldenis:
>It would be nice if why3find came back online soon...

should we host a mirror on our github?

Lysxia:
My plan is to wait and hope it gets resolved by next week. So we hold off from merging anything till then.

PR #1849, opened 2025-12-03
Add `Ghost::into_inner_unsized`

Lysxia:


Lysxia:
I'm not sure I want to merge this. After making `PtrOwn<T>` unsized, I would want to convert `Ghost<PtrOwn<T>>` to `Ghost<Box<PtrOwn<T>>` (which is what this PR enables) but also the other way, which seems to require yet another intrinsic, and is overall quite annoying. Instead, I'm experimenting with just replacing all occurrences of `Ghost<PtrOwn<T>>` with `Ghost<Box<PtrOwn<T>>>`, which makes this change unnecessary.

jhjourdan:
Ok, understood.

That being said, the but with the `build` CI target remains. And in particuler, I discovered that there *is* an option `--creusot-contracts` to `cargo creusot new`, which is completely ignored by the code.

PR #1850, opened 2025-12-04
Allow setting path to creusot-contracts in cargo creusot new

Lysxia:
CI uses the `--creusot-contracts` option which was accidentally left unimplemented. This implements it.

### DIFF
--- a/guide/src/command_line_interface.md
+++ b/guide/src/command_line_interface.md
@@ -109,7 +109,7 @@ which ensures that the Coma artifacts are always up to date.
 ### `new`
 
 ```
-cargo creusot new <NAME> [--main]
+cargo creusot new <NAME> [--main] [--creusot-contracts <PATH>]
 ```
 
 Create or update package named `<NAME>`.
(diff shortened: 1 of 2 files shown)


jhjourdan:
@Lysxia @lafeychine

CI is broken on this PR for Nix, and this does not seem to be related to Why3find, so I guess this is a bug in our code.

lafeychine:
We found the bug with @Lysxia, it seems not to be related with Nix, and more related with changes made by this PR

If one clones Creusot in `/<DIR>/creusot`, `cargo creusot new` will add a patch that points `creusot_contracts` to `/<DIR>/creusot/creusot-contracts`.
However, this introduces issues when the directory is moved or deleted, e.g. `/tmp`.

Therefore, the patch should not be done implicitly or point to a hardcoded path.

jhjourdan:
So we hit a Nix bug one week after installing it in or CI? Ironic!

jhjourdan:
Ah, no, I understand, that's not a Nix bug.

Lysxia:
Oops... Merged too quickly (because the branch requirement for `nix-build` was missing). We also need to copy `creusot-contracts-proc` and `pearite-syn` (Done in #1852)

PR #1854, opened 2025-12-12
Loop invariants: infer type invariants

### LINKED ISSUE #1122: Static analysis for type invariants

jhjourdan:
Fixes #1122 .

This also does various refactorings of the FMIR AST, and of the other static analyses (this probablly fied some soundness bugs in the temp propagation analysis).

Another improvement is the replacement of *assertions* of the type invariant when creating and resolving mutable borrows by *checkings*, that do not polute the proof context. It used to contain a lot of redundants hypotheses, which all go away thanks to that. It seems to (slightly) reduce the size of proof.json files, and makes the contexts much more readable.


xldenis:
Would it be realistic to also generate the `produced` invariant of for loops during this analysis?

jhjourdan:
I don't know how to do this in a satisfying way. The `produced` variable is a snapshot, and snapshots have no type invariant. We could constrain them to enforce the type invariant, but I don't see a rationnale for that, and IMO I find it rather confusing.

PR #1855, opened 2025-12-12
Ensure compatibility with #![no_std] projects

### LINKED ISSUE #1607: Supporting no-std crates
The hashbrown crate doesn't require std but we cannot import `creusot_contracts` without std. My workaround consists in removing `#![no_std]`.

mcoulmance:
To ensure compatibility with no_std, we need to enforce the following discipline in creusot-contracts:
- Every item that is declared in core should be imported as such, for example:
  ```rust
  use std::cmp::Ordering       // <-- incorrect
  use core::cmp::Ordering      // <-- correct
  ```
- A new Cargo feature "std" has been added to this crate, every std-dependent part of the standard library we want
  to deal with in creusot-contracts should be compiled with a #[cfg(feature = "std")] so it can be disabled in no_std mode.
  For example: 
  ```rust 
     #[cfg(feature = "std")]
     impl View for std::time::Instant
    ...
  ``` 

This is still a WIP, if you find any corner case I may have missed please notify me

Fixes #1607.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -231,7 +231,7 @@ jobs:
         cd test-project-nostd
         echo $TLC > rust-toolchain
         cargo fmt --check
-        cargo build 
+        cargo build
         cargo build --target thumbv7em-none-eabi -Zbuild-std=core,alloc
         cargo creusot prove
         cargo creusot prove -- --target thumbv7em-none-eabi -Zbuild-std=core,alloc


creusot-contracts/src/std/string.rs
```diff
@@ -1,5 +1,6 @@
 use crate::prelude::*;
-use std::ops::Deref;
+#[allow(unused_imports)]
```
Lysxia:
```suggestion
#[cfg(feature = "std")]
```

cargo-creusot/src/new.rs
```diff
@@ -148,7 +181,11 @@ pub fn create_project(name: String, args: NewInitArgs) -> Result<()> {
         if args.main {
             write("src/main.rs", &bin_template(&name));
         }
-        write("src/lib.rs", LIB_TEMPLATE);
+        if args.no_std {
+            write("src/lib.rs", LIB_TEMPLATE_NOSTD);
```
Lysxia:
```suggestion
            write("src/lib.rs", format!("#![no_std]\n{LIB_TEMPLATE}"));
```

creusot-contracts/src/ghost/ptr_own.rs
```diff
@@ -229,6 +232,7 @@ impl<T: ?Sized> PtrOwn<T> {
     ///
     /// Creusot will check that all calls to this function are indeed safe: see the
     /// [type documentation](PtrOwn).
+    #[cfg(feature = "std")] // TODO: Remove std dependency ?
```
jhjourdan: This function needs to deallocate memory, so it clearly depends on std. Why do you want to break this dep?

creusot-contracts/src/logic/fset.rs
jhjourdan: The no-std mode should not remove features of ghost code. So, as a general rule, there should not be any cfg-gated ghost function.

In this specific case, I think this function should just go away. We should not store unsized values in `FSet`, they should be wrapped. This seems to be legacy from the time where we allowed that.

creusot-contracts/src/std/panicking.rs
```diff
@@ -1,7 +1,8 @@
 use crate::prelude::*;
 #[cfg(feature = "nightly")]
 use core::panicking::*;
-use std::{any::Any, fmt};
+#[allow(unused_imports)]
```
jhjourdan:
```suggestion
#[cfg(feature = "std")]
```

creusot-contracts/src/std/slice.rs
```diff
@@ -2,8 +2,9 @@
 use crate::resolve::structural_resolve;
 use crate::{ghost::PtrOwn, invariant::*, logic::ops::IndexLogic, prelude::*};
 #[cfg(feature = "nightly")]
-use std::alloc::Allocator;
-use std::{
+#[allow(unused_imports)]
```
jhjourdan:
```suggestion
#[cfg(all(feature= "nightly", feature = "std"))]
```

creusot-contracts/src/peano.rs
```diff
@@ -27,18 +27,33 @@
 
 use crate::{
     logic::ord::ord_laws_impl,
-    prelude::{Clone, Default, *},
+    //prelude::{Clone, Default, *},
```
jhjourdan: This should be kept in order to shadow STD's definitions.

creusot-contracts/src/peano.rs
```diff
@@ -27,18 +27,33 @@
...
+    prelude::*,
 };
-use std::cmp::Ordering;
+use core::cmp::Ordering;
 
 /// A peano integer wrapping a 64-bits integer.
 ///
 /// See the [module](crate::peano) explanation.
-#[derive(Clone, Copy, Default, Eq)]
+#[derive(Copy, Eq)] // TODO: derive(Clone, Default) complains when compiling with no_std
```
jhjourdan: What's the problem, exactly?

jhjourdan:
Generally, this looks good! Could you fix the conflicts so that we can see what CI has to say?

It seems to me that the only remaining "surprising" issue is the problem with Clone/Default. Could you elaborate?

creusot-contracts/src/ghost/ptr_own.rs
mcoulmance: This comment came during a discussion I had with @arnaudgolfouse, but I'm not sure I remember what his
point exactly was.
@arnaudgolfouse any idea ?

creusot-contracts/src/peano.rs
mcoulmance: The problem was that the `derive` macro introduced a dependency to `std`, but it seems that it was caused by 
the shadowing issue you mentioned in the previous comment. I just applied a patch to fix it, and it now works !

mcoulmance:
The only remaining issue here is the CI. It seems that the no_std targets are a bit trickier to use than the other ones. I'm investigating on that matter.

jhjourdan:
Could you press the "Resolve conversation" when you address a comment and there is nothing more to discuss about it?

creusot-contracts/src/logic/seq.rs
```diff
@@ -97,7 +97,7 @@ impl<T> Seq<T> {
     /// ```
     #[logic]
     #[builtin("seq.Seq.get")]
-    pub fn index_logic_unsized(self, ix: Int) -> Box<T> {
+    pub fn index_logic_unsized(self, ix: Int) -> &'static T {
```
Lysxia:
```suggestion
    pub fn index_logic_unsized<'a>(self, ix: Int) -> &'a T {
```

PR #1858, opened 2025-12-15
Support ranges in IndexLogic

Lysxia:
Provides a shorthand `s[start..end]` for `s.subsequence(start, end)`, including all the other variants of ranges.


jhjourdan:
Thanks, very handy indeed.

PR #1859, opened 2025-12-15
Refactor permission types

jhjourdan:
This PR contains two commits.

I hope the first one is not controversial. It makes the API of `PermCell` and `PtrOwn` simpler and more coherent. HEre is the commit message:

> Refactor PermCell and PtrOwn API.
> PermCell:
> - Get rid of PermCell Id. Directly use the PermCell value instead
> - Rename the logical function id into tied
> 
> PtrOwn:
> - Rename the logical function ptr into tied, for coherence.

If you want, you can bikeshed on names. `tied` is the best name we could come up with @lafeychine, but we were not entirely satsfied.

In the second commit, I go even further, by foctoring out the permission types `PermCellOwn` and `PtrOwn` in to a new type `Perm`. I don't know if it's really an improvement: it's fewer types to remember (and even fewer when remembering that we are going to add even more permissions with atomics), and it makes sure that we use the same convention everywhere. But OTOH, there is some friction added by this: the `tied` logic function returns a borrow even for the pointer case, where we know that the container type will always be sized. Also, it requires specifying the `disjoint_lemma` function by going through a logic function, which is not particularly elegant either.

What do you think, @Lysxia , @lafeychine , @arnaudgolfouse ?


creusot-contracts/src/ghost/perm.rs
```diff
@@ -0,0 +1,99 @@
...
+
+/// Token that represents the ownership of the contents of a container object. The container is
+/// either an interrior mutable type (e.g., `Perm` or atomic types) or a raw pointer.
+///
+/// A `Perm` only exists in the ghost world, and it must be used in conjunction with its container
+/// in order to read or write the value.
+///
+/// # Pointer permissions
+///
+/// A particular case of permissions is the case of permissions for raw pointers (i.e., `C` is
```
Lysxia, line 29:
```suggestion
/// A particular case of permissions is the case of permissions for raw pointers (i.e., `C`
```

creusot-contracts/src/ghost/perm.rs
```diff
@@ -0,0 +1,99 @@
...
+/// Specifically, creating too many pointers contradicts the [`Perm::disjoint_lemma`],
+/// and allocating too large objects contradicts the [`Perm::invariant`] that
+/// allocations have size at most `isize::MAX`.
+#[opaque]
+pub struct Perm<C: ?Sized + Container>(#[allow(unused)] PhantomData<C::Value>);
+
+impl<C: ?Sized + Container> Perm<C> {
+    /// Returns the underlying container that is managed by this permission.
+    #[logic(opaque)]
+    pub fn tied<'a>(self) -> &'a C {
```
Lysxia: How about using `&'static C` to avoid quantifying over a lifetime?

creusot-contracts/src/std/slice.rs
```diff
@@ -82,10 +82,10 @@ pub trait SliceExt<T> {
     fn to_ref_seq(&self) -> Seq<&T>;
 
     #[check(terminates)]
-    fn as_ptr_own(&self) -> (*const T, Ghost<&PtrOwn<[T]>>);
+    fn as_ptr_own(&self) -> (*const T, Ghost<&Perm<*const [T]>>);
```
Lysxia: Replace all the `own` suffixes with `perm`?

creusot-contracts/src/ghost/perm.rs
```diff
@@ -0,0 +1,99 @@
...
+/// 2. allocating too large objects.
+///
+/// Note that we already can't guard against these issues in program code.
+/// But preventing them in ghost code is even more imperative to ensure soundness.
+///
+/// Specifically, creating too many pointers contradicts the [`Perm::disjoint_lemma`],
+/// and allocating too large objects contradicts the [`Perm::invariant`] that
+/// allocations have size at most `isize::MAX`.
+#[opaque]
+pub struct Perm<C: ?Sized + Container>(#[allow(unused)] PhantomData<C::Value>);
```
Lysxia: Make it unsized? (Note that the soundness issue with this is not limited to the PR for slices: `from_mut` in `ptr.rs` is currently affected.)

```suggestion
pub struct Perm<C: ?Sized + Container>(#[allow(unused)] [PhantomData<C::Value>]);
```

Lysxia:
This is a great idea.

`tied` sounds a bit too generic. A better word may be `ward` (as in "a protected person or object or place").

creusot-contracts/src/ghost/perm.rs
jhjourdan: In general, that's a bad idea tu use `&'static` everywhere, because this requires that the types `C` and `C::Values` outilive `'static`. So I prefer `'a` (plus, it's actually shorter, even with the `<'a>`).

creusot-contracts/src/ghost/perm.rs
jhjourdan: You're right... I actually did that on the first place, and then rolled back because it's so invasive to use `Box<Perm<...>>` everywhere. I thought we could find a way around by enforcing that the ward does not change in the prophecy.

But you're right that `from_mut` is a desperate case. I'm going to make `Perm` unsized...

jhjourdan:
> tied sounds a bit too generic. A better word may be ward (as in "a protected person or object or place").

Indeed, I like it. Renamed.

jhjourdan:
So, should I merge despite the added friction to our case studies?

Lysxia:
I'm in favour. I like reducing duplication.

jhjourdan:
@arnaudgolfouse : you did not give us any input (this is fine, you are in vacation). I'm going to merge this for now, but we are 100% open to changes suggestions from your end. But for now, we need to proceed because the content of the tutorial depends on whether or not we merge this.

PR #1861, opened 2025-12-17
Add impl FromIteratorSpec<()> for ()

Lysxia:
Like `map_inv` as a generalization of `map`, I wanted a `for_each_inv`. It can be encoded trivially as a combination of `map_inv` and `collect::<()>`.

### DIFF
--- a/tests/creusot-contracts/creusot-contracts.coma
+++ b/tests/creusot-contracts/creusot-contracts.coma
@@ -55088,39 +55088,6 @@ module M_std__iter__zip__impl_IteratorSpec_for_Zip_A_B__produces_trans__refines 
               /\ produces_Zip_A_B b bc c
               /\ (forall result: (). produces_Zip_A_B a (Seq.(++) ab bc) c -> produces_Zip_A_B a (Seq.(++) ab bc) c)
 end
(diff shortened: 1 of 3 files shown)


tests/creusot-contracts/creusot-contracts.coma
```diff
@@ -55088,7 +55088,7 @@ module M_std__iter__zip__impl_IteratorSpec_for_Zip_A_B__produces_trans__refines
               /\ produces_Zip_A_B b bc c
               /\ (forall result: (). produces_Zip_A_B a (Seq.(++) ab bc) c -> produces_Zip_A_B a (Seq.(++) ab bc) c)
 end
-module M_std__iter__impl_IteratorSpec_for_refmut_I__produces_refl__refines (* <&mut I as std::iter::IteratorSpec> *)
+module M_std__iter__impl_IteratorSpec_for_refmut_I__produces_trans__refines (* <&mut I as std::iter::IteratorSpec> *)
```
jhjourdan, line 55091: There is something weird heren the order of the modules should not change... Whatever.

jhjourdan:
As you want.

(That being said, I tend to not like using `foreach` where a `for` loop can be used and is generally clearer.)

PR #1864, opened 2026-01-06
Remove why3find install

lafeychine:


jhjourdan:
CI is failing, so there is still somethig that does not work with our way of exporting the prelude in the test suite.

Lysxia:
META file?

lafeychine:
I successfully ran commands in an empty Ubuntu VM, I do not know where the problem comes from

Lysxia:
These are the tests launched with why3. The `-L` option also needs an update with the new location of the creusot package.

https://github.com/creusot-rs/creusot/blob/6130c1898e33de6b90383ef15c4cbed254611ca8/why3tests/tests/why3.rs#L183

https://github.com/creusot-rs/creusot/blob/6130c1898e33de6b90383ef15c4cbed254611ca8/why3tests/tests/why3.rs#L220

lafeychine:
I finally achieved the same failing state as the CI.

Thanks @Lysxia for the pointers, these were indeed the locations to update

PR #1865, opened 2026-01-06
Clean up linked_list example

Lysxia:


Lysxia:
@jhjourdan should we define `Seq::map` with `Seq::create` like the old `seq_map` in `linked_list.rs` that this PR removes?

jhjourdan:
LGTM.

PR #1867, opened 2026-01-07
Fix `check(terminates)` in linked_list example

### LINKED ISSUE #1866: `#[check(terminates)]` crashes in `linked_list` example
Adding `#[check(terminates)]` to any of the methods in `linked_list.rs` crashes with:

```
thread 'rustc' (114333) panicked at creusot/src/translation/traits.rs:383:17:
not implemented: Cannot handle builtin implementation of `std::ptr::Pointee` for `Link<T>`
```

(`Link` is the new name of `Cell` from #1865)

Lysxia:
Close #1866

### DIFF
--- a/tests/should_succeed/linked_list.rs
+++ b/tests/should_succeed/linked_list.rs
@@ -45,11 +45,13 @@ impl<T> View for List<T> {
 }
 
 impl<T> List<T> {
+    #[check(ghost)]
     #[ensures(result@ == Seq::empty())]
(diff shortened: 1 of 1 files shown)


jhjourdan:
Have you checked that the implied bounds are still in the list?

Lysxia:
Yes they are in this case. Let me add a comment to remember that.

jhjourdan:
Good, thanks!

PR #1868, opened 2026-01-07
Rename creusot-contracts => creusot-std.

jhjourdan:
I wanted to do this change for a long time, because "creusot-contracts" means nothing to me.

A problem I did not handle is taking care of crates.io, since I don't manage this. @Lysxia? Unfortunately, it is not possible to "rename" a crate on crates.io, so we will need to crete a new one, and clearly deprecate creusot-contracts.

@xldenis Since this is a rather big change, Iwould be happy to know what you think about this.


jhjourdan:
Of course, @arnaudgolfouse, @Lysxia, @lafeychine, @Armael and @mcoulmance can comment too, but I expect them to be notified by this PR anyway.

xldenis:
I mean it's fine with me but as you noted it does break crates.io for perhaps a bit marginal gain?

jhjourdan:
The point is that the later we do the change, the more complicated it will be. For now, no big external project depend on creusot-contracts, so I would say the "loss" is marginal too.

Lysxia:
This renaming makes sense to me too.

PR #1873, opened 2026-01-09
Add example sum_first_n_iter

Lysxia:
@jhjourdan I'm trying to work out more interesting examples of using closures, but I'm a bit stuck with this one. Could you have a look?

### DIFF
--- a/tests/should_succeed/sum_first_n_iter.coma
+++ b/tests/should_succeed/sum_first_n_iter.coma
@@ -1,0 +1,305 @@
+module M_sum_first_n_iter
+  use creusot.int.UInt64
+  use mach.int.Int
+  use creusot.prelude.MutBorrow
+  use seq.Seq
+  use creusot.prelude.Any
+  use creusot.int.UInt32
+  
+  type t_RangeInclusive_usize
+  
+  predicate inv_RangeInclusive_usize (_1: t_RangeInclusive_usize)
+  
+  function start_log_RangeInclusive_usize (self: t_RangeInclusive_usize) : UInt64.t
+  
+  function end_log_RangeInclusive_usize (self: t_RangeInclusive_usize) : UInt64.t
+  
+  function deep_model_usize [@inline:trivial] (self: UInt64.t) : int = UInt64.t'int self
+  
+  meta "rewrite_def" function deep_model_usize
+  
+  predicate is_empty_log_RangeInclusive_usize (self: t_RangeInclusive_usize)
+  
+  axiom is_empty_log_RangeInclusive_usize_spec:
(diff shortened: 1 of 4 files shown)


jhjourdan:
I don't like the idea of unfolding the specs of  `map_inv`. They are rather big, and may clutter the SMT files if unfolded too aggressively.

xldenis:
I think the problem is this requires the non-linear arithmetic theory no? that is known to instantly kill smt performance esp in presence of quantifiers.

jhjourdan:
No. Apparently, in my investigations, it seems related to the way we handle closed intervals.

Lysxia:
Trying to encapsulate the sum formula led me to this working solution, but it still requires inlining the `preservation` method.

```
extern crate creusot_std;
use creusot_std::prelude::*;

#[logic(open)]
pub fn triangular(n: Int) -> Int {
    pearlite! { n * (n + 1) / 2 }
}

#[requires(triangular(n@) <= u32::MAX@)]
#[ensures(result@ == triangular(n@))]
pub fn sum_first_n_iter(n: usize) -> usize {
    let mut sum = 0;
    proof_assert! { forall<i> 0 <= i ==> triangular(i + 1) == triangular(i) + i + 1 };
    let _ = (1..=n).map_inv(|x, produced| {
        proof_assert! { x@ == produced.len() + 1 };
        proof_assert! { sum@ == triangular(produced.len()) };
        sum += x;
    }).collect::<()>();
    sum
}
```

jhjourdan:
Please bear we me, I'm trying to understand what is happening.

jhjourdan:
Okay, I spent a few hours on this. I have several working alternatives, but nothing really satisfactory. Yours is probably the best. Non-linear arithmetic is definitely one of the reasons things do not work. The several layers of abstractions that are inherent to iterators are another reason. Somehow, when provers see that they need to do a bit of non-linear computations, they put it on a very low priority and prefer looping in doing plenty of other things... And because of the various definitions about iterators, there ARE plenty of other things to do.

About inlining the definitions:
   - For some reason, on my machine, your example works without inlining.
   - Thinking more about it, I think that indeed some of these definitions deserve inlining, because for the client they (almost) only appear as a goal, and inlining them may help Why3 splitting them.
   - Perhaps the one case for which I'm unsure is produces_one.
   - if we do that for map_inv, then we need to do that for map too.

For the tutorial, I don't know : should we use a different example, or should we use your variant?

jhjourdan:
I guess the current version is showable. But it is disappointing, it is so unstable. The proof breaks as soon as I play with the `inline` tags or by moving the assertions up and down.

Lysxia:
For the tutorial I'll use "sum a vector" instead which will look a little bit nicer. But it actually hides some of the same issues, and in particular I still need the same `inline`.

PR #1877, opened 2026-01-11
Better automation for RAs

jhjourdan:


creusot-std/src/logic/ra.rs
```diff
@@ -126,10 +126,37 @@ pub trait RA: Sized {
...
+    /// Specialized version of [`Self::associative`], in the case where `a.op(b) == None`.
+    ///
+    /// By commutativity, it also covers the case where `b.op(c) == None`.
+    #[logic(law)]
+    #[requires(a.op(b) == None)]
+    #[requires(b.op(c) == Some(bc))]
+    #[ensures(a.op(bc) == None)]
+    fn associative_none(a: Self, b: Self, c: Self, bc: Self) {
+        Self::associative(a, b, c);
+    }
```
xldenis, line 146: if this becomes a common pattern it feels like there should be a mechanism to generate specialized forms of laws logical statements from a general one.

PR #1883, opened 2026-01-16
Use a timeout multiplier of 2 when running the nighlty for more stability

jhjourdan:


jhjourdan:
Even using a multiplier of 10 does not help... It seems like provers really have a different behavior on macos.

We need to do something about the nightly. It is *not* satisfying to have a nightly which is failing every night...

Lysxia:
Switching the macos job to use nix sounds like a good idea.

jhjourdan:
I'm going to close this for now, since it does not solve the issue.

PR #1886, opened 2026-01-20
doc: Fix a couple of links

Lysxia:


creusot-std/src/logic/ra/auth.rs
```diff
@@ -16,8 +16,8 @@ use crate::{
 /// - the relation between the two is specified by [`AuthViewRel`]: it asserts that
 ///   `Frag` must always be included in `Auth`
 ///
-/// If this type is directly used as a ghost resource, on should rather use
-/// [`crate::ghost::resource::auth`], which provides convenient wrappers which the provers prefer.
+/// If this type is directly used as a ghost resource, one should rather use
+/// [`crate::ghost::resource::Authority`], which provides convenient wrappers which the provers prefer.
```
Lysxia: @jhjourdan is this correct?

jhjourdan: `crate::ghost::resource::auth` used to be a module, that contained two types: `Authority` and `Fragment`. So both types should be mentioned here.

PR #1892, opened 2026-02-03
ci: Disable Nightly job on MacOS

Lysxia:
It's too broken due to solver differences between Linux and MacOS.

Maybe we could make it a Nix-on-MacOS job but it's not clear that we would gain much from it.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -7,7 +7,7 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest]
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v4


jhjourdan:
Do we have an idea of whether its the fault of one specific prover or all of them?

Lysxia:
Good point. It might be specific to CVC5 (perhaps also CVC4). I'm trying to fix the proofs to use Alt-ergo and Z3 first. There are some holdouts, with a notable one in `examples/iterators/17_filter.coma`, goal `vc_less_than.2.1` can only be solved by CVC5.

Lysxia:
There are still some goals that can only be solved by CVC5 on Linux. How should we deal with them? Ideas:

1. Disable the MacOS job altogether. (This PR.)
2. Disable the few broken tests on MacOS.
3. Improve the proofs to make them pass with other solvers (or still in CVC5 but in a portable way).

I'm inclined to do 2 because it still lets us know that things mostly work on MacOS at least. (And 3 seems like a lot of effort for a fragile result.)

jhjourdan:
I'm afraid neither 2 nor 3 will be stable in the end, because future tests will surely generate new VCs for which Why3find will choose CVC5 and for which MacOS will fail.

Lysxia:
OK let's do this. At least we are still testing that the installation succeeds on MacOS in the main workflow.

PR #1898, opened 2026-02-04
doc: Fix broken link to termination.md in gnome_sort tutorial

ejackson-vp:
The relative link `./termination.md` in `guide/src/tutorial/gnome_sort.md` was incorrect since the file is in the `tutorial/` subdirectory.

Changed the path from `./termination.md` to `../termination.md` to correctly navigate up one directory level to reach `guide/src/termination.md`.


Lysxia:
Nice catch, thanks!

PR #1899, opened 2026-02-05
fix: add assumptions for wildcard patterns on integer match

### LINKED ISSUE #1248: Wildcard patterns on integers are missing assumptions
```rust

pub fn f(x: usize) {
    match x {
        1 => assert!(x == 1),
        _ => assert!(x != 1),
    }
}
```

The `assert(x != 1)` fails because we generate a default branch with no assumptions.

ejackson-vp:
Fixes #1248

## Summary

When matching on integer types (`Int`/`Uint`), the default/wildcard branch now assumes that the discriminant is not equal to any of the explicitly matched values. This allows the prover to use this information when verifying code in the wildcard branch.

## Example

```rust
pub fn f(x: usize) {
    match x {
        1 => assert!(x == 1),
        _ => assert!(x != 1),  // now provable
    }
}
```

The generated Why3 code now includes an assume statement:
```
default -> (! -{x <> 1}- bb1)
```

instead of just:
```
default -> (! bb1)
```

For multiple matched values, the assumptions are combined:
```
default -> (! -{(x <> 1 /\ x <> 2) /\ x <> 3}- bb1)
```

## Changes

- Added assumption building for `Branches::Int` and `Branches::Uint` in `creusot/src/backend/program.rs`
- Added test case `tests/should_succeed/bug/1248.rs` covering:
  - Single value (usize)
  - Multiple values (usize)  
  - Signed integers (i32)
  - Negative values (i32)
- Updated expected output for 3 existing tests affected by this change

### DIFF
--- a/tests/should_succeed/bug/1248/proof.json
+++ b/tests/should_succeed/bug/1248/proof.json
@@ -1,0 +1,17 @@
+{
+  "profile": [],
+  "proofs": {
+    "M_multiple_values_usize": {
+      "vc_multiple_values_usize": { "prover": "cvc5@1.3.1", "time": 0.008 }
+    },
+    "M_negative_values_i32": {
+      "vc_negative_values_i32": { "prover": "cvc5@1.3.1", "time": 0.008 }
+    },
+    "M_single_value_i32": {
+      "vc_single_value_i32": { "prover": "cvc5@1.3.1", "time": 0.007 }
+    },
+    "M_single_value_usize": {
+      "vc_single_value_usize": { "prover": "cvc5@1.3.1", "time": 0.007 }
+    }
+  }
+}


creusot/src/backend/program.rs
```diff
@@ -946,6 +946,25 @@ impl<'tcx> Branches<'tcx> {
...
+                // Build assumption for default branch: discr != val1 && discr != val2 && ...
+                let default_assumption = brs
+                    .iter()
+                    .map(|&(mut val, _)| {
+                        let why_ty =
+                            Type::qconstructor(names.in_pre(ity_to_prelude(ctx.tcx, *ity), "t"));
+                        let e = if names.bitwise_mode() {
+                            if val < 0 && *ity != IntTy::I128 {
+                                let target_width = ctx.tcx.sess.target.pointer_width;
+                                val += 1 << ity.normalize(target_width).bit_width().unwrap();
```
Lysxia: shouldn't this use `wrapping_add` to deal with overflow? But I see this was inherited from the existing code below so that needs to be fixed too.

Instead of iterating through `brs` twice, can you deduplicate the code by doing one pass to collect `(e, tgt)` in a `Vec`, and reuse that to build the final expressions?

jhjourdan: +1 at deduplication.

About overflow: I don't think there is a problem, because `val` is a negative i128, so clearly `val + (1 << 8/16/32/64)`  is within the range of i128.

Lysxia: oh duh!

creusot/src/backend/program.rs
```diff
@@ -946,6 +946,25 @@ impl<'tcx> Branches<'tcx> {
...
+                            }
+                            Exp::Const(Constant::Uint(val as u128, Some(why_ty)))
+                        } else {
+                            Exp::Const(Constant::Int(val, Some(why_ty)))
+                        };
+                        discr.clone().neq(e)
+                    })
+                    .reduce(|acc, e| acc.log_and(e))
+                    .unwrap_or_else(Exp::mk_true);
```
jhjourdan:
```suggestion
                    .rfold(Exp::mk_true(), |acc, e| e.log_and(acc));
```

This should reassociate conjuncts and avoid the unpleasant series of parentheses. Also, note that `log_and` is a smart constructor that will absorb the trailing "true" conjunct.

creusot/src/backend/program.rs
```diff
@@ -975,7 +994,16 @@ impl<'tcx> Branches<'tcx> {
...
+                // Build assumption for default branch: discr != val1 && discr != val2 && ...
+                let default_assumption = brs
+                    .iter()
+                    .map(|&(val, _)| {
+                        let why_ty =
+                            Type::qconstructor(names.in_pre(uty_to_prelude(ctx.tcx, *uty), "t"));
+                        discr.clone().neq(Exp::Const(Constant::Uint(val, Some(why_ty))))
+                    })
+                    .reduce(|acc, e| acc.log_and(e))
+                    .unwrap_or_else(Exp::mk_true);
```
jhjourdan: Same remarks about deduplication and ressociating conjuncts here.

tests/should_succeed/bug/1248.coma
```diff
@@ -0,0 +1,121 @@
...
+  use creusot.int.Int32
+  use creusot.prelude.Any
+  
+  meta "compute_max_steps" 1000000
+  
+  meta "select_lsinst" "all"
+  
+  let rec negative_values_i32 (x: Int32.t) (return (x'0: ())) = (! bb0
+    [ bb0 = any
+      [ br0 -> {x = (4294967295: Int32.t)} (! bb7)
```
jhjourdan: Your test uncovered an independent bug where `4294967295` appears here instead of `-1`. Its unclear why.

Could you comment out this test for now (which is blocking CI), while we fix this bug?

jhjourdan:
Thank you for your contribution! I have some minor comments, but I think we should merge it as soon as possible.

tests/should_succeed/bug/1248.coma
Lysxia: Fixed in #1901

tests/should_succeed/bug/1248.rs
```diff
@@ -0,0 +1,53 @@
...
+//         -1 => assert!(x == -1),
+//         0 => assert!(x == 0),
+//         1 => assert!(x == 1),
+//         _ => {
+//             assert!(x != -1);
+//             assert!(x != 0);
+//             assert!(x != 1);
+//         }
+//     }
+// }
```
Lysxia: This no longer needs to be commented since the bug was fixed.

Lysxia:
To update the why3 tests: `cargo test --test why3 -- 1248 --update`

ejackson-vp:
Thanks for the feedback and coaching! I think I've addressed all the items now

jhjourdan:
Thanks!

PR #1900, opened 2026-02-05
Add requirements for Relaxed memory feature

lafeychine:


Lysxia:
I'm not yet sure we want the changes to the test suite, so to not overwrite them I've opened #1929 and I'll leave this branch live for a while.

PR #1901, opened 2026-02-05
Fix translation of negative integer patterns

Lysxia:
Bug found in [#1899](https://github.com/creusot-rs/creusot/pull/1899/changes#r2768099292)


jhjourdan:
Thanks. I still don't understand why it's a problem for patterns but not for integers litterals...

Lysxia:
The rustc API has proper abstract types to represent constants so that you must call the appropriate decoding functions to get integer values out. Integer patterns are just naked `i128` in MIR so it's easy to forget to decode them properly.

PR #1903, opened 2026-02-05
Refactor logical reborrowing and the final borrow analysis.

jhjourdan:
This follows some work we did with @lafeychine on RustHornBelt, that clarified a few details in my mind.

It is a bit frustrating that we cannot prove this specification:

```
#[ensures(result = &mut ** b)]
fn unnest(x : &mut &mut T) -> &mut T {
    &mut ** x
}
```

It is both due to an incompleteness in the final borrow analysis, and the fact that this kind of unnesting is not supported in Pearlite (we would need special syntax for that).

It is a bit frustrating, because RustHornBelt clearly tells what we should do in programs, but I don't think we want a new syntax for this in Pearlite.


xldenis:
why would we need new syntax?

jhjourdan:
Because in Pearlite, `&mut **x` may mean two different things: either `^&mut **x`  is defined as  `^*x`, or as `*^x`. We have chosen to use the first definition, which makes a semantics of Pearlite which is composable (i.e., `let b = *x; &mut *b` is equivalent to `&mut **x`, and which may be useful in some cases (see `write_inner_borrow` in `should_succeed/bug/final_borrow.rs`). Hence there is no syntax for the other definition.

Somehow, when you do logical reborrowing, there would need to be a way to describe which is the deref which provides prophecies.

PR #1913, opened 2026-02-10
Expose the `logic::fmap` module

dianegolfouse:


jhjourdan:
Two comments:

1- Generally, I don't like the fact that an item is accessible through several different paths. This adds confusion (of course, the prelude is an exception). So if we make `fmap` public, then we should ne longer re-export `FMap`. Same thing for `FSet`, `Seq` and `OrdLogic`.

2- I think we decided that we would not support the use case of logiclal containers where they would contain unsized object, because it is too much boilerplate to have two versions of each function. Cannot you wrap the elements in `Box` instead?

Lysxia:
Yeah, for symmetry we rather want to hide `logic::fset`, and move the extra lemmas it currently contains to methods on `FSet`.

dianegolfouse:
For 1, I copied the behavior of collections in Rust's `std`. We could only expose `FMap` in `creusot-std::logic` though (resp. `FSet`). We still need the `fmap` module for the iterator types (that I would rather not see spill into `logic`).

For 2, yeah I thought `FSet` could still contain unsized data: I'll simply remove the new method.

jhjourdan:
> For 1, I copied the behavior of collections in Rust's `std`. We could only expose `FMap` in `creusot-std::logic` though (resp. `FSet`). We still need the `fmap` module for the iterator types (that I would rather not see spill into `logic`).

Interesting. I never realized Rust was doing that already. Let's say it's acceptable then, you're right.

dianegolfouse:
> Yeah, for symmetry we rather want to hide `logic::fset`, and move the extra lemmas it currently contains to methods on `FSet`.

I moved the lemmas to the impl block (and made them opaque, I don't know why they were open).

jhjourdan:
@Lysxia : if you are happy (you are the author of the lemmas of FSet tht @arnaudgolfouse moved), then please merge!

jhjourdan:
And, well, again, the creusot-std session is obsolete. @arnaudgolfouse : have you reinstalled why3 since the last bump?

dianegolfouse:
> And, well, again, the creusot-std session is obsolete. @arnaudgolfouse : have you reinstalled why3 since the last bump?

Ah yes, thanks !

PR #1914, opened 2026-02-10
Pointer operations on `Rc` and `Arc`

dianegolfouse:


dianegolfouse:
I mirrored `Rc::clone` behavior for `Arc::clone`, which feels correct (clone does change the reference count, but the pointer you get is exactly the same).

jhjourdan:
Good! I'm happy to merge this as-is, but it seems like to be actually usable, some kind of injectivity lemma needs to be added. However, I don't know what we should use: if we say that `as_ptr` is injective with respect to addresses, then we are unsound because of the ABA problem (the Rc can be deallocated and then reallocated at the same address but with a different payload). But if we say that it is injective wrt. the pointer value, it is kind of useless because we cannot compare pointer values in Rust code (only their addresses). Also, the interraction with `get_mut` is unclear, because the pointer stays the same, but the value changes.

jhjourdan:
(BTW, obsolete creusot-std has an obsolete session.)

PR #1915, opened 2026-02-11
Replace FMap iterators with FMap, hide `logic::{fmap, fset}`

Lysxia:
This makes `FMap` the only export left in `logic::fmap` so we can remove the module (since we already have `logic::fmap::FMap`.


creusot-std/src/logic/fmap.rs
```diff
@@ -614,89 +586,20 @@ impl<K, V> IteratorSpec for FMapIter<K, V> {
...
-        let ac = ab.concat(bc);
-        proof_assert!(forall<x> ab.contains(x) ==> ac.contains(x));
-        proof_assert!(forall<i> 0 <= i && i < bc.len() ==> ac[i + ab.len()] == bc[i]);
-        proof_assert!(forall<k> (forall<i> 0 <= i && i < ac.len() ==> ac[i].0 != k) ==> {
-            (forall<i> 0 <= i && i < ab.len() ==> ab[i].0 != k) &&
-            (forall<i> 0 <= i && i < bc.len() ==> bc[i].0 != k) &&
-            a@.get(*k) == b@.get(*k) && b@.get(*k) == c@.get(*k)
-        });
+    #[ensures(result.len() == self.len())]
+    #[ensures(forall<k, v> (self.get(k) == Some(v)) == (result.get(&k) == Some(&v)))]
```
jhjourdan, line 600: We could have a function `as_ref_logic`, which would be the logical counterpart of `as_ref_ghost`, and specify functions in terms of it, instead of extensionally. The advantage would be that we could tell Creusot to extract it to the identity in WhyML, because shared references are erased by Creusot.

jhjourdan:
Good (although I'd be very happy with the small improvement I am proposing.)

creusot-std/src/logic/fmap.rs
Lysxia: For some reason `FMap` does not translate to Why3's `fmap` so `FMap<&K, &V>` and `FMap<K, V>` to not translate to equal types, and we cannot make an `as_ref_logic` that extracts to the identity. I'll open an issue to remember this.

Lysxia:
Further improvement tracked in #1916

PR #1918, opened 2026-02-11
Fix dummy ghost!: don't insert check(ghost)

Lysxia:


jhjourdan:
ARe we sure that this can never change the semantics in any ways?

Lysxia:
In dummy mode `check(ghost)` erases itself anyway. The issue is that attributes on expressions (here closures) are unstable, so this broke builds with stable rustc (for example, parallel_add).

jhjourdan:
Oh, I see. I went to the safe way by doing the same on both modes, but if it breaks on stable rustc, let's merge this.

PR #1938, opened 2026-02-20
Implement `OrdLogic` for `PositiveReal`

dianegolfouse:


jhjourdan:
I'm happy. with merging this, although I think in practice it is probably better to use `RA::incl` instead.

PR #1939, opened 2026-02-20
Specify `MaybeUninit`

dianegolfouse:
This gives a basic specification for `MaybeUninit`. It _cannot_ be used to verify code that initializes a `MaybeUninit` field by field, such as:
```rust
let mut x: MaybeUninit<(i32, i32)> = MaybeUninit::uninit();
unsafe {
    addr_of_mut!((*x.as_mut_ptr()).0).write(1);
    addr_of_mut!((*x.as_mut_ptr()).1).write(2);
    assert_eq!(x.assume_init(), (1, 2));
};
```

### DIFF
--- a/creusot-std/src/std/mem.rs
+++ b/creusot-std/src/std/mem.rs
@@ -1,6 +1,14 @@
 use crate::prelude::*;
-#[cfg(creusot)]
 use core::mem::*;
+
+impl<T> View for MaybeUninit<T> {
+    type ViewTy = Option<T>;
+
+    #[logic(opaque)]
+    fn view(self) -> Option<T> {
+        dead
+    }
+}
 
 extern_spec! {
     mod core {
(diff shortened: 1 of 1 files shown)


jhjourdan:
> This gives a basic specification for `MaybeUninit`. It _cannot_ be used to verify code that initializes a `MaybeUninit` field by field, such as:
> 
> ```rust
> let mut x: MaybeUninit<(i32, i32)> = MaybeUninit::uninit();
> unsafe {
>     addr_of_mut!((*x.as_mut_ptr()).0).write(1);
>     addr_of_mut!((*x.as_mut_ptr()).1).write(2);
>     assert_eq!(x.assume_init(), (1, 2));
> };
> ```

Is this really allowed in Rust?

creusot-std/src/std/mem.rs
```diff
@@ -54,6 +62,42 @@ extern_spec! {
...
+            impl<T> MaybeUninit<T> {
+                #[ensures(result@ == Some(val))]
+                #[check(ghost)]
+                fn new(val: T) -> Self;
+
+                #[ensures(result@ == None)]
+                #[check(ghost)]
+                fn uninit() -> Self;
+
+                #[requires(self@ == None)] // NOTE: we could allow calling this on an initialized `MaybeUninit`, but this would leak the contents.
```
jhjourdan, line 75: Fair enough. But then, we should have a primitive that does not do anything operationally, by changes the view from `Some(..)` to `None`.

jhjourdan:
LGTM.

dianegolfouse:
> > This gives a basic specification for `MaybeUninit`. It _cannot_ be used to verify code that initializes a `MaybeUninit` field by field, such as:
> > ```rust
> > let mut x: MaybeUninit<(i32, i32)> = MaybeUninit::uninit();
> > unsafe {
> >     addr_of_mut!((*x.as_mut_ptr()).0).write(1);
> >     addr_of_mut!((*x.as_mut_ptr()).1).write(2);
> >     assert_eq!(x.assume_init(), (1, 2));
> > };
> > ```
> 
> Is this really allowed in Rust?

Yes, [the example in the stdlib](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#initializing-a-struct-field-by-field) even uses types with ownership to drive the point home.

creusot-std/src/std/mem.rs
dianegolfouse: To be fair I'm not really convinced about this precondition, we could just remove it if it becomes an issue.

PR #1943, opened 2026-02-22
fix: handle copying binaries over broken symlinks during installation

FranklinChen:
Bug: ./INSTALL fails with "No such file or directory" when re-installing over a stale symlink

When re-running ./INSTALL on a macOS machine that had I had previously set up long ago with an opam-managed Alt-Ergo (i.e. bin/alt-ergo was a symlink into an opam switch), the installer failed with:

   Install Alt-Ergo 2.6.2...Skipped download (cached)
   Error: No such file or directory (os error 2)

Root cause: bin/alt-ergo was a symlink pointing to /Users/.../.opam/ocaml.4.14.1/bin/alt-ergo, left over from the previous installation. That opam switch no longer existed, making the symlink broken. std::fs::copy follows symlinks on the destination side rather than replacing them, so it tried to write through the dangling symlink to a nonexistent path and failed with ENOENT.

Fix: Remove the destination before copying, mirroring the pattern already used in Args::symlink_file. A regression test exercises this scenario directly: it creates a broken symlink at the destination, then asserts that Args::copy succeeds and replaces it with the actual file.


Lysxia:
Thanks for the fix!

Can you use [`std::env::temp_dir`](https://doc.rust-lang.org/std/env/fn.temp_dir.html) for the test instead of tempdir?

FranklinChen:
Done!

Lysxia:
In the end I decided to remove the test because I don't find it that useful.

PR #1946, opened 2026-02-23
Add DeepModel for str and String

Lysxia:


jhjourdan:
Recent smt solvers have special support for Unicode strings. One day we should plug String to that'

PR #1954, opened 2026-02-27
Some improvement to resource algebras

dianegolfouse:
- add a convenience method to `Authority` to add a new fragment inside of it
- document updates


creusot-std/src/logic/ra/auth.rs
```diff
@@ -86,3 +88,31 @@ impl<R: UnitRA, U: LocalUpdate<R>> Update<Auth<R>> for AuthUpdate<U> {
...
+/// Add some data to both an authority and a fragment, at the same time.
+///
+/// Used in `Authority::add_fragment`.
+pub(crate) struct AuthIncrease<R>(pub Snapshot<R>);
+
+impl<R: UnitRA> LocalUpdate<R> for AuthIncrease<R> {
+    #[logic(open)]
+    fn premise(self, from_auth: R, from_frag: R) -> bool {
+        from_auth.op(*self.0) != None
+            && from_frag.op(*self.0) != None
```
jhjourdan: That should not be necessary. This is derived from the first premise.

creusot-std/src/logic/ra/auth.rs
```diff
@@ -86,3 +88,31 @@ impl<R: UnitRA, U: LocalUpdate<R>> Update<Auth<R>> for AuthUpdate<U> {
...
+            && AuthViewRel::rel(Some(from_auth), from_frag)
```
jhjourdan: Same here. This is implied by the requires clause below.

creusot-std/src/logic/ra/fmap.rs
```diff
@@ -142,6 +142,9 @@ impl<K, V: RA> FMap<K, V> {
     }
 }
 
+/// Add a key-value to an authority/fragment pair of [`FMap`]s.
+///
+/// It requires the key to not already be in the map.
```
jhjourdan:
```suggestion
/// It requires that the key is not in the map yet.
```

creusot-std/src/logic/ra/sum.rs
```diff
@@ -162,6 +204,10 @@ impl<R: RA, U: Update<R>, V: RA> Update<Sum<V, R>> for SumUpdateR<U> {
     }
 }
 
+/// Apply an [update](LocalUpdate) to the [`Left`](Sum::Left) variant of an
+/// authority/fragment pair of [`Sum`]s.
+///
+/// This requires both the authority and the fragment to be in the `Left` state.
```
jhjourdan: Well, actually it's either of them, since the other will be implied.

creusot-std/src/logic/ra/sum.rs
```diff
@@ -210,6 +256,10 @@ impl<R1: RA, R2: RA, U: LocalUpdate<R1>> LocalUpdate<Sum<R1, R2>> for SumLocalUp
     }
 }
 
+/// Apply an [update](LocalUpdate) to the [`Right`](Sum::Right) variant of an
+/// authority/fragment pair of [`Sum`]s.
+///
+/// This requires both the authority and the fragment to be in the `Right` state.
```
jhjourdan: Either

creusot-std/src/logic/ra/update.rs
```diff
@@ -5,23 +5,50 @@ use crate::{
...
     fn update(self, from: R, ch: Self::Choice) -> R;
 
+    /// Frame preservation.
+    ///
+    /// This is a lemma that must be proven by implementors to ensure that the
+    /// update is consistent.
+    ///
+    /// It states that if we were to apply that update, the resulting resource
+    /// would have _more_ valid compositions that what we started with, whatever
+    /// `Choice` we made.
```
jhjourdan: Rather, for any frame, there is the choice that preserves validity. The choice depends on the frame. The whole point is that there are some updates where no choice apply for any frame.

creusot-std/src/logic/ra/view.rs
```diff
@@ -263,6 +265,10 @@ impl<R: ViewRel, Choice> Update<View<R>> for ViewUpdate<R, Choice> {
     }
 }
 
+/// Apply an [update](Update) to a [`View`], that changes both the authority and the
+/// fragment part.
+///
+/// This is the same as [`ViewUpdate`], but deterministic.
```
jhjourdan: That's not quite the purpose of this update. If you look at the trait instance, we completely ignore the fragment of `from`, even in the premise (in contrary to `ViewUpdate`). So, here, wlog, we can assume that the fragment is unit.

The purpose of this update is actually quite close to your `IncreaseUpdate`, but stated generally for the `View` RA: insert an element in the fragment, and update the authority accordingly.

Note that these updates are more or less direct translations of lemmas in Iris, and they may need to be fine tuned to Creusot. For example, we may consider not discarding the fragment here, but rather compose it with the inserted one. But it's not clear to me that this improves provability because we may need to rely on the fact that the unit cancels out if the fragment is initially the unit.

creusot-std/src/logic/ra/view.rs
```diff
@@ -296,6 +302,11 @@ impl<R: ViewRel> Update<View<R>> for ViewUpdateInsert<R> {
     }
 }
 
+/// Apply an [update](Update) to a [`View`], that replace the view with a pure
+/// authority.
+///
+/// This is the same as [`ViewUpdateInsert`], with the fragment part set to the
+/// unit value of [`R::Frag`](ViewRel::Frag).
```
jhjourdan: And this is the not the same as `Insert`, it is the dual: remove a fragment, and update the authority accordingly.

creusot-std/src/logic/ra/auth.rs
```diff
@@ -86,3 +88,31 @@ impl<R: UnitRA, U: LocalUpdate<R>> Update<Auth<R>> for AuthUpdate<U> {
         self.0.frame_preserving(auth, from.frag(), Some(f));
     }
 }
+
+/// Add some data to both an authority and a fragment, at the same time.
+///
+/// Used in `Authority::add_fragment`.
+pub(crate) struct AuthIncrease<R>(pub Snapshot<R>);
```
jhjourdan: This is called `op_local_update` in Iris (https://github.com/JasonGross/iris-coq/blob/master/theories%2Falgebra%2Flocal_updates.v#L28) and when this makes sense, I would prefer keeping the names from Iris so that Iris users do not get confused. 

Isn't `OpLocalUpdate` a good name for this?

creusot-std/src/logic/ra/view.rs
dianegolfouse: > If you look at the trait instance, we completely ignore the fragment of `from`

We could do the same in `ViewUpdate`, and replace `match from.frag().op(frame) { ... }` with `R::rel(from.auth(), frame)`, no? It still proves frame preservation, and it is less restrictive...

jhjourdan: That's the other way around, isn't it?

`R::rel(from.auth(), frame)` is weaker than `match from.frag().op(frame) { ... }`, hence the premise of `ViewUpdateInsert` is ***stronger*** than the premise of `ViewUpdate`.

dianegolfouse: Thanks for the explanations, I think I am getting it now!

creusot-std/src/logic/ra/auth.rs
dianegolfouse: done

PR #1955, opened 2026-02-27
Add a sequentially-consistent message-passing example

lafeychine:
Thanks to @arnaudgolfouse for helping me out with the `State` enumeration and loop invariant

### DIFF
--- a/creusot-std/src/std/sync/atomic_sc.rs
+++ b/creusot-std/src/std/sync/atomic_sc.rs
@@ -6,6 +6,9 @@ use crate::{
 /// Creusot wrapper around [`std::sync::atomic::AtomicI32`]
 pub struct AtomicI32(::std::sync::atomic::AtomicI32);
 
+unsafe impl Send for Perm<AtomicI32> {}
+unsafe impl Sync for Perm<AtomicI32> {}
+
 impl Container for AtomicI32 {
     type Value = i32;
 
--- a/examples/message_passing_sc.rs
+++ b/examples/message_passing_sc.rs
@@ -1,0 +1,125 @@
+extern crate creusot_std;
+
+use creusot_std::{
+    cell::PermCell,
+    ghost::{
+        Committer,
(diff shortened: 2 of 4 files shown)


creusot-std/src/std/sync/atomic_sc.rs
```diff
@@ -44,6 +47,26 @@ impl AtomicI32 {
         (res, Ghost::conjure())
     }
 
+    /// Wrapper for [`std::sync::atomic::AtomicI32::fetch_add`].
```
jhjourdan, line 50: Wrong doc

examples/message_passing_sc.rs
```diff
@@ -0,0 +1,125 @@
...
+            let excl_snap = snapshot!(excl);
+
+            #[invariant(excl == *excl_snap)]
+            #[invariant(tokens.contains(MESSAGE_PASSING()))]
+            loop {
+                let (atomic_res, data_own) = atomic.load(ghost! { |c: &mut Committer<_>| {
+                    inv.open(tokens.reborrow(), |inv: &mut MessagePassingAtomicInv| {
+                        if let State::Readable(excl_state) = &inv.state {
+                            excl.as_mut().unwrap().valid_op_lemma(excl_state);
+                        }
```
jhjourdan, line 95: Cannot you merge this with the pattern matching bellow?

jhjourdan:
Nice! This looks good!

I have a few minor comments, but this LGTM overall.

Out of curiosity, why did you choose to use an enum instead of a struct of options as you initially imagined? If using a struct, I would expect the invariant to be slightly more difficult to state but the ghost code shorter. Anyway, I like this version.

lafeychine:
> Out of curiosity, why did you choose to use an enum instead of a struct of options as you initially imagined? If using a struct, I would expect the invariant to be slightly more difficult to state but the ghost code shorter.

Both the protocol invariant and the ghost state were difficult to write in the struct version (e.g. I needed to use `mem::swap` on Options)

As this invariant uses states, enums are indeed preferable

PR #1958, opened 2026-03-03
Add `AddAssign`, etc. for `Int`

Lysxia:


jhjourdan:
Could we add preconditions to `Div`, `Rem`, `DivAssign` and `RemAssign` stating that the divisor needs to be nonzero? This reflects what is happening with machine integers.

Also, if you add thse instantces, you should also add `extern_spec`s for all the arithmetic traits instances for bounded integers. We most often don't use them because the arithmetic operators translate to primitive MIR operators, but the user may actually use function calls or abstract over the type, inwhich case the extern spec is important.

PR #1959, opened 2026-03-04
Rename `AddLogic::add` to `AddLogic::add_logic`, etc.

Lysxia:
Avoid collision with `Add::add` that happened when I tried to add an extern spec to an impl for `Add`, in a module where `AddLogic` was imported. This also points to a flaw in `extern_spec!` for trait impls (it throws away the trait, which results in the above ambiguity) to be fixed soon.

### DIFF
--- a/creusot-std/src/logic/ops/arithmetic.rs
+++ b/creusot-std/src/logic/ops/arithmetic.rs
@@ -11,7 +11,7 @@ pub trait AddLogic<Rhs = Self> {
     type Output;
 
     #[logic]
-    fn add(self, other: Rhs) -> Self::Output;
+    fn add_logic(self, other: Rhs) -> Self::Output;
 }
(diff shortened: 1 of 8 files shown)


creusot-std/src/logic/ops/arithmetic.rs
```diff
@@ -80,5 +80,5 @@ pub trait NthBitLogic {
     /// In [`bitwise_proof`] mode, this is translated to [`nth`](https://www.why3.org/stdlib/bv.html#nth_188) in Why3.
     /// Otherwise this is an abstract function.
     #[logic]
-    fn nth_bit(self, n: Int) -> bool;
+    fn nth_bit_logic(self, n: Int) -> bool;
```
Lysxia: no need to rename this one

PR #1960, opened 2026-03-04
Add extern specs for arithmetic operations

Lysxia:
Address https://github.com/creusot-rs/creusot/pull/1958#issuecomment-3992262539


Lysxia:
Note that this only specifies `Add<Rhs=$int_type> for $int_type`, not all of the ref wrappers (`Add<Rhs=&u32> for u32`, `Add<Rhs=u32> for &u32`, `Add<Rhs=&u32> for &u32`). I'm working on fixing `extern_spec!` to allow this.

NOW FIXED

Issue #1957, opened 2026-03-03
Termination of `next` for `Range`


Lysxia:
(Context: trying to prove termination of simple for loops.)

In theory, fixing this is as simple as adding an `extern_spec`. The difficulty is that the impl is polymorphic in the index type:

```
impl<A: Step> Iterator for Range<A> {
  fn next ...
}
```

So the termination of `next` depends on the termination of `Step::forward_checked`.

jhjourdan:
We more and more need first-class tags for proof modes.

This would cover:
- Termination
- panic freedom vs. UB freedom only
- ghost mode vs. program mode
- SC vs. Weak memory
- others?

Then we would use these in specifications. This would solve this issue, but also simplify the specs for closures, better handle the case of ghost s non-ghost functions, etc...

PR #1962, opened 2026-03-05
 Add `Seq::{as_refs, as_muts, remove, extract, extend, eq_ref}` and refactor `Seq` iterators

Lysxia:
- Add a mut iterator for `Seq`: it is basically `as_muts`
- Also implement `get_ghost` and `get_mut_ghost` using these primitives (showing one use of `extract` instead of `remove`)


creusot-std/src/logic/seq.rs
```diff
@@ -395,6 +395,16 @@ impl<T> Seq<T> {
     #[ensures(forall<a: Seq<T>, b: Seq<T>, x>
         a.concat(b).contains(x) == a.contains(x) || b.contains(x))]
     pub fn concat_contains() {}
+
+    /// Equality between `Seq<T>` and `Seq<&T>`
+    ///
+    /// In Why3 these are the same type and this is translated to polymorphic equality `(=)`.
+    #[logic]
+    #[builtin("(=)")]
+    pub fn eq_ref(self, rhs: Seq<&T>) -> bool {
```
jhjourdan: Couldn't you encode this from `pub fn to_owned_seq(self) -> Seq<T>`?

In general, I prefer limiting the number of logical methods that use axiomatically this magical cast.
I'm not necessarily against adding this function (although it's so short that it could actually just be inlined), but I think it should not be axiomatized.

creusot-std/src/logic/seq.rs
```diff
@@ -614,15 +624,18 @@ impl<T> Seq<T> {
...
     #[check(ghost)]
     #[ensures(match self.get(index) {
         None => result == None,
         Some(v) => result == Some(&v),
     })]
     pub fn get_ghost(&self, index: Int) -> Option<&T> {
-        let _ = index;
-        panic!()
+        // Trick: we can't write 0 outside of a `ghost!` block
+        if index - index <= index && index < self.len_ghost() {
```
jhjourdan:
```suggestion
        if Int::new(0) <= index && index < self.len_ghost() {
```

Lysxia: `Int::new(0): Ghost<Int>`

creusot-std/src/logic/seq.rs
Lysxia: Oh yes, I missed that one.

creusot-std/src/logic/seq.rs
jhjourdan: Ah, oups, sorry, it actually returns a `Ghost<Int>`...

Please make this a FIXME, then. This would be fixed by #1957.

creusot-std/src/logic/seq.rs
```diff
@@ -738,6 +800,25 @@ impl<T> Seq<T> {
         let _ = mid;
         panic!("ghost code")
     }
+
+    /// Borrow every element of a borrowed sequence.
+    #[trusted]
+    #[check(ghost)]
+    #[ensures(self.eq_ref(result))]
+    pub fn as_ref(&self) -> Seq<&T> {
```
jhjourdan:
```suggestion
    pub fn borrow_elements(&self) -> Seq<&T> {
```

In Rust, `as_ref` functions usually return... a reference, which is not the case here.

(I'm open to another name, but this one is probably not a good one.)

creusot-std/src/logic/seq.rs
```diff
@@ -738,6 +800,25 @@ impl<T> Seq<T> {
...
+    }
+
+    /// Mutably borrow every element of a borrowed sequence.
+    #[trusted]
+    #[check(ghost)]
+    #[ensures(result.len() == self.len())]
+    #[ensures((^self).len() == self.len())]
+    #[ensures(forall<i> 0 <= i && i < self.len() ==> *result[i] == (*self)[i])]
+    #[ensures(forall<i> 0 <= i && i < self.len() ==> ^result[i] == (^self)[i])]
+    pub fn as_mut(&mut self) -> Seq<&mut T> {
```
jhjourdan: Same here. Suggestion `borrow_mut_elements`.

creusot-std/src/logic/seq.rs
Lysxia: I was thinking of `Option::as_ref(&Option<T>) -> Option<&T>` but maybe you have a point.

creusot-std/src/logic/seq.rs
```diff
@@ -865,19 +938,17 @@ impl<T> IntoIterator for Seq<T> {
 ///     }
 /// }
 /// ```
-pub struct SeqIter<T> {
-    inner: Seq<T>,
-}
+pub struct Iter<T>(Seq<T>);
```
jhjourdan, line 953: Why a newtype here? Cannot we simply pretend that `Seq` itself is an iterator, like we do for e.g., `FMap`?

Lysxia: This is to allow `IntoIterator for &mut Seq<T>`, which conflicts with `Iterator for Seq<T>`. Alternatively, one could explicitly call `as_mut` when iterating over `&mut Seq<T>`. (Either way, we should make it agree with `FMap`.)

creusot-std/src/logic/seq.rs
jhjourdan: Actually, I wish we could have two logical functions:

`as_ref_logic` : `Seq<T> -> Seq<&T>`

and

`as_mut_logic` : `&mut Seq<T> -> Seq<&mut T>`

Which do the obvious thing, translated as a builtin(identity) or axiomatized in the obvious way, and then use these two functions in the specs of `as_ref` ans `as_mut` and of the iterators.

creusot-std/src/logic/seq.rs
```diff
@@ -1001,32 +1033,33 @@ impl<T> Resolve for Seq<T> {
...
 }
 
-impl<T> Resolve for SeqIter<T> {
-    #[logic(open, prophetic, inline)]
+impl<T> Resolve for Iter<T> {
+    #[logic(open, prophetic)]
     #[creusot::trusted_trivial_if_param_trivial]
     fn resolve(self) -> bool {
-        pearlite! { resolve(self@) }
+        pearlite! { forall<i : Int> resolve(self@.get(i)) }
```
jhjourdan: Why this change? This is not really true because it states things about the values outside of the sequence bounds, and, it adds a `#[trusted]` attrbute where it is not needed.

creusot-std/src/logic/seq.rs
```diff
@@ -1001,32 +1033,33 @@ impl<T> Resolve for Seq<T> {
...
     }
 
-    #[logic(open, prophetic)]
+    #[trusted]
+    #[logic(open(self), prophetic)]
```
jhjourdan:
```suggestion
    #[logic(prophetic)]
```

creusot-std/src/logic/seq.rs
Lysxia: Collateral damage from removing the iterator then readding it.

creusot-std/src/logic/seq.rs
jhjourdan: Hum... Actually you are right. It feels a bit wrong to me, perhaps because there are several refs in such a sequence... Perhaps `as_refs`?

creusot-std/src/logic/seq.rs
jhjourdan: Why does it conflict?

Lysxia: It conflicts with `IntoIterator for I where I: Iterator`.

And now that I'm looking at it, if we chose to make `Iterator for Seq<T>` and not have `IntoIterator for &mut Seq<T>`, that would have the drawback that iterating over `&mut Seq<T>` would consume it. Accordingly, I should probably revert making `FMap` an iterator.

creusot-std/src/logic/seq.rs
```diff
@@ -652,8 +654,57 @@ impl<T> Seq<T> {
...
+    pub fn extract(mut self, index: Int) -> T {
+        proof_assert! { forall<i> index < i && i < self.len() ==> self[i] == self[index + 1..][i - index - 1] }
+        self.split_off_ghost(index).pop_front_ghost().unwrap()
+    }
+
+    /// Remove an element from a sequence.
+    #[check(ghost)]
+    #[requires(0 <= index && index < self.len())]
+    #[ensures(result == self[index])]
+    #[ensures(^self == (*self)[..index].concat((*self)[index + 1..]))]
```
Lysxia: Maybe it would be convenient to have a logic version of `remove` here too.

PR #1963, opened 2026-03-06
doc: Fix doc links generated by extern_spec

Lysxia:


jhjourdan:
Should we have a tool that checks all the links generated in the doc? Does such a tool already exist?

Lysxia:
There are already warnings, but we just don't check them in CI :)

PR #1967, opened 2026-03-06
Fix termination checker

### LINKED ISSUE #1966: Termination checker doesn't catch the simplest recursive function
```rust
#[logic]
#[ensures(false)]
fn q() {
  q()
}
```

Lysxia:
Fix #1966

~~There are quite a few tests with structurally recursive functions, so they need `#[trusted]`. Or we could implement a check for structural recursion, or define it as a well-founded relation.~~ EDIT: see first comment.

### DIFF
--- a/creusot-std/src/stubs.rs
+++ b/creusot-std/src/stubs.rs
@@ -46,6 +46,7 @@ pub fn old<'a, T: ?Sized>(_: T) -> &'a T {
     dead
 }
 
+#[creusot::no_translate] // avoid the termination checker
 #[logic(opaque)]
 #[intrinsic("dead")]
 #[allow(unconditional_recursion)]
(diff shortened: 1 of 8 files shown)


jhjourdan:
Wait, we do have a checker for structural induction. We should not trust the functions that use it. Instead, we should fix the structural inducton checker.

Lysxia:
Ah OK, I just made the check in `terminates.rs` too strict, it should skip logic functions (but still record that they are recursive).

jhjourdan:
Thanks for debugging that.

There is still another problem, I think: structural recursion is checked in the VCGen, which is disabled for logic functions without a contract (see `translate_logic`).

I guess we should do this check earlier somewhere else when we detect that there is recursion without a variant clause.

Lysxia:
We now call the structural recursion check in `terminates.rs` (also fixed some bugs), so elsewhere we only need to worry about variants.

jhjourdan:
This looks very good, thank you!

PR #1972, opened 2026-03-10
Specify `alloc::Layout`

dianegolfouse:


creusot-std/src/std/alloc.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+}
+
+/// Requirements for creating a [`Layout`].
+#[logic(open)]
+pub fn layout_requirements(size: Int, align: Int) -> bool {
+    pearlite! {
+        align != 0 &&
+        (exists<a: Int> a.pow2() == align) &&
+        size <= isize::MAX@ &&
```
jhjourdan:
```suggestion
        0 <= size && size <= isize::MAX@ &&
```

creusot-std/src/std/alloc.rs
```diff
@@ -0,0 +1,99 @@
...
+        dead
+    }
+}
+
+/// Requirements for creating a [`Layout`].
+#[logic(open)]
+pub fn layout_requirements(size: Int, align: Int) -> bool {
+    pearlite! {
+        align != 0 &&
+        (exists<a: Int> a.pow2() == align) &&
```
jhjourdan:
```suggestion
        (exists<a: Int> a >= 0 && a.pow2() == align) &&
```

creusot-std/src/std/alloc.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    #[logic(opaque)]
+    fn align_log(self) -> usize {
+        dead
+    }
+}
+
+/// Requirements for creating a [`Layout`].
+#[logic(open)]
+pub fn layout_requirements(size: Int, align: Int) -> bool {
```
jhjourdan, line 49: One possibility would be to require `size` and `align` to be positive `usize` values that fit an `isize`. I think that would simplify quite a few specs.

creusot-std/src/std/alloc.rs
dianegolfouse: Does it matter? `pow2` is unspecified if `a < 0`, no?

creusot-std/src/std/alloc.rs
jhjourdan: Also, there is one missing requirement:

> `size`, when rounded up to the nearest multiple of align,
must not overflow `isize` (i.e., the rounded value must be
less than or equal to `isize::MAX`).

creusot-std/src/std/alloc.rs
```diff
@@ -0,0 +1,99 @@
...
+        #[ensures(result.size_log() == size && result.align_log() == align)]
+        const unsafe fn from_size_align_unchecked(size: usize, align: usize) -> Layout;
+
+        #[ensures(result == self.size_log())]
+        const fn size(&self) -> usize;
+
+        #[ensures(result == self.align_log())]
+        const fn align(&self) -> usize;
+
+        #[requires(layout_requirements(size_of_logic::<T>(), align_of_logic::<T>()@))]
```
jhjourdan, line 79: Is there any Rust type where this is not true?

creusot-std/src/std/alloc.rs
dianegolfouse: Yeah, but the actual program values this specifies are `usize`, so you would still need to require that they are in bounds for `isize`

creusot-std/src/std/alloc.rs
```diff
@@ -0,0 +1,99 @@
...
+        #[ensures(result.size_log()@ == size_of_logic::<T>())]
+        #[ensures(result.align_log() == align_of_logic::<T>())]
+        const fn new<T>() -> Layout {
+            unsafe { Layout::from_size_align_unchecked(size_of::<T>(), align_of::<T>()) }
+        }
+
+        #[requires(layout_requirements(self.size_log()@, align@))]
```
jhjourdan:
```suggestion
```

creusot-std/src/std/alloc.rs
```diff
@@ -0,0 +1,99 @@
...
+        #[ensures(if layout_requirements(self.size_log()@, align@) {
+            exists<res> result == Ok(res) &&
+                res.size_log() == self.size_log() &&
+                res.align_log() == align
```
jhjourdan: Nope, this is the max of the old one and of this one.

creusot-std/src/std/alloc.rs
```diff
@@ -0,0 +1,99 @@
...
+        } else {
+            exists<err> result == Err(err)
+        })]
+        const fn align_to(&self, align: usize) -> Result<Layout, LayoutError> {
+            Layout::from_size_align(self.size(), align)
+        }
+
+        #[ensures(if n@ * size_of_logic::<T>() >= isize::MAX@ {
```
jhjourdan:
```suggestion
        #[ensures(if n@ * size_of_logic::<T>() > isize::MAX@ {
```

creusot-std/src/std/alloc.rs
```diff
@@ -0,0 +1,99 @@
...
+            exists<err> result == Err(err)
+        } else {
+            let s = crate::logic::such_that(|s: usize| s@ == n@ * size_of_logic::<T>());
+            if layout_requirements(s@, align_of_logic::<T>()@) {
+                exists<res> result == Ok(res) &&
+                    res.size_log() == s &&
+                    res.align_log() == align_of_logic::<T>()
+            } else {
+                exists<err> result == Err(err)
```
jhjourdan, line 107: Isn't this dead code?

creusot-std/src/std/alloc.rs
```diff
@@ -0,0 +1,99 @@
...
+        #[ensures(result == self.size_log())]
+        const fn size(&self) -> usize;
+
+        #[ensures(result == self.align_log())]
+        const fn align(&self) -> usize;
+
+        #[requires(layout_requirements(size_of_logic::<T>(), align_of_logic::<T>()@))]
+        #[ensures(result.size_log()@ == size_of_logic::<T>())]
+        #[ensures(result.align_log() == align_of_logic::<T>())]
+        const fn new<T>() -> Layout {
```
jhjourdan, line 83: We could add an `ensures` clause telling that size is a multiple of alignement, since this is apparently the case for types (at least `array` assumes this).

creusot-std/src/std/alloc.rs
jhjourdan: That's the whole point : it is unspecified, so not necessarily positive.

creusot-std/src/std/alloc.rs
jhjourdan: I don't think your requirements implies this requirement, which is part of Rust's documentation.

creusot-std/src/std/alloc.rs
dianegolfouse: Yes ! `[u8; usize::MAX]` for example

creusot-std/src/std/alloc.rs
dianegolfouse: This is already a property of `align_of_logic`

creusot-std/src/std/alloc.rs
dianegolfouse: > One possibility would be to require `size` and `align` to be positive `usize` values that fit an `isize`. I think that would simplify quite a few specs.

That's what I originally did, but this is annoying since `size_of_logic`/`align_of_logic` return `Int`s, so you need some shenanigans to pass them to `layout_requirements`.

jhjourdan: I see. We need a way to caset from `Int` to machine integers.

creusot-std/src/std/alloc.rs
jhjourdan: Ok, fair enough.

jhjourdan:
Generally, this is typically the kind of modules where I would expect a formal proof to make sure none of the spec is unsound.

PR #1973, opened 2026-03-10
Allow naming of result in `#[ensures]`

### LINKED ISSUE #1735: `result` should not be a pseudo-keyword
> I would like `result` to be accessed with a special syntax which would not conflict with identifiers, like `res!` or so.

Lysxia:
This allows writing `#[ensures(|myresult| myresult == blah())]`

And warn instead of error when `result` is used as parameter name.

Close #1735

### DIFF
--- a/creusot-std-proc/src/creusot/specs.rs
+++ b/creusot-std-proc/src/creusot/specs.rs
@@ -12,7 +12,7 @@ use proc_macro::TokenStream as TS1;
 use proc_macro2::{Span, TokenStream};
 use quote::quote;
 use syn::{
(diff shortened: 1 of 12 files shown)


jhjourdan:
Interesting. What about using a macro like `result!()`?

jhjourdan:
Hum, I like the idea that you can bind pairs!

jhjourdan:
Ok, LGTM.

Lysxia:
> Interesting. What about using a macro like `result!()`?

I don't think it's worth the extra syntactic noise, but maybe we can use the idea for other "special variables" in the future.

jhjourdan:
Agreed.

PR #1976, opened 2026-03-11
Add message passing example, in a relaxed memory setting

lafeychine:


creusot-std/src/ghost/invariant.rs
```diff
@@ -315,6 +392,9 @@ impl<T: Protocol> AtomicInvariant<T> {
 #[opaque]
 pub struct NonAtomicInvariant<T: Protocol>(PhantomData<*mut T>);
 
+impl<T> !Send for NonAtomicInvariant<T> {}
```
jhjourdan: MAybe we already discussed that and I forgot something, bu I don't see why it would not be `Send`. I agree it should not be `Sync`, though.

PR #1977, opened 2026-03-11
"Options" version of the relacq message passing example.

jhjourdan:


jhjourdan:
I implemented another change in this PR: the various atomic operations do no longer return a `Ghost<A>`:

- The ghost closure can instead mutate its environment if needed.
- It is tedious to have a function that returns 2 values: calls need to destructure the pair, and specs need to consider `result.0` and `result.1`.

jhjourdan:
For some reason, we no longer need the `DEPTH` annotation now.

PR #1983, opened 2026-03-12
Hide `SyncView` constructor

Lysxia:


jhjourdan:
What's the point?

jhjourdan:
I mean, if this is required, then either the macro should enforce this, or it should add it automatically.

Lysxia:
It's not required but it makes the API cleaner to not expose unusable things.

PR #1985, opened 2026-03-13
Add other atomic types and update MP examples

lafeychine:


jhjourdan:
See https://stackoverflow.com/questions/33999341/generating-documentation-in-macros for fixing the problem with macros and doc.

jhjourdan:
While you are at it, you can add minimum support for `AtomicPtr`, which we will surely need for some concurrent algorithms.

Apart from that (and the fact it needs to be rebased), LGTM, thanks!

jhjourdan:
BTW, since we also want to use them for relaxed accesses, we should rename the file `atomic_relacq.rs` into `atomic_weak.rs`, `atomic_history.rs` or simply `atomic.rs`?

Issue #1940, opened 2026-02-20
Relax the ghost check for 'ghost in ghost' situations


dianegolfouse:
The following is forbidden by Creusot today:
```rust
ghost! {
    let x = Box::new(1i32);
    let g = ghost!(x);
};
```
```
error: cannot move a non-ghost variable into a `ghost!` block
```
This should be allowed though, as this:
```rust
ghost! {
    let x = Box::new(1i32);
    let g = Ghost::new(x);
};
```
is perfectly fine.

PR #1997, opened 2026-03-20
Proof modes

### LINKED ISSUE #1532: Mode for only verifying the absence of UB
There are different verification goals one may have, which result in different contracts. These are different "modes" we may want to support:

- Creusot's currently intended goal is to prevent *most* panics. For example, we allow the size overflow panic in `Vec::push`, because that's a PITA to deal with for little gain, but we forbid the out-of-bounds panic in `index`.
- Another choice is to only prevent UB, but with guarantees applicable even for unverified clients. In that case all safe functions must have a `true` precondition, including `panic!` which is considered safe in this setting. Only unsafe and private functions may have preconditions other than `true`.
- Going in the opposite direction, it's natural to imagine another mode where all panics are forbidden, even in `Vec::push`, which would have to have yet another contract.
- Generalizing further, we might as well leave it entirely up to the user to decide what behaviors to forbid, defining their own modes.

In terms of implementation, I currently think that "modes" would simply be a way to choose between different contracts. I wonder whether this should interact with `#[terminating]` and `#[pure]` attributes, or other features.

We can start with two modes to keep it simple: on top of the current "best effort" mode, I really want to add a "no UB, panics allowed" mode for the verify-rust-std challenge.

Lysxia:
Starting with a proposal for what they should look like, before implementing it.

Fixes #1532 (safety mode)  

Path towards #1957 (using termination mode), #1940 (using ghost mode)


xldenis:
what's an example use case for having the mode as a conclusion?

Lysxia:
> what's an example use case for having the mode as a conclusion?

Ghost-only functions would require ghost mode.

xldenis:
> Ghost-only functions would require ghost mode.

so `true => mode!(ghost)`? is there an example of a non-trivial precondition?

Lysxia:
It seems that closures need something a bit more sophisticated than what I've written so far: the mode of a closure may be independent of the function it's passed to.

guide/src/proof_modes.md
```diff
@@ -0,0 +1,84 @@
...
+Panicking is safe. We can `panic!` anywhere without causing
+undefined behavior (its precondition in safety mode is `true`) and of course,
+it won't return (its postcondition is `false`).
+
+#### Arithmetic
+
+On built-in integer types, arithmetic operations are also safe: they may either panic
+(in debug profiles) or perform wrapping operations (in release profiles).
+Thus, in safety mode the precondition of these operations is `true` and
+their postconditions are those of the wrapping operations.
```
Lysxia: A simpler solution is to keep the "won't overflow" precondition, because overflow of built-ins is undesirable basically always.

jhjourdan:
If we do that, we need to adapt the specifications of closures to take the mode as parameter.

jhjourdan:
We need to discuss that directly. I think there is a direct way to encode all that (including safety) using first-class Booleans and without disabling preconditions in the safety mode.

jhjourdan:
> It seems that closures need something a bit more sophisticated than what I've written so far: the mode of a closure may be independent of the function it's passed to.

Yes, and that is why the logical functions `precondition` and `postconditionxxx` should take an additional mode parameter.

guide/src/proof_modes.md
```diff
@@ -0,0 +1,84 @@
...
+Creusot introduces a "safety mode" (`mode!(safety)`) to prove the safety of **safe functions
+that contain unsafe blocks**. Indeed, there is a slight gap between the notion of safety and Creusot contracts.
+Safety is a property that must hold *for all inputs* of safe functions, whereas Creusot contracts
+often feature *nontrivial preconditions* that restrict the domain of inputs for which the function
+is intended to be useful.
+
+Hence, we follow different rules to determine verification conditions for proving
+safety than for proving adherence to a contract. In safety mode:
+
+1. We ignore the contract of the function to verify: we don't assume preconditions, and we don't assert postconditions.
```
Kixunil: Why not assert postconditions?

Lysxia: We already prove them in normal (non-safety) mode.

Kixunil: So the same function would be proven in both modes?

Lysxia: Yes

Lysxia:
My implementation of proof modes is getting close to finished (in a separate branch, `modes-wip`). The changes to the compiler are mostly complete, and most of the work left is to update tests and examples. I've updated this RFC to reflect the current syntax.

One big remaining task is to update proofs related to iterators, notably maps and filters as they involve closures, and we have to be careful about what mode gets passed to `precondition`/`postcondition_mut`.

There are now warnings about functions that Creusot can't prove safe due to too restrictive `requires` clauses. The following examples are affected:

- Union-find: `find` and most other operations have a precondition `uf.in_domain(elem)`.
- Persistent array: `reroot` has a precondition `pa.partial_invariant()`.

It's not clear how to handle them. Maybe we can just disable the warning for them and explain why in a comment.

Another use case to figure out comes from the slices case study. Higher-order functions like `binary_search_by` require that their closure argument is callable on elements of the slice. Creusot is going to warn about this precondition. It seems that `precondition` should always be true if `mode.nopanic()` is false. At the same time that sounds quite restrictive, but I can't quite put my finger on why.

I also had plans to add support for "ghost-only functions" (as opposed to `#[check(ghost)]` which is "ghost-callable functions") but that can be postponed to a later PR.

TODO:

- [ ] Update iterator proofs
- [ ] Safety proofs in examples
- [ ] Safety proofs in slices case study
- [ ] Termination of some instances of `next` (#1957)

Lysxia:
One more wrinkle to mention. I added syntax to bind the `mode` parameter also in `proof_assert!`, `invariant!`, and `variant!`, which is useful whenever `precondition`/`postcondition` are involved:

```rust
proof_assert!{ |mode| f.precondition(args, mode) }
```

Notably, this can't be confused with a `Mapping` because the type of an assertion is `bool`.

But a `snapshot!` can contain a `Mapping`, so we can't reuse the closure syntax to bind modes... One solution is to add a new macro, say `snapshot_with_mode!`.

jhjourdan:
> One more wrinkle to mention. I added syntax to bind the `mode` parameter also in `proof_assert!`, `invariant!`, and `variant!`, which is useful whenever `precondition`/`postcondition` are involved:
> 
> ```rust
> proof_assert!{ |mode| f.precondition(args, mode) }
> ```
> 
> Notably, this can't be confused with a `Mapping` because the type of an assertion is `bool`.
> 
> But a `snapshot!` can contain a `Mapping`, so we can't reuse the closure syntax to bind modes... One solution is to add a new macro, say `snapshot_with_mode!`.

What about using a special syntax in specs (invariants, assertions, snapshot macro, requires/ensures) to mention the current mode? Something like `mode!()`?

guide/src/proof_modes.md
```diff
@@ -0,0 +1,211 @@
...
+
+In the following example:
+- the precondition says that `f(x)` terminates if `0 <= x`,
+- the postcondition says that `f(x)` terminates only if `0 <= x` (because
+  the function may be called on any input with `mode!().terminates()` set to `false`,
+  and the postcondition holds whenever it terminates).
+
+```rust
+#[check(terminates)]
+#[requires(mode!().terminates() ==> 0 <= x)]
```
Kixunil: The text above mentions closure syntax but this doesn't use closure syntax.

guide/src/proof_modes.md
```diff
@@ -0,0 +1,211 @@
...
+    if x != 0 {
+        f((x - 1) / 2)
+    }
+}
+```
+
+In more details, the `#[check(terminates)]` attribute does the following:
+
+- it removes the implicit precondition `!mode!().terminates()`,
+- if the function is recursive, it must have a `#[variant]`
```
Kixunil, line 74: The function in example above is recursive but does not have `#[variant]`.

guide/src/proof_modes.md
```diff
@@ -0,0 +1,211 @@
...
+There is a slight gap between safety and the usual meaning of Creusot contracts.
+Safety is the property that a function avoids undefined behavior *for all inputs* of safe functions,
+whereas Creusot contracts often feature *nontrivial preconditions* that restrict the domain of inputs
+for which the function is intended to be useful.
+
+Proof modes enable you to write preconditions that become trivial (`true`) in the context of proving safety.
+
+The functions with safety proof obligations are **the safe functions that contain unsafe blocks**.
+For those functions, Creusot requires that their preconditions are syntactically of the following forms
+where `mode` is of course the mode argument that was just bound in the precondition
```
Kixunil: This phrase seems to be referring to the closure syntax?

guide/src/proof_modes.md
```diff
@@ -0,0 +1,211 @@
...
+and `$M` is one of `nopanic`, `terminates`, or `ghost`:
+
+- `!mode!().$M()`
+- `mode!().$M() ==> _`
+
```
Kixunil, line 122: It'd be helpful to have a full example. While I think I understand it, I'm not really certain I do and I'm not sure how it will interact with type invariants.

As I currently understand it, proving the implementation of `ops::Index<usize> for [T]` would look like this:

```rust
impl<T> Index<usize> for [T] {
    type Output = T;

    // is this correct?
    #[requires(mode!().nopanic() ==> index < self.len())]
    fn get(&self, index: usize) -> &Self::Output {
        assert!(index < self.len());
        unsafe { self.get_unchecked(index) }
    }
}
```

Another interesting example:

```rust
pub struct ArrayVec<T: Copy, const N: usize> {
    data: [MaybeUninit<T>; N],
    // invariant: len <= N and all items up to len are initialized
    len: usize,
}

impl<T: Copy, const N: usize> ArrayVec<T, N> {
    /// Pushes element; panics if the buffer is full.
    // is this correct?
    #[reqiures(mode!().nopanic() ==> self.len < N)]
    fn push(&mut self, element: T) {
        // the access checks self.len < N
        self.data[self.len] = MaybeUninit::new(element);
        self.len += 1;
    }

    // As I understand it, this doesn't need a precondition because the invariant alone proves this will not be UB?
    fn pop(&mut self) -> Option<T> {
        if self.len == 0 {
            return None;
        }
        self.len -= 1;
        Some(unsafe { self.data.get_unchecked(self.len).assume_init_read() })
    }
}

impl<T: Copy, const N: usize> Invariant for ArrayVec<T, N> {
    fn invariant(self) -> bool {
        self.len < N && self.elements_initialized() // I don't care to figure out right now how to express initializedness
    }
}
```

Kixunil (commented):
Disclaimer: I still consider myself a newbie in all this, so my review may be very stupid. Feel free to tell me to shut up. :)

One thing that makes this a bit confusing is that it looks like an effect system except inverted - mode `true` means that the function MUST NOT have some effect (panic, looping, ...)

I don't know if it can be improved but if this actually is equivalent to effect system (I'm not deep enough in language theory to say so) maybe just call it effect syatem and invert the meaning for `true` to mean "does have effect"?

Or maybe if the macro was called `required_mode!()` it would make the entire thing clearer? Sucks it's longer name though.

guide/src/proof_modes.md
Lysxia: Thanks, I recently switched the syntax so mentions of the "closure syntax" are outdated indeed.

guide/src/proof_modes.md
Lysxia: Yes that looks correct. I can add a concrete example like this.

jhjourdan:
> One thing that makes this a bit confusing is that it looks like an effect system except inverted - mode `true` means that the function MUST NOT have some effect (panic, looping, ...)
> 
> I don't know if it can be improved but if this actually is equivalent to effect system (I'm not deep enough in language theory to say so) maybe just call it effect syatem and invert the meaning for `true` to mean "does have effect"?

I think you have a point. We already discussed with @Lysxia and we agree that the name "proof mode" is not ideal, and I tend to agree that the polarity of Booleans would be better switched. I.e., we should have a Booelan `panic` to say that the panic effect is allowed, and a Boolean `diverges` to allow non-termination.

There are a few consequences I am not very comfortable with this change though, so I'm happy to take your input @Lysxia and @Kixunil:
- I am envisioning to use this structure for storing the available namspaces for Iris-style invariants. Basically, it's a set of names, and if a given name is in this set, then you are allowed to open invariants with this name. It's a bit far-fetched to consider that opening an invariant is an effect (is it?).
- The `ghost` Boolean is a bit weird, because it both allows and forbid things: when it is `true`, you are allowed to run ghost-only functions (like functions in `Seq`), but not things like `Vec::push`. So it may be a bit confusing (in particular, we cannot say like other Booleans that putting it to true allows more things).

What do you both think?

Issue #1671, opened 2025-08-27
Use new trait solver API


Lysxia:
From discussion at #1565, the way we currently use the trait solver via `codegen_select_candidate` is not ideal because (1) we lose information in case of ambiguity (2) we need access to the proof tree for sound termination checking (#1232). Using `ProofTreeVisitor` instead seems promising, but it's completely undocumented, so it's going to take some trial and error to figure out.

Lysxia:
Allegedly, just manually calling `codegen_select_candidate` recursively is a viable solution to reconstruct the proof tree. Source: https://rust-lang.zulipchat.com/#narrow/channel/144729-t-types/topic/Find.20impls.20used.20to.20solve.20a.20constraint/with/576047117

Also, the API (whether `ProofTreeVisitor` or `SelectionContext::select`) still makes no sense to me.

jhjourdan:
And what about ambiguity?

I think we have cases were a global instance competes with a local instance (from a trait constraint). In this case, the selector answers an error because of the ambiguity, while we may actually have a closed instance.

Also, if the solution is to call `codegen_select_candidate` recursively in the termination checker, we still need to do it! (Though I can understand you are bored with that.)

Lysxia:
The plan is still to fix the termination checker #1232, but keeping `codegen_select_candidate`.

It's true that the ambiguity issue remains.

Lysxia:
An example (reduced from #1565) where ambiguity results in picking an abstract instance over a concrete one, where picking the concrete one would enable proving an assertion:

```rust
extern crate creusot_std;
use creusot_std::prelude::*;

pub trait Tr {
    fn f(self) -> Self;
}

impl Tr for i32 {
    #[ensures(result == self)]
    fn f(self) -> Self {
        self
    }
}

pub fn g() where i32: Tr
{
    let r = 0i32.f();
    proof_assert! { r@ == 0 };
}
```

However one can also just remove the constraint on the concrete type. Are there less artificial examples of this issue?

jhjourdan:
TBH, I don't know. While I'm doing tests, it becomes less and less clear when the solver answers "ambiguous". I am fine with ignoring this problem until someone complains about it, if the problem with the termination checker is solved.

Lysxia:
At the moment, the only case of ambiguity that I'm aware of is between a trait impl and a redundant trait bound. In that case we can just pick the trait bound, which matches what we are currently doing already.

[...]

Issue #881, opened 2023-10-04
Soundness and empty types


jhjourdan:
Why3 assumes that every type is non-empty, and so do all the SMT solvers.

However, for several reasons, it seems like we need to support empty types:
- The "never" type (`!` in Rust), and friends like `Infaillible` are use in crucial places in the standard library (see the `Try` trait).
- Types with invariants may be empty if the invariant is absurd. In #880, we are in the process of removing the `is_inhabited` law, which guaranteed that types with invariants are not empty. The reasons are:
     - It is cumbersome to prove inhabitedness of type invariants, like Why3 does, (because many types do not have easy ways to exhibit inhabitants, both in Pearlite and in Why3 logic language)
     - Sometimes, invariants are purposely not provably uninhabited. This is the case of the `Map` iterator, which can be uninhabited if the pre and post-conditions of the closure do not match.

jhjourdan:
My plan for this issue is to:
[ ] Syntactically detect types that might be empty. Apart from `!`, these types are ADTs such that all the constructors have at least one recursive field. We need to be careful that empty types might be mutually recursive.
[ ] Translate these types to an ADT with one additional constructor with no field (i.e., we add a value to the type), and emit a warning.
[ ] Modify the structural type invariant of these types to ensure that no value of this type in *executable code* should be equal to this additional value.

Hence, in executable code, the values of such a type are exactly those of the type declared by the user, and the user can use the type invariant to leverage this in her reasoning. In the logic, on the contrary, we have added a constructor, so that we are sure that no type in the logic is uninhabited.

We need to be careful, though, with the syntactic criterion, especially when dealing with mutually recursive types or nested recursive types. Examples of types that need to be flagged as potentially empty:

```rust
struct A{ // Empty type
    x: B;
}

struct B { // Empty type
    y: Box<A>;
}

struct C { // Non-empy type, because in A has been added a value
   x : A;
}

struct D<T> {  // Non empty type
   x : Box<T>;
}

struct E { // Empty type
    x : D<E>;
}
```

jhjourdan:
Of course, @Lysxia, you are welcome to attack this issue. Please simply tell us that you are working on it.

jhjourdan:
I'm removing the soundness tag, as I don't think the current implementation is unsound wrt. this problem.

This is rather an "enhancement", since the current problem is that we don't support some empty algebraic types (Why3 will reject them.)

jhjourdan:
[...]

PR #2006, opened 2026-03-26
Termination check: handle recursion through proof trees and add recursive types check

### LINKED ISSUE #1232: Remaining unsoundness in the termination check

Lysxia:
Close #1232

- Handle recursion through proof trees (the original examples of #1232).
- Termination checking of impl methods must happen under bounds of the original trait method declaration (see [this comment from that same issue](https://github.com/creusot-rs/creusot/issues/1232#issuecomment-4014344872)).
- Add recursive types check:
    - Types can only recurse in "strictly positive" positions.
    - Recursive associated types are rejected (this also makes use of the new support for proof trees).
    - Recursive traits are rejected (detecting additional counterexamples in #1232).
 - New attributes
    - `#[trusted(ghost)]` (more fine grained than `#[trusted] #[check(ghost)]`, i.e., it still enables Coma translation of program functions)
    - `#[trusted(terminates)]` (idem, and can also be used on types)
    - `#[trusted(positive(T, U))]` to postulate "strictly positive" type arguments (can also be used in `extern_spec!`)
- Guide: expanded the Termination chapter with more details, recording the counterexamples found in #1232.

Some related things that are not handled in this PR:

- There remains a recursion check in the translation pass, which may reject types that pass the new recursive type check (see the commented out example in the guide).
- We'd like to use the new trait solver to better handle ambiguity, but AFAICT this only happens when a bound overlaps with an impl, in which case we should be able to pick the bound and do nothing more, which happens to match the current behavior, so we don't have a pressing reason to use the more "proper" entry point to the trait solver (#1671)
- We'd like to better detect empty recursive types (`struct Never(Box<Never>)`, #881)

Finally, I put traits in the same dependency graph as types, but there are currently no edges between traits and types. When checking traits, I only find dependencies to other traits in trait bounds (for instance, `trait Tr0 where T1: Tr2<T3>` has an edge `Tr0 -> Tr2`, and `T1` and `T3` are ignored). And when checking types, there are no edges due to bounds (in `struct T<A: Tr>(A)`, `Tr` is ignored; but we will track dependencies to trait impls whenever `T` is instantiated) or trait objects `dyn Tr` (because they are unsupported). Can you find a counterexample where that breaks soundness?


jhjourdan:
Ok, many thanks for that!

There is a lot of things to review, and I need to think about this, because there is a lot of potential for unsoundness.

Lysxia:
>  there are currently no edges between traits and types

I came up with an argument for that. You can split every trait in two: its associated types on one side, and its methods and supertrait constraints on the other side. Once associated type declarations are viewed as standalone constructs, they depend on nothing. That's why there is no need to track dependencies between ADTs and associated type declarations (which are different from associated trait *instances*, for which I do track dependencies).

The only remaining dependencies from types to traits are via trait objects, which are basically not supported at the moment.

jhjourdan:
Cannot associated types depend on associated types of subtraits?

Lysxia:
I don't see how.

Another way to think about it is that associated types are like extra type parameters (as they would actually be in Rocq), with differences only relevant to type inference.

```rust
trait Tr {
   type A;
}

struct S<T: Tr>(T::A);

// is like

trait Tr<A> {
}

struct S<A, T: Tr<A>>(A);
```

guide/src/termination.md
```diff
@@ -56,7 +86,201 @@ A function with the `check(terminates)` attribute cannot:
...
+There is some experimental support for allowing recursive occurrences through other type constructors.
+For example, the argument of `Option<_>` is strictly positive, allowing this type:
+
+```rust
+pub enum List<T>(T, Option<Box<List<T>>>);
+```
+
+The attribute `#[trusted(positive(T))]` can be used on a `struct` or `enum`
+to postulate that certain arguments are strictly positive.
+
```
jhjourdan, line 183:
```suggestion
The attribute `#[trusted(positive(T))]` can be used on a `struct` or `enum`
to postulate that certain arguments are strictly positive. Note that Creusot
does not check anything about this annotation (hence the use of the
`#[trusted` attribute). It is the user's responsibility to not use this annotation
on non-strictly positive types.
```

guide/src/termination.md
```diff
@@ -56,7 +86,201 @@ A function with the `check(terminates)` attribute cannot:
...
+struct Dummy;
+
+impl P for Dummy {
+    // This associated type is recursive through the instantiation of `B`.
+    type A = Mapping<B<Dummy>, bool>;
+}
+```
+
+If this were accepted by Creusot, the associated type `Dummy::A` would be equivalent to
+`Mapping<Dummy::A, bool>`, which would be unsound as explained in a [previous counterexample](#counterexample).
```
jhjourdan: The link is ambiguous.

creusot-std/src/cell/predcell.rs
```diff
@@ -7,6 +7,7 @@ use crate::{logic::Mapping, prelude::*};
 /// Cell over predicates
 ///
 /// A wrapper around `std::cell::Cell` that allows predicates to be used to specify the contents of the cell.
+#[trusted(positive(T))]
```
jhjourdan: I'm really not sure for this one. A `PredCell<T>` is essentially a predicate over `T`, so it is essentially negative. I really suspect one can build a contradiction using this.

creusot/src/translation/specification.rs
```diff
@@ -17,6 +20,49 @@ use rustc_span::{DUMMY_SP, Span};
...
+    Impure,
+    /// Terminates (but might panic, e.g., `Vec` overflow)
+    Terminates,
+    /// Callable in ghost context (must terminate successfully)
+    Ghost,
+}
+
+impl ProgramPurity {
+    pub fn is_ghost(self) -> bool {
+        self == ProgramPurity::Ghost
```
jhjourdan:
```suggestion
        self >= ProgramPurity::Ghost
```

(More robust wrt. maintainance.)

creusot/src/translation/traits.rs
```diff
@@ -191,6 +191,155 @@ pub(crate) fn evaluate_additional_predicates<'tcx>(
...
+        // returns an error, while we would like it to return any of the instances.
+        // We need to find another entry point of the trait solver.
+        // In the meantime, pretend that we have an instance that we do not know
+        Err(CodegenObligationError::Ambiguity) => return UnknownFound,
+        Err(_) => return None,
+        Ok(source) => Found(source),
+    }
+}
+
+pub fn select_method<'tcx>(
```
jhjourdan:
```suggestion
fn select_method<'tcx>(
```

creusot/src/translation/traits.rs
jhjourdan, line 1: This file is containing two mostly independent things:

- Generating refinement logical formulas for impls. This really belong to the frontend of Creusot (module `translation`).
- Resolution logic, which is both used in the frontend and the backend.

So I think they should belong to two different modules, because they don't have anything to do with each other. Could you move the resolution logic to another file? Perhaps at the root of the `creusot` crate?

creusot/src/validate/recursive_types.rs
```diff
@@ -0,0 +1,515 @@
...
+
+fn walk_adt<'tcx, E>(
+    ctx: &TranslationCtx<'tcx>,
+    def_id: DefId,
+    mut f: impl for<'a> FnMut(TypeNode, &'a [(ty::Ty<'tcx>, Option<usize>)], Span) -> Result<(), E>,
+) -> Result<(), E> {
+    let def = ctx.adt_def(def_id);
+    let typing_env = ctx.typing_env(def.did());
+    let mut path = Vec::new();
+    for fdef in def.variants().iter().flat_map(|vdef| vdef.fields.iter()) {
```
jhjourdan: There is `AdtDeff::all_fields` for that.

creusot/src/validate/recursive_types.rs
```diff
@@ -0,0 +1,515 @@
...
+                Unit | Empty | Snapshot(_) | Box(_) | Identity(_) => Ok(()),
+                Enum | Struct { partially_opaque: false } => {
+                    let index = index.unwrap();
+                    if positivity.get(ctx, def.did(), index) {
+                        Ok(())
+                    } else {
+                        Err(NotStrictlyPositive(def.did(), index))
+                    }
+                }
+                Namespace => unreachable!(),
```
jhjourdan:
```suggestion
                Opaque { .. } | Struct { partially_opaque: true } | Builtin(_) | Identity(_) => {
                    if ctx.trusted_positivity(def.did(), index.unwrap()) {
                        Ok(())
                    } else {
                        Err(Abstract(def.did()))
                    }
                }
                Snapshot(_) | Box(_) => Ok(()),
                Enum | Struct { partially_opaque: false } => {
                    let index = index.unwrap();
                    if positivity.get(ctx, def.did(), index) {
                        Ok(())
                    } else {
                        Err(NotStrictlyPositive(def.did(), index))
                    }
                }
                Unit | Empty | Namespace => unreachable!(),
```

`Unit` and `Empty` should not appear here, because they don't have type parameters, and `Identity` is essentially a builtin.

creusot/src/validate/recursive_types.rs
```diff
@@ -0,0 +1,515 @@
...
+            } else {
+                Ok(())
+            }
+        }),
+        TraitImpl(_) => Err((RecursiveAssocType, node)),
+    }
+}
+
+/// Check whether the given `path` forbids recursion,
+/// using visibility relative to `type_id`.
```
jhjourdan: (Disclaimer: perhaps we already discussed this IRL, but I don't remember. I might also have changed my mind in between. If that is the case, then tell me and please excuse me.)

Do we need to make positivity relative to visibility? Cannot we follow Rust's way of doing things for e.g., variance (i.e., variance is determined by taking all the fields into account, including private ones)? Then, just like variance, positivity is an implicit part of the public interface of a type, that somehow "leaks" from its implementation. This may even make it possible to get rid of (most of uses of) `#[trusted(positive)]`, but asking users to annotate their opaque types with `PhantomData<Mapping<T, bool>>` if they want their type to be non-positive.

creusot/src/validate/recursive_types.rs
```diff
@@ -0,0 +1,515 @@
...
+                    self.0.remove(index);
+                }
+                _ => t.super_visit_with(self),
+            }
+        }
+    }
+
+    let generics = ctx.generics_of(def_id);
+    let mut positivity = ArgPositivity(DenseBitSet::new_filled(generics.count()));
+    for field in ctx.adt_def(def_id).variants().iter().flat_map(|vdef| vdef.fields.iter()) {
```
jhjourdan:
```suggestion
    for field in ctx.adt_def(def_id).all_fields() {
```

creusot/src/validate/recursive_types.rs
```diff
@@ -0,0 +1,515 @@
...
+    fn did(self) -> DefId {
+        use TypeNode::*;
+        match self {
+            Type(did) | Trait(did) | TraitImpl(did) => did,
+        }
+    }
+}
+
+/// Edges for `TypeGraph`
+/// Metainformation about the kind of dependency this edge represents.
```
jhjourdan, line 320:
```suggestion
/// Metainformation about the kind of dependency this edge represents.
/// Only used for emitting error messages.
```

It's important to document that, because we remember only one edge kind between two nodes.

creusot/src/validate/recursive_types.rs
```diff
@@ -0,0 +1,515 @@
...
+            add_trait_impl(ctx, impl_id.to_def_id(), &mut graph);
+        }
+    }
+
+    graph
+}
+
+fn add_type(ctx: &TranslationCtx, def_id: DefId, graph: &mut TypeGraph) {
+    if is_opaque(ctx.tcx, def_id)
+        || is_trusted(ctx.tcx, def_id)
```
jhjourdan:
```suggestion
        || is_trusted_item(ctx.tcx, def_id)
```

Essentially, we want types in `#[trusted]` modules to behave like of they wer trusted directly.

creusot/src/validate/recursive_types.rs
```diff
@@ -0,0 +1,515 @@
...
+        } else {
+            TypeEdge::AdtDef(span)
+        };
+        Ok::<(), !>(graph.add_edge(node, tgt, edge))
+    })
+    .unwrap()
+}
+
+fn add_trait(ctx: &TranslationCtx, def_id: DefId, graph: &mut TypeGraph) {
+    if is_trusted(ctx.tcx, def_id) {
```
jhjourdan:
```suggestion
    if is_trusted_item(ctx.tcx, ctx.parent(def_id)) {
```

A trait annotated with `#[trusted]` means that its implementations should be marked with `#[trusted]`, not that its content is trusted or anything. So here we are only interested in whether the parent module is trusted.

AFAIK, there is no need to trust a trait declaration, so this should be enough.

creusot/src/validate/recursive_types.rs
```diff
@@ -0,0 +1,515 @@
...
+            TypeEdge::AdtDefBound(span, adt.did())
+        } else {
+            TypeEdge::AdtDef(span)
+        };
+        Ok::<(), !>(graph.add_edge(node, tgt, edge))
+    })
+    .unwrap()
+}
+
+fn add_trait(ctx: &TranslationCtx, def_id: DefId, graph: &mut TypeGraph) {
```
jhjourdan, line 386: Document the fact that we are not recording here any trait -> type or trait -> impl dependency, and we think this is fine because xxxx.

creusot/src/validate/recursive_types.rs
```diff
@@ -0,0 +1,515 @@
...
+                    node,
+                    TypeNode::Trait(predicate.trait_ref.def_id),
+                    TypeEdge::FnBound(item.def_id, span),
+                );
+            }
+        }
+    }
+}
+
+fn add_trait_impl(ctx: &TranslationCtx, def_id: DefId, graph: &mut TypeGraph) {
```
jhjourdan, line 421: We are not recording dependencies to other trait impl that are required to prove e.g., subtrait predicates. I hope this is fine...

creusot/src/validate/terminates.rs
```diff
@@ -772,3 +790,112 @@ fn is_structurally_recursive<'tcx>(ctx: &TranslationCtx<'tcx>, self_id: DefId) -
...
     s.valid()
 }
+
+fn as_predicates<'tcx>(
+    tcx: TyCtxt<'tcx>,
+    clauses: impl IntoIterator<Item = ty::Clause<'tcx>>,
+) -> impl Iterator<Item = ty::TraitRef<'tcx>> {
+    clauses.into_iter().filter_map(move |clause| {
+        // WARNING: `clause` is not normalized (but we don't seem to use the fact that it should be)
+        let Some(clause) = clause.as_trait_clause() else { return None };
```
jhjourdan:
```suggestion
        let clause = clause.as_trait_clause()?;
```

creusot/src/validate/terminates.rs
```diff
@@ -772,3 +790,112 @@ fn is_structurally_recursive<'tcx>(ctx: &TranslationCtx<'tcx>, self_id: DefId) -
 
     s.valid()
 }
+
+fn as_predicates<'tcx>(
```
jhjourdan:
```suggestion
fn trait_refs_of_clauses<'tcx>(
```

creusot/src/validate/terminates.rs
```diff
@@ -772,3 +790,112 @@ fn is_structurally_recursive<'tcx>(ctx: &TranslationCtx<'tcx>, self_id: DefId) -
...
+    clauses: impl IntoIterator<Item = ty::Clause<'tcx>>,
+) -> impl Iterator<Item = ty::TraitRef<'tcx>> {
+    clauses.into_iter().filter_map(move |clause| {
+        // WARNING: `clause` is not normalized (but we don't seem to use the fact that it should be)
+        let Some(clause) = clause.as_trait_clause() else { return None };
+        let ty::PredicatePolarity::Positive = clause.polarity() else { return None };
+        let trait_ref = tcx.instantiate_bound_regions_with_erased(clause).trait_ref;
+        if !matches!(tcx.def_kind(trait_ref.def_id), DefKind::Trait) {
+            // Some bounds can be trait aliases and we skip them.
+            // The implied bounds should already appear separately.
```
jhjourdan, line 801: I've never understood whether this is indeed the case. Do we have a test for that?

(I know that this is unrelated to this PR, but still, that's good to check that now.)

creusot/src/validate/terminates.rs
```diff
@@ -772,3 +790,112 @@ fn is_structurally_recursive<'tcx>(ctx: &TranslationCtx<'tcx>, self_id: DefId) -
...
+    typing_env: TypingEnv<'tcx>,
+    clauses: impl IntoIterator<Item = ty::Clause<'tcx>>,
+) -> Vec<DefId> {
+    let mut nodes = Vec::new();
+    let mut predicates: Vec<_> = as_predicates(tcx, clauses).collect();
+    while let Some(trait_ref) = predicates.pop() {
+        let ImplSelection::Found(source) = select_trait_impl(tcx, typing_env, trait_ref) else {
+            continue;
+        };
+        let ImplSource::UserDefined(source) = source else { continue };
```
jhjourdan, line 821: So in the end you decided to not use the new ProofTree API?

This seems good like this, apart from the fact that I'm afraid that we will miss some dependencies due to builtin impls.

For now, I don't this this should be a problem, because Creusot does not handle these builtin impls anyway (see #2094). But with this implementation technique, we **are** missing dependencies introduced by impls for `Clone`.

For a list of them (for `Clone` only), see `instantiate_constituent_tys_for_copy_clone_trait` in rustc source. I have not found any other builtin trait that could cause this issue (but there might be some, including in the future).

creusot/src/validate/terminates.rs
```diff
@@ -772,3 +790,112 @@ fn is_structurally_recursive<'tcx>(ctx: &TranslationCtx<'tcx>, self_id: DefId) -
...
+
+/// For a method in a trait impl (`impl_item_id`), construct a `ParamEnv` that
+/// combines bounds from the parent impl and bounds from the original declaration (`trait_item_id`)
+fn param_env_for_termination(tcx: TyCtxt, trait_item_id: DefId, impl_item_id: DefId) -> ParamEnv {
+    let impl_id = tcx.impl_of_assoc(impl_item_id).unwrap();
+    let trait_ref = tcx.impl_trait_ref(impl_id).instantiate_identity().skip_normalization();
+    let trait_item_args = GenericArgs::identity_for_item(tcx, trait_item_id);
+    let args = tcx.mk_args_from_iter(
+        trait_ref.args.iter().chain(trait_item_args.iter().skip(trait_ref.args.len())),
+    );
```
jhjourdan: That's exactly `rebase_onto`.

creusot/src/validate/terminates.rs
```diff
@@ -772,3 +790,112 @@ fn is_structurally_recursive<'tcx>(ctx: &TranslationCtx<'tcx>, self_id: DefId) -
...
+
+/// Find path from src to tgt with at least one edge.
+fn find_path_from<N, E>(
+    graph: &graph::DiGraph<N, E>,
+    scc: &HashSet<graph::NodeIndex>,
+    src: graph::NodeIndex,
+    tgt: graph::NodeIndex,
+    path: &mut Vec<graph::NodeIndex>,
+    visited: &mut HashSet<graph::NodeIndex>,
+) -> Result<(), ()> {
```
jhjourdan: It's a bit weird to use `Result` for that, because the error case is actually not an error, it is what we are searching for.

jhjourdan (approved):
That's very good work, which solves many difficult problems in Creusot. Thanks.

Why did you not use the proof tree API from rustc, finally?

Do we agree that this does not fully addresses #1232, since it does not solve the problem with default functions?

creusot/src/validate/recursive_types.rs
Lysxia: The initial idea was indeed to not expose positivity of abstract types, but I don't mind changing that.

creusot/src/validate/terminates.rs
Lysxia: I had already done this before our discussion with lcnr that we should really (re)try using the proof tree API.

creusot/src/validate/terminates.rs
Lysxia: I'll add a test for it. (`should_fail/terminates/implied_bounds.rs`)

creusot/src/validate/terminates.rs
Lysxia: I'll add a `FIXME` comment to remember handling builtin impls

Lysxia:
> Why did you not use the proof tree API from rustc, finally?

I had already done this before our discussion with lcnr where we decided that we should really (re)try using the proof tree API.

> Do we agree that this does not fully addresses https://github.com/creusot-rs/creusot/issues/1232, since it does not solve the problem with default functions?

That's right. I'll open another issue to track that.

creusot/src/validate/recursive_types.rs
Lysxia: I changed it so positivity no longer depends on visibility.

creusot/src/translation/traits.rs
Lysxia: Split resolution logic into `resolution.rs`

creusot/src/backend/ty.rs
```diff
@@ -37,6 +37,24 @@ pub(crate) fn classify_adt<'tcx>(
     scope: DefId,
     def: AdtDef<'tcx>,
     subst: GenericArgsRef<'tcx>,
+) -> AdtKind<'tcx> {
+    classify_adt_(ctx, Some(scope), def, subst)
+}
+
+/// Never returns `Opaque { always: true }`, and all non-empty structs are `Struct { partially_opaque: false }`.
```
jhjourdan:
```suggestion
/// Never returns `Opaque { always: false }`, and all non-empty structs are `Struct { partially_opaque: false }`.
```

creusot/src/backend/ty.rs
```diff
@@ -37,6 +37,24 @@ pub(crate) fn classify_adt<'tcx>(
...
+pub(crate) fn classify_adt_unscoped<'tcx>(
+    ctx: &TranslationCtx<'tcx>,
+    def: AdtDef<'tcx>,
+    subst: GenericArgsRef<'tcx>,
+) -> AdtKind<'tcx> {
+    classify_adt_(ctx, None, def, subst)
```
jhjourdan: Isn't it directly equivalent to `classify_adt_(ctx, Some(def.did()), def, subst)`?

I.e., we consider that visibility is wrt. the struct itself. This seems both shorter and somewhat easier to think about.

jhjourdan: (And then you no longer need `classify_adt_`.

Lysxia: You're right!

jhjourdan:
FWIW, I'm OK to merge this as soon as you are ready.

Lysxia:
Merging!

PR #2016, opened 2026-04-01
Make `Tokens::split` takes full ownership of `self`

dianegolfouse:


jhjourdan:
Don't we need both?

dianegolfouse:
I don't think so, if we need the `&mut` one we can just call `reborrow` before splitting.

jhjourdan:
Ok, fair enough.

PR #2021, opened 2026-04-07
Add `sub_logic` law: `ptr - (ptr + ofs) == -ofs`

Lysxia:


jhjourdan:
That's implied by the other post-cond and associativity of offset, right?

Lysxia:
Yes, they are equivalent, but as always solvers have trouble instantiating it.

Issue #1299, opened 2024-12-10
Panic in `clone_map.rs`


Lysxia:
```rs
extern crate creusot_contracts;
use creusot_contracts::*;

#[variant(i)]
pub fn count_paths(i: usize) {
    (0..i).map(|j| count_paths(j));
}
```

```
thread 'rustc' panicked at creusot/src/backend/clone_map.rs:383:17:
assertion `left == right` failed
  left: 2
 right: 1
```

The mentioned line:

https://github.com/creusot-rs/creusot/blob/739b2d0ecb8103f7fba017c0b09346954951bcb4/creusot/src/backend/clone_map.rs#L381-L386

jhjourdan:
Ok, the error message is not nice. But the problem here is that this is essentially mutually recursive functions (closure / function), which we do not support.

arnaudgolfouse:
> Ok, the error message is not nice. But the problem here is that this is essentially mutually recursive functions (closure / function), which we do not support.

But we do (in programs)? E.g.
```
pub fn count_paths(i: usize) {
    (0..i).map(|j| closure(j));
}

fn closure(j: usize) { count_paths(j) }
```
Works. The error is caused by the presence of a closure.

jhjourdan:
Right. We do not support it in the case the mutual recursion is between a closure and the function which contains it. That's clearly a limitation we need to work on.

One issue is that Coma would not support (yet?) spec inference in this case.

PR #2026, opened 2026-04-09
Streamline tracking of dependencies and allow recursive calls in closures

Lysxia:
Previously, dependencies started as `dep_set: IndexSet` before being converted into the `expansion_queue`.

Now, that logic is gathered in a single `DepGraphBuilder` structure to directly build the final graph.
Thanks to that, `update_graph` is just a small loop now.

A further simplification is that `expansion_queue` now only contains unvisited vertices instead of edges,
since the edges are added to the graph immediately.


xldenis:
> A further simplification is that expansion_queue now only contains unvisited vertices instead of edges,
since the edges are added to the graph immediately.

that always felt like a failure when I had to move the expansion queue to track edges. I had originally wanted to get it working with just nodes, so I'm glad you've managed to fix the bugs that prevented that!

jhjourdan:
Generally, this looks good to me.

However, I think we can fix #1299 along the way, and it might be very well the case that the fix would be a *simplification* of Creusot.

In #1299, I said that the problem was that Why3 was not compatible with that. I think I was wrong: in program code, we should consider the source program function just like any other function (ignoring the fact that it is recursive), and, in the names map, we do not associate the name of the source function with the generated body. Then, a recursive call (in a closure or not) will generate a stub for the source function, with the right specification, and therefore we will generate the right VCs. We translate a recursive function to a no recursive one, but who cares?

Should we do something similar for logic functions, to support polymorphic recursion, and even perhaps mutual recursion?

Lysxia:
That does sound feasible! It would require tweaking the termination checker, so maybe #2006 should be merged before that happens.

jhjourdan:
Are you trying to `Lazy.force` my review of #2006 ? 😉

I'll do that as soon as the deadline chain that I currently have comes at its end.

Lysxia:
Working on fixing #1299, I just need to give different names to the function and its recursive copy.

Lysxia:
This is trickier than it seemed because the closure must know the variant to be able to assert that the variant decreases in recursive calls. My plan is to make the coma function corresponding to the closure take the variant as an argument. This change should have little impact if we restrict this change to `check(terminates)` functions with a variant.

Lysxia:
cc @jhjourdan 

It seems that to allow closures to make recursive calls, they must capture the variant. As fixing #1299 is getting more and more involved, let's postpone it and only keep the original refactor in this PR.

jhjourdan:
Cannot we have a partial fix for functions which are not `#[terminates]`?

jhjourdan:
This already covers a lot of cases, because in most of the cases of program code, we do only partial termination.

jhjourdan:
The bottom line being: we do not support proving termination of mutually recursive functions. Recursive calls from a closure is essentially the same thing as mutual recursion. When we will support termination of mutually recursive functions, it should be easy to port it to recusive calls in closures.

Lysxia:
Ah yes, I can try to handle the case without `terminates`.

jhjourdan:
I'm ready to merge, if you are.

PR #2033, opened 2026-04-13
Read `CREUSOT_STD` environment variable in `patch_dep` function

Halbaroth:
The path returned by `creusot_std_path` is incorrect when using a Nix installation. It seems to be a hardcoded path pointing to the location of the library in the build directory of Nix.

`cargo creusot new` works as expected by reading the `CREUSOT_STD` environment variable. This commit propagates this workaround to `cargo creusot init`.


lafeychine (approved):
This proposal makes sense to me.

Do you have any other suggestions @Lysxia?

Lysxia (commented):
Good catch! `cargo_std` should be `creusot_std` (in the code and in the commit).

PR #2035, opened 2026-04-13
Add Alt-Ergo Free in the flake

Halbaroth:
We released Alt-Ergo Free 2.4.3 under the CeCILL-C license a few months ago. This commit replaces the unfree version with the free one.

Now, it is possible to run `nix develop` to enter the Nix shell without using the --impure flag!


jhjourdan:
I am right that this implies that the flake and the standard installation process will use a different alt-ergo version? I would prefer if it's not the case.

Perhaps the simplest would be to have a flag that one could pass to both installation scripts to choose between the recent solvers and the free solvers. What do you think? Do you think you could implement that?

jhjourdan:
BTW, CI is failing for a reason that seem related to Nix.

Halbaroth:
A simple solution is to switch to Alt-Ergo free everywhere. We can add an `unfree` option to `creusot-install` and an unfree devShell in the flake. 

To install the free version:
```console
nix shell "github:creusot-rs/creusot" 
```
and the unfree one:
```console
NIX_ALLOW_UNFREE=1 nix shell "github:creusot-rs/creusot#unfree" --impure
```

> BTW, CI is failing for a reason that seem related to Nix.

The CI failed because it is looking for the version of Alt-Ergo derivation but the name of the derivation is `alt-ergo-free`.

jhjourdan:
That sounds like an option.

Does the test suite passes with alt-ergo 4.3.2?

Is alt-ergo 4.3.2 free of known soundness bug?


> > BTW, CI is failing for a reason that seem related to Nix.
> 
> The CI failed because it is looking for the version of Alt-Ergo derivation but the name of the derivation is `alt-ergo-free`.

Then can you fix the CI script for using the right derivation name?

Thanks!!

Halbaroth:
Good question, Alt-Ergo free is not free of known soundness bugs because fixing soundness bugs is a part of our commercial offering. There are two such bugs:
1. In the FPA theory https://github.com/OCamlPro/alt-ergo/pull/1113. 
2. The `distinct` definition of Alt-Ergo was wrong https://github.com/OCamlPro/alt-ergo/pull/890. This bug is easy to trigger :/

I have an alternative suggestion. We can add a free devShell in the flake and run 
```
nix develop .#free
```
or 
```
nix shell "github:...#free 
``` 

We don't touch to creusot-install.

flake.nix
```diff
@@ -88,6 +103,46 @@
...
+        pkgs.mkShell {
+          inputsFrom = [creusot];
+          packages = [why3Framework rust.toolchain.dev];
+
+          CREUSOT_DATA_HOME = why3Framework;
+          LD_LIBRARY_PATH = pkgs.lib.makeLibraryPath [rust.toolchain.dev];
+          DYLD_FALLBACK_LIBRARY_PATH = pkgs.lib.makeLibraryPath [rust.toolchain.dev];
+
+          passthru = rust.toolchain.dev.passthru.availableComponents;
+        };
```
lafeychine: I would move this function close to the definition of `devShells`

flake.nix
```diff
@@ -207,15 +240,12 @@
...
-        inputsFrom = [packages.creusot];
-        packages = [packages.why3Framework rust.toolchain.dev];
-
-        CREUSOT_DATA_HOME = packages.why3Framework;
-        LD_LIBRARY_PATH = pkgs.lib.makeLibraryPath [rust.toolchain.dev];
-        DYLD_FALLBACK_LIBRARY_PATH = pkgs.lib.makeLibraryPath [rust.toolchain.dev];
-
-        passthru = rust.toolchain.dev.passthru.availableComponents;
+      devShells = {
+        default = mkShell {creusot = packages.creusot;};
```
lafeychine, line 237: Could you run `nix fmt .`?

flake.nix
```diff
@@ -23,7 +23,17 @@
...
+      overlays = [
+        rust-overlay.overlays.default
+        (super: self: {
+          ocamlPackages = self.ocamlPackages.overrideScope (
+            final: _: {
+              ocplib-simplex_0_4 = final.callPackage ./nix/pkgs/ocplib-simplex.nix {};
+            }
+          );
+          alt-ergo-free = self.callPackage ./nix/pkgs/alt-ergo-free.nix {};
+        })
```
lafeychine: I would avoid creating an overlay, for future maintainability reasons

nix/alt-ergo-free.nix
```diff
@@ -0,0 +1,13 @@
+{
```
lafeychine, line 1: I would put the contents of the `pkgs` directory in this file, similarly to the `cvc4.nix` file

nix/pkgs/alt-ergo-free.nix
```diff
@@ -0,0 +1,39 @@
...
+    ocplib-simplex_0_4
+    zarith
+    seq
+  ];
+
+  meta = {
+    description = "High-performance theorem prover and SMT solver";
+    homepage = "https://alt-ergo.ocamlpro.com/";
+    license = lib.licenses.cecill-c;
+    maintainers = with lib.maintainers; [redianthus];
```
lafeychine: Would be `redianthus` the maintainer of this file in Creusot?
If not, I wouldn't have included any maintainers here

nix/pkgs/ocplib-simplex.nix
```diff
@@ -0,0 +1,40 @@
+{
+  lib,
+  fetchFromGitHub,
+  stdenv,
+  autoreconfHook,
+  ocaml,
+  findlib,
+}:
+stdenv.mkDerivation (finalAttrs: {
```
lafeychine: For maintainability reasons, I would have use `rec {` instead of `finalAttrs`

nix/pkgs/ocplib-simplex.nix
```diff
@@ -0,0 +1,40 @@
...
+    findlib
+  ];
+
+  preInstall = ''
+    mkdir -p $out/lib/ocaml/${ocaml.version}/site-lib/
+  '';
+
+  postInstall = ''
+    mv $out/lib/${finalAttrs.pname} $out/lib/ocaml/${ocaml.version}/site-lib
+  '';
```
lafeychine: Is it not possible to use what is present in [`nixpkgs`](https://github.com/NixOS/nixpkgs/blob/a13cdfe520d87db401dd000fbd67cad728162a60/pkgs/development/ocaml-modules/ocplib-simplex/def)?

nix/pkgs/ocplib-simplex.nix
```diff
@@ -0,0 +1,40 @@
...
+stdenv.mkDerivation (finalAttrs: {
+  pname = "ocplib-simplex";
+  version = "0.4.1";
+
+  src = fetchFromGitHub {
+    owner = "OCamlPro";
+    repo = finalAttrs.pname;
+    rev = "v${finalAttrs.version}";
+    hash = "sha256-bhlTBpJg031x2lUjwuVrhQgOGmDLW/+0naN8wRjv6i4=";
+  };
```
lafeychine: Since the only modification is the version used, is it not possible to use `overrideAttrs` instead, as in `z3.nix`?

flake.nix
```diff
@@ -88,6 +103,46 @@
 
         envBuilder = rust.lib.overrideToolchain rust.toolchain.build;
       };
+      mkWhy3Framework = {isFree ? false}: let
```
lafeychine, line 97: Just a tiny linebreak:
```suggestion
      };

      mkWhy3Framework = {isFree ? false}: let
```

flake.nix
```diff
@@ -88,6 +103,46 @@
...
+        free-solvers = with lib.pkgs; [cvc4 cvc5 why3 why3find z3];
+        solvers =
+          if isFree
```
lafeychine, line 101: Just a tiny `let in` for `solvers`:
```suggestion
        solvers = let
          free-solvers = with lib.pkgs; [cvc4 cvc5 why3 why3find z3];
        in
```

lafeychine (changes requested):
Thank you very much for the PR. I have some remarks, but looks really great overall.

I hope that the Nix file wasn't too hard to read!

If you no longer have time to spend on this, I can finish it on top of your work.

nix/pkgs/alt-ergo-free.nix
Halbaroth: I just a copy-past of https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/by-name/al/alt-ergo-free/package.nix 
I removed the mention.

nix/pkgs/ocplib-simplex.nix
Halbaroth: By sheer of curiosity, can you explain? I used `rec` in my first submission on nixpkgs but the maintainer asked me to replace it with `finalAttrs` for the very same reason. I thought that `finalAttrs` was helpful to debug infinity loops?

nix/pkgs/ocplib-simplex.nix
Halbaroth: We cannot use ocplib-simplex 0.5 because this release introduced breaking changes in the API. I released ocplib-simplex 0.4 on master: https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/ocaml-modules/ocplib-simplex/0_4.nix

nix/pkgs/ocplib-simplex.nix
Halbaroth: This is not the only modification. ocplib-simplex 0.5 uses dune as build system but ocplib-simplex 0.4 used autotools.

nix/pkgs/ocplib-simplex.nix
Halbaroth: Oh, okay you want to use the old release of ocplib-simplex. I didn't know it has been released but I wrote a new version. We can reuse the old one if you prefer.

nix/pkgs/ocplib-simplex.nix
lafeychine: > I thought that finalAttrs was helpful to debug infinity loops?

It could be, I do not know.

The reason here is that `mkDerivation` is commonly seen without this extra argument.
Someone with little knowledge in Nix could more easily understand `rec` than `finalAttrs`

lafeychine (approved):
LGTM

jhjourdan:
I guess I'm going to trust you Nix guys...

jhjourdan (approved):

PR #2041, opened 2026-04-15
Use unique identifier for `--package` option of cargo creusot

Halbaroth:
The option `--package` of `cargo build` scans all the dependencies of the current crate and features. If there are several matches, it emits an error. For instance, in hashbrown crate:
```console
cargo creusot
```
prints
```
error: specification `hashbrown` is ambiguous
```
as hashbrown appears as a dependency (of creusot-std) and the current crate itself.

This commit prints the unique identifier `packages.id` of Cargo_metadata.


jhjourdan:
Thanks!

PR #2043, opened 2026-04-15
Fix ICE when normalizing types

### LINKED ISSUE #2040: ICE MRE: Failed to normalize Alias(Projection) for DeepModel during derive(PartialEq) on struct containing custom enum
Sorry to bother you again and appreciate your time
Here's the Minimal Reproducible Example:

### Summary
Creusot panics with an ICE when a struct/enum contains a field of a custom enum type that does not implement `DeepModel`. The panic is triggered during the standard Rust `derive(PartialEq)` macro expansion for the outer type.

### Minimal Reproduction
cargo.toml
[package]
name = "creusot-ice-test"
version = "0.1.0"
edition = "2021"

[dependencies]
creusot-std = "0.10.0"

src/lib.rs
```rust
use creusot_std::prelude::*;
use std::clone::Clone;
use std::cmp::PartialEq;
use std::cmp::Eq;

#[derive(Clone, Copy, PartialEq, Eq)]
pub enum Inner { A, B }

#[derive(Clone, Copy, PartialEq, Eq)]
pub enum Outer { Bar { inner: Inner } }

#[cfg(creusot)]
impl Outer {
    #[ensures(match (*self) {
        Outer::Bar { inner, .. } => result == (match inner {
            Inner::A => true,
            Inner::B => false,
        }),
    })]
    pub fn test_nested_match(&self) -> bool {
        match self {
            Outer::Bar { inner } => match *inner {
                Inner::A => true,
                Inner::B => false,
            },
        }
    }
}
```

### Steps to Reproduce
1. Create a new crate with `creusot-contracts` as a dependency.
2. Paste the code above into `src/lib.rs`.
3. Run `cargo creusot prove`.

### Exact Error Output

```text
error[E0277]: the trait bound `Inner: creusot_std::model::DeepModel` is not satisfied
 --> src/lib.rs:8:24
  |
7 | #[derive(Clone, Copy, PartialEq, Eq)]
  |                       --------- in this derive macro expansion
8 | pub enum Outer { Bar { inner: Inner } }
  |                        ^^^^^^^^^^^^ unsatisfied trait bound
  |
help: the trait `creusot_std::model::DeepModel` is not implemented for `Inner`
 --> src/lib.rs:6:1
  |
6 | pub enum Inner { A, B }
  | ^^^^^^^^^^^^^^
  = note: required for `&Inner` to implement `creusot_std::model::DeepModel`

error: internal compiler error: /rustc-dev/de6d33c033441c5880b863f94d7a3ec8cad141bd/compiler/rustc_middle/src/ty/normalize_erasing_regions.rs:171:13: Failed to normalize Alias(Projection, AliasTy { args: [&'{erased} Inner], def_id: DefId(20:978 ~ creusot_std[0ec0]::model::Dee

<img width="1247" height="989" alt="Image" src="https://github.com/user-attachments/assets/3fd4c615-1461-4f35-94ff-2009d6cc2950" />

pModel::DeepModelTy), .. }) in typing_env=TypingEnv { typing_mode: Analysis { defining_opaque_types_and_generators: [] }, param_env: ParamEnv { caller_bounds: [] } }, maybe try to call `try_normalize_erasing_regions` instead

thread 'rustc' panicked at /rustc-dev/de6d33c033441c5880b863f94d7a3ec8cad141bd/compiler/rustc_middle/src/ty/normalize_erasing_regions.rs:171:13:
Box<dyn Any>
```

Lysxia:
Close #2040

### DIFF
--- a/creusot/src/backend/clone_map/elaborator.rs
+++ b/creusot/src/backend/clone_map/elaborator.rs
@@ -121,7 +121,21 @@ impl<'a, 'ctx, 'tcx> Expander<'a, 'ctx, 'tcx> {
 
         let mut pre_sig = EarlyBinder::bind(ctx.sig(def_id).clone())
             .instantiate(ctx.tcx, subst)
-            .normalize(ctx, typing_env);
+            .try_normalize(ctx, typing_env)
+            .unwrap_or_else(|err| {
+                let source = names.source_id();
+                ctx.fatal_error(
+                    ctx.def_span(source),
+                    format!(
+                        "Failed to translate the contract of {} needed by {}.",
+                        ctx.def_path_str_with_args(def_id, subst),
+                        ctx.def_path_str(source),
+                    ),
+                )
+                .with_span_note(err.span, err.msg)
+                .with_help("There is probably a missing trait bound.")
+                .emit()
+            });
 
         if ctx.def_kind(def_id) == DefKind::Closure {
             // Inline the body of closures
--- a/creusot/src/translation/pearlite.rs
+++ b/creusot/src/translation/pearlite.rs
@@ -23,7 +23,7 @@ mod normalize;
 pub mod visit;
 
 pub(crate) use from_thir::from_thir;
-pub(crate) use normalize::normalize;
+pub(crate) use normalize::{NormalizationError, normalize, try_normalize};
 
 #[derive(Copy, Clone, Debug, TyDecodable, TyEncodable, TypeFoldable, TypeVisitable)]
 pub enum BinOp {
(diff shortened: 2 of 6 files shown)


jhjourdan:
Instead of showing a cryptic error message which could hide another, more important, ICE, I would prefer figuring out where does this normalization comes from and make sure with stop compilation if there has been a type error earlier.

Lysxia:
I'll make the error message more precise. The error comes from normalizing the extern spec of `eq` in a context without the necessary `DeepModel` trait bound (for example when using the std `derive(Eq)` instead of ours).

Lysxia:
There are other fishy calls of normalize that I haven't had the time to look at (e.g., `contract_of` in `specification.rs`). I've left some `span_bug` to have an extra span to look at in case they happen.

A better fix may be to not error in this case, but emit a warning and replace the contract with `#[requires(false)]` (or `#[requires(f.precondition())]`), which will let us compile more things by default even if they can't be proved. However that requires more work to get rustc's normalization functions to not error.

Lysxia:
Subsumed by #2060

Issue #2028, opened 2026-04-12
Two issues with #[bitwise_mode]`


jhjourdan:
1- In this mode, it is impossible to access slices, because the resulting coma code is invalid. The problem lies in the use of `Slice64.get` that takes a bounded integer, rather than a bit vector integer. I guess the solution is to have two version of the Slice modules in the prelude, depending on whether w are in bitwise mode.

2- Shift operators use integers conversions for the offset operand. This results in terms of the form `of_int (t'int x)`, which the provers have trouble converting to `x` (sadly...). The solution is to go through 256 bits integers instead.

@mcoulmance you have been working on bit vector, do you think you could work on this?

claudemarche:
Just my 2 cents:

1- I believe that instead of a copy of the Slices module, there should an alternative function Slice64.get_bv taking a bv as parameter, instead of a bounded integer.

2- I don't think conversion to 256-bits is the right solution : converting the second arg to the bv of the size of the first argument instead. This is because in SMT-LIB, and in Why3 stdlib, the shift operators take bv of the same size. 

I'm available to discuss these if you like

jhjourdan:
> Just my 2 cents:
> 
> 1- I believe that instead of a copy of the Slices module, there should an alternative function Slice64.get_bv taking a bv as parameter, instead of a bounded integer.
> 

The problem is that `get` is not the only function of the `Slice` module which is manipulating machine integers. Almost all of the functions of the module are. So we would have to duplicate most of the module anyway. And in addition, a separate module prevents Why3 from importing all the lemmas about bitvectors/machine integers when they are not needed by the current mode.

> 2- I don't think conversion to 256-bits is the right solution : converting the second arg to the bv of the size of the first argument instead. This is because in SMT-LIB, and in Why3 stdlib, the shift operators take bv of the same size.

There are two problems in doing that:
- We don't want to deal with the combinatorics of all the conversions between every sizes of bit vectors. Going through BV256 is a way of using a "middle" type for the conversion.
- We do not only have to perform the shift operation, but we also need to check that there is no overflow in the shift operand. If we convert directly to the right BV type, there might be an overflow during the conversion, so we would have to check for overflow in a polymorphic fashion, before the conversion. Converting it to BV256 makes it possible to do the conversion as a precondition of the shift operator, which is more convenient.

[...]

PR #2045, opened 2026-04-17
Improvements for bitwise mode

mcoulmance:
See #2028 for context


creusot/src/backend/program.rs
```diff
@@ -590,10 +590,14 @@ impl<'tcx> RValue<'tcx> {
                         // so we need to import the prelude for the right operand
                         let qname = match r_ty.kind() {
                             TyKind::Int(ity) => {
-                                lower.names.in_pre(ity_to_prelude(lower.ctx.tcx, *ity), "to_int")
+                                let conv =
+                                    if lower.names.bitwise_mode() { "to_BV256" } else { "to_int" };
```
jhjourdan, line 594: Does this do sign-extension?

(I wonder whether we do the right thing wrt. sign extension in the case of other arithmetic operators, too.)

jhjourdan:
Could you please not update all the proof.json files, to make it clear which proof break?

creusot/src/backend/program.rs
mcoulmance: For signed integer the definition of "to_BV256" uses "stoBig" from Why3, which does sign extension

prelude-generator/int.in.coma
```diff
@@ -133,13 +133,13 @@ module UInt$bits_count$BW
     { [@expl:remainder by zero] b <> zeros  \/ t'int b <> 0 }
     (ret (result: t) { t'int result = mod (t'int a) (t'int b) } { result = urem a b })
     = any
-  let shl (a:t) (b:int)
-    { [@expl:out-of-bounds shifting] b >= 0 /\ b <= size  }
-    (ret (result :t) { result = shl a (of_int b) })
+  let shl (a:t) (b:BV256.t)
+    { [@expl:out-of-bounds shifting] (BV256.t'int b) >= 0 /\ (BV256.t'int b) <= size  }
```
jhjourdan: Please use the comparison operators from `BV256` instead of converting to integers, just like we do for other arithmetic operators.

tests/should_succeed/slices/bitwise.rs
```diff
@@ -0,0 +1,36 @@
+use creusot_std::prelude::*;
+
```
jhjourdan, line 2: Can you confirm that this test used to fail?

jhjourdan (approved):
Overall, LGTM.

One minor remark: checking a function in bitwise mode should be doable without the use of integer arithmetic, since the conversions are badly handled by solvers. So the preconditions or the shift operators should be either stated as a disjunction with both variants (like with other arithmetic operators), or with only the BV version. But the BV variant is really important.

tests/should_succeed/slices/bitwise.rs
mcoulmance: Before the patch we had the following errors:

for "all_ldz":
```
[coma typing] type error in application (expected <t>, got <t1>)
```

for "boolvec_to_bitvec":
```
This term has type creusot.int.UInt64.t, but is expected to have type bv.BV64.t
```

Lysxia (approved):
This looks good to me! Let's merge this today before the release.

PR #2048, opened 2026-04-20
Force why3find to only use solvers that are mentionned in why3.conf

mcoulmance:
This should avoid problems when running the testsuite in an environment where other solvers we don't care about are installed.


Lysxia:
Interestingly, `cargo-creusot` avoids this problem by adding `~/.local/share/creusot/bin` to the `PATH`... I think it may be a bit more disciplined to also get `cargo-creusot` to pass this option to why3find but it seems to work already so I don't know if we must do it.

PR #2052, opened 2026-04-21
Upgrade toolchain to nightly-2026-04-21

Lysxia:
- `instantiate` wraps its result in `Unnormalized<_>` so this is a good opportunity to revisit how we handle normalization
- `DefKind::Const` and `AssocConst` now carry a boolean to keep track of whether they are "suitable for generics", if I understand correctly. I don't think it affects us (hence this PR just ignores that boolean), but while looking into it I learned that people working on const generics have at least thought about having a termination checker to enable more powerful const generics: https://hackmd.io/OsVAAEgOS2ajDwxg44kN-A?view
- the only coma diff shows that `AtomicBool` has become a synonym for `Atomic<Bool>` (etc.) which seems relevant to concurrency folks https://github.com/rust-lang/rust/pull/153015

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,9 +32,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.21"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
+checksum = "824a212faf96e9acacdbd09febd34438f8f711fb84e09a8916013cd7815ca28d"
 dependencies = [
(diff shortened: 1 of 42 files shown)


Lysxia:
I've been looking into removing `Namer::normalize` (as there is a TODO saying to remove it), the following uses remain:
- normalize RPITs in the result of a function
- normalize ADT field types

The first one seems not too hard to avoid, but the second seems more essential.

I got rid of other uses by pushing normalization to the translation phase (MIR -> FMIR and Pearlite THIR -> Term), although it feels rather fragile... The tests pass but I wouldn't be surprised if there are still terms containing unnormalized or unerased types that we just happen to not look at.

Lysxia:
I think I'm about done cleaning up the normalizes for this PR.

PR #2057, opened 2026-04-22
Changes to the `Protocol`

dianegolfouse:
The `Protocol` trait now has a separate `public` function, that computes the public part from the private part.
This allows us to implement `NonAtomicInvariant::open_mut`, that changes the public part along with the private.

Proofs are unaffected.


jhjourdan:
Thanks!

PR #2065, opened 2026-04-25
Define the `GhostShared` type

dianegolfouse:
Useful in combination with `FullBorrow` to create shared borrow with a synthetic lifetime.


jhjourdan:
@dianegolfouse I did a few changes, feel free to comment. But since you are supposed to be in vacations, I merge now.

PR #2072, opened 2026-04-27
tests: Remove dependency on git2

Lysxia:
We only used it for doing `git diff`, which we can do more simply as a `Command`. This removes a good number of transitive dependencies.


jhjourdan:
Nice!

xldenis:
:fire:

PR #2075, opened 2026-04-28
Draft: Add spin lock example.

jhjourdan:
This is an attempt at verifying a spin lock implementation.

I am reasonably satisfied with these examples, which are not so long, and which the SMT solvers solve quite easily. There is one problem in both the SC version and the RelAcq version: we cannot prove the `into_inner` function, because we cannot setup a "contract" that a `FullBorrow` should follow and which `BorrowEnd` may rely on. Without that, it is impossible to prove in `into_inner` that the permission that we get back is indeed the permission for the location protected by the lock.

### DIFF
--- a/examples/spin_lock/relacq.coma
+++ b/examples/spin_lock/relacq.coma
@@ -17,7 +17,7 @@ module M_impl_SpinLock_T__new (* SpinLock<T> *)
   
   type t_Perm_PermCell_T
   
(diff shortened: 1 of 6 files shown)


jhjourdan:
Cc @dianegolfouse we need to discuss this: how shouldwe make `into_inner` provable.

PR #2076, opened 2026-04-28
Fix translation of const to respect visibility

Lysxia:
Close #2009

- [x] add `#[ensures]` for consts
    - [x] translate the ensures to an axiom in callers
- [x] add an opacity check for consts (my plan is that the body of a const is as visible as the const itself unless it has an `#[ensures]`)
- [x] add an attribute `#[constant(eval)]` to force the evaluation of consts. Currently evaluating consts is the default, but it's a bad default because it breaks opacity. We need to keep the ability to evaluate consts because `MIN` and `MAX` for machine integers are defined using bitwise operations (`isize::MIN = !isize::MAX`, `isize::MAX = (usize::MAX >> 1) as isize`, `usize::MAX = !0`), and we don't want to expose these definitions to users.
- [x] Translate consts to verify their type invariants.
- [x] In the translation of terms, error when we encounter a non-visible constructor (due to an evaluated const) instead of outputting bad coma.


Lysxia:
I've rewritten the opacity check to operate on THIR instead of Pearlite terms to be able to apply it to `const` items. In the process I discovered that we need to explicitly call `ctx.sig(_)` and `ctx.term(_)` on every function where that is applicable otherwise the exported metadata is going to be missing some terms or specifications. Previously the opacity check was doing that incidentally. Now there is a function `force_translation` to do that explicitly.

PR #2078, opened 2026-04-29
Prove logical atomicity of a fetch-and-add operation implemented using CAS

jhjourdan:
Interestingly, we do not need to have an abort case in the atomic update like in Iris or Verus. The reason is that we do not need the permission to the atomic variable when then compare_exchange operation fails.

### DIFF
--- a/examples/logically_atomic_faa.rs
+++ b/examples/logically_atomic_faa.rs
@@ -1,0 +1,42 @@
+use creusot_std::{
+    ghost::FnGhost,
+    prelude::*,
+    std::sync::{atomic::Ordering, atomic_sc::AtomicI32, committer::Committer},
+};
+
+#[requires(forall<c: &mut Committer<AtomicI32, i32, Ordering::SeqCst, Ordering::SeqCst>>
+    !c.shot_store() ==> c.ward() == *at ==> c.val_store() == val + c.val_load() ==>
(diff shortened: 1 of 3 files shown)


examples/logically_atomic_faa.rs
```diff
@@ -0,0 +1,42 @@
...
+)]
+#[ensures(exists<c: &mut Committer<AtomicI32, i32, Ordering::SeqCst, Ordering::SeqCst>>
+    !c.shot_store() && c.ward() == *at && c.val_store() == val + c.val_load() &&
+    c.val_load() == result && f.postcondition_once((c,), ())
+)]
+pub fn fetch_add<F>(at: &AtomicI32, val: i32, f: Ghost<F>) -> i32
+where
+    // Interestingly, we do not need to have an "abort" case in the atomic update like in Iris or
+    // Verus. The reason is that we do not need the permission to the atomic variable when then
+    // compare_exchange operation fails.
```
xldenis, line 19: Why does Verus require permission?

jhjourdan: The separation logic rule requires the permission because in the SL setting, it might not be safe to access the atomic variable (it may not be allocated, or a non-atomic access may race, ...). In our setting, we know that it is safe, because Rust's type system is ensuring this to us.

I think the Verus developers just copied the SL logic rule without figuring out that this simplification was possible.

PR #2081, opened 2026-04-30
Specify `NonNull<T>`

dianegolfouse:


creusot-std/src/std/ptr/nonnull.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+impl<T: ?Sized> DeepModel for NonNull<T> {
+    type DeepModelTy = <*mut T as DeepModel>::DeepModelTy;
+    #[trusted]
+    #[logic(opaque)]
+    #[ensures(result.addr == self.addr_logic())]
+    fn deep_model(self) -> Self::DeepModelTy {
+        self.view().deep_model()
+    }
```
jhjourdan:
```suggestion
    #[logic(open, inline)]
    fn deep_model(self) -> Self::DeepModelTy {
        self.view().deep_model()
    }
```

Wouldn't it make sense to make this definition fully public?

dianegolfouse: You're right !

jhjourdan:
Thanks, LGTM.

PR #2082, opened 2026-04-30
Fixes for `FullBorrow` and `EndBorrow`

dianegolfouse:
@jhjourdan note that `FullBorrow` was not `#[opaque]`, I hope this does not break your proofs.


jhjourdan:
That should not change anthing important for clients , because anyway they cannot observe the fields contents.

PR #2086, opened 2026-05-04
Support DoubleEndedIterator slice

Halbaroth:
I need to implement these iterators to verify some parts of IndexMap.


Halbaroth:
Tests failed because I am trying to prove a wrong formula :/ 
I fixed it, as follows
```rust
    #[logic(open)]
    fn produces_back(self, visited: Seq<Self::Item>, o: Self) -> bool {
        pearlite! {
          self@.to_ref_seq() == o@.to_ref_seq().concat(visited.reverse()))
        }
    }
``` 
I managed to prove lemmas in Why3 for the stdlib and now I am trying to fix my test :)

Halbaroth:
Ok, I think that I understood the reason of the failure. If I wrote:
```rust
iter.next_back().map(|a| proof_assert!(a@ == 6));
iter.next().map(|a| proof_assert!(a@ == 1));
``` 
Creusot doesn't know if the previous `next` operation succeed in the second assertion. Now, if I write
```rust
match iter.next_back() {
     Some(a) => {
        proof_assert!(a@ == 6);
        match iter.next() {
          Some(a) =>
            proof_assert!(a@ == 1),
          None => ()
        }},
      None => ()
    }
```
Provers manage to solve it :)

jhjourdan:
FWIW, I am working on that. I will add impls for many other iterators of the same kind, only require automated proofs on the Why3 side, ...

Halbaroth:
Ok! My implementation doesn't scale well. My test 
```rust
pub fn check_iter_1() {
    let numbers = vec![1, 2, 3, 4];
    let mut iter = numbers.iter();
    let a = iter.next_back().unwrap();
    proof_assert!(a@ == 4);
    let b = iter.next_back().unwrap();
    proof_assert!(b@ == 3);
    let c = iter.next_back().unwrap();
    proof_assert!(c@ == 2);
    let d = iter.next_back().unwrap();
    proof_assert!(d@ == 1);
}
```
works with `creusot` strategy for three elements but for four I used `auto 3` strategy. 

Have you a dev branch? I could use it for IndexMap? Otherwise, I will use my naive approach on a local branch to continue.

jhjourdan:
I'm not very surprised by the scalability issue. Iterators are better suited to do loops rather than this kind of proofs which actually involves a lot of quantifier instantiation / sequence reasoning.

creusot-std/src/logic/seq.rs
```diff
@@ -413,6 +417,19 @@ impl<T> Seq<T> {
...
+    pub fn concat_assoc(self, other1: Self, other2: Self) {}
+
+    #[logic]
+    #[ensures(self.concat(Seq::empty()) == self)]
+    #[ensures(Seq::empty().concat(self) == self)]
+    pub fn concat_empty(self) {}
+
+    #[logic]
+    #[ensures(self.concat(other).reverse() == other.reverse().concat(self.reverse()))]
+    pub fn reverse_concat(self, other: Self) {}
```
Halbaroth, line 432: Nice :), I wanted to add these lemmas but I didn't know where to add them.

Halbaroth:
Can I ask few questions about your code? It's not a review but I want to learn how to contribute in Creusot.

jhjourdan:
Sure

creusot-std/src/logic/seq.rs
jhjourdan: That's a bit sad, first because they ar already available in Why3, but importing them by default makes some proofs fail. Second because I would have liked the SMT solvers to automatically use them through the theory of sequences, but I have never been able to make the theory of sequence work properly in Why3...

jhjourdan:
I shortened the `iter_mut` test, and added an assertion in the `iter` test. This should be good to merge.

Anyway, iterator specifications are designed for reasoning about loops, and not for this kind of test. Sequence reasoning is difficult.

creusot-std/src/std/array.rs
```diff
@@ -86,15 +86,45 @@ impl<T, const N: usize> IteratorSpec for IntoIter<T, N> {
...
 
-    #[logic(open, law)]
+    #[logic(law)]
     #[requires(a.produces(ab, b))]
     #[requires(b.produces(bc, c))]
     #[ensures(a.produces(ab.concat(bc), c))]
-    fn produces_trans(a: Self, ab: Seq<Self::Item>, b: Self, bc: Seq<Self::Item>, c: Self) {}
+    fn produces_trans(a: Self, ab: Seq<T>, b: Self, bc: Seq<T>, c: Self) {
+        let _ = ab.reverse_concat(bc);
+        let _ = Seq::<T>::concat_assoc;
```
Halbaroth, line 101: Is this the standard Creusot approach for lemma functions? If a VC isn't proved using `./t why3 ...`, do you recommend writing these lemmas instead of using the Why3 IDE?

creusot-std/src/std/collections/hash_map.rs
```diff
@@ -85,11 +85,11 @@ impl<K: DeepModel, V, A: Allocator> IteratorSpec for IntoIter<K, V, A> {
         pearlite! { resolve(self) && self@.is_empty() }
     }
 
-    #[logic(open, law)]
```
Halbaroth, line 88: Why did you remove `open` here? It was useless because the implementation is empty or it has another effect?

Halbaroth:
I plan to implement `IteratorSpec` for `Drain` too. Have you already implemented it?

creusot-std/src/std/collections/hash_map.rs
jhjourdan: These laws are just lemmas, which return unit. There is no reason to expose the body, because it is only unit.

Nothing very fancy here.

creusot-std/src/std/array.rs
jhjourdan: Yes. The recommended workflow is to only use Why3IDE to debug proofs, but not keep the session files.

jhjourdan:
We cannot fully specify `Drain` in Creusot for now, because its behavior depends on whether or not the destructor of `Drain` is called or not.

PR #2100, opened 2026-05-12
tutorial: add example flake.nix to installation section

doyougnu:
Hi there. I was going through the installation process and thought I could improve the nix section. Any and all comments appreciated.


jhjourdan:
@lafeychine You are the Nix expert.

In particular, we need to figure out what to tell about the free version of alt-ergo.

lafeychine:
Thanks for your contribution. However, sadly, I do not understand what the point is.
Is it a proposal to indicate how one could integrate Creusot in a project using flakes? 

I have some questions:
1. Creusot works with whatever Rust version the project has.
    There is no need to pin a specific toolchain, only `creusot.packages.${system}.default` would be required.

2. I am against hiding the unfree flag, particularly through the use of a third-party tool. 
    If someone wants to hide the flag from the CLI, they should take responsibility for doing so.
    This repository should not provide pointers on how to hide this flag.
    By the way, I realise that the free option is only available for development shells, I will submit a PR shortly.

3. I do not understand why one would need either `openssl` or `pkg-config` to use Creusot.

However, I agree with the `shellHook` part, which ensures that the absolute paths in the configuration file are up to date (except that `why3 config detect` is not the right way to do that). In my opinion, Creusot should handle this check instead of third parties. There are ongoing discussions about how to do this properly.

doyougnu:
Thanks for the reply. Taken in order:

>  Is it a proposal to indicate how one could integrate Creusot in a project using flakes?

I'm running NixOS and use flakes for each project. So yes the proposal is to just provide a simple flake to integrate Creusot into a project. My desire was just to provide a template for anyone who wants to use creusot on nixos to get up and running as quickly as possible. I'm just starting to get deeper into the rust ecosystem so I found myself wanting an example like this.

> Creusot works with whatever Rust version the project has.

Great I'll remove it.

> I am against hiding the unfree flag...

Totally fair. I'll update this section. 

> I do not understand why one would need either openssl or pkg-config to use Creusot.

They don't, this was my mistake.

>  In my opinion, Creusot should handle this check instead of third parties. There are ongoing discussions about how to do this properly.

Should I remove `why3 config detect` then?

guide/src/installation.md
```diff
@@ -49,6 +49,51 @@ All you have to do is activate flakes temporarily by using `--extra-experimental
...
+            echo "========== Creusot shell loaded! =========="
+            why3 config detect
+          '';
+        };
+      }
+    );
+}
+```
+Note, you'll need to allow non-free packages to be able to build the `alt-ergo` package.
+
```
lafeychine: I have made two PRs (#2102 and #2103), which allow one to use Creusot's overlay.
Since I also changed `flake-utils` to use `flake-parts`, I have rewritten the file you have proposed as below.
Please let me know if there are any changes you would like to make.

````suggestion
## Use Creusot within a project (using `nix`)

Here is a sample `flake.nix` that provides a shell with Creusot as a dependency, using `flake-parts`:
```
{
  inputs = {
    # Both `nixpkgs` and `flake-parts` are pinned to the same version as Creusot's
    nixpkgs.follows = "creusot/nixpkgs";
    flake-parts.follows = "creusot/flake-parts";

    creusot.url = "github:creusot-rs/creusot";
  };

  outputs =
    inputs@{
      creusot,
      flake-parts,
      nixpkgs,
      self,
    }:
    flake-parts.lib.mkFlake { inherit inputs; } {
      systems = [
        "aarch64-darwin"
        "x86_64-linux"
      ];

      perSystem =
        {
          pkgs,
          system,
          ...
        }:
        {
          # `pkgs` will also contain the `creusot` set, thanks to Creusot's overlay
          _module.args.pkgs = import nixpkgs {
            inherit system;
            overlays = [ creusot.overlays.default ];
          };

          # `nix fmt`
          formatter = pkgs.nixfmt-tree;

          # `nix develop`
          devShells.default = pkgs.mkShell {
            packages = [
              # `mkCreusotWrapped` produces a wrapped version of Creusot, along with all its dependencies (similar to `clang`).
              # The `isFree` argument allows switching between the free and non-free version of the solver `alt-ergo`.
              (pkgs.creusot.mkCreusotWrapped { isFree = true; })

              # Further packages could be added, notably `cargo` for the `cargo creusot` command
              pkgs.cargo
              pkgs.clippy
              pkgs.rust-analyzer
              pkgs.rustfmt
              # ...
            ];
          };
        };
    };
}
```

````

doyougnu:
I verified the new version locally and had no issues. +1 from me to merge @lafeychine thank you for you labor

PR #2104, opened 2026-05-14
Fix git repository URL in book

Eh2406:
I was having trouble getting back from the book to the repository, Eventually figured out that the book had the wrong link. Thought I would fix it for the next person.


Lysxia:
Thanks!

PR #2108, opened 2026-05-18
split `Ghost` of tuples up to size 8

dianegolfouse:


jhjourdan:
Why not factoring the case of pairs with other tuple sizes? Why not supporting 1-uples and even unit?

dianegolfouse:
The pairs, because I liked having the docs say 'pair' :)
As for the rest, why not, but I never felt the need for them myself.

PR #2109, opened 2026-05-18
Fix sorting of modules

Lysxia:
Compare string representations instead of meaningless and unstable indices


jhjourdan:
I thought stable hashes were stable...

jhjourdan:
Is the comparison function valid in the case several modules have the same name?

Lysxia:
This didn't use stable hashes, but the `usize` underlying a `Symbol` which depends on the order in which strings are interned.

The sorting happens at the end, after module name collisions have been handled (by `naming.rs`), so each name should be unique at that point.

PR #2112, opened 2026-05-20
Implement `SliceIndexSpec` for intervals

Halbaroth:
The `SliceIndexSpec` trait is not implemented for `(Bound(_), Bound(_))` type but `SliceIndex` is implemented for these types in the standard library. 

I'm concerned about usize overflow in the formula:
```rust
(start + 1usize..end).in_bounds(seq)
```
It seems that the Rust runtime panics on usize overflows, so we don't produce wrong formulae at least.
A solution is to implement `SliceIndexSpec` for ranges on `Int` type and write 
```rust
(start@ + 1..end).in_bounds(seq)
```
You might have a better idea?

### DIFF
--- a/creusot-std/src/std/slice.rs
+++ b/creusot-std/src/std/slice.rs
@@ -277,18 +277,20 @@ impl<T> SliceIndexSpec<[T]> for RangeToInclusive<usize> {
     }
 }
 
-#[logic(open)]
-pub fn to_logic_range(interval: (Bound<usize>, Bound<usize>), len: Int) -> Range<Int> {
-    match interval {
-        (Bound::Included(start), Bound::Included(end)) => start.view()..end.view() + 1,
-        (Bound::Included(start), Bound::Excluded(end)) => start.view()..end.view(),
-        (Bound::Included(start), Bound::Unbounded) => start.view()..len,
-        (Bound::Excluded(start), Bound::Excluded(end)) => start.view() + 1..end.view(),
-        (Bound::Excluded(start), Bound::Included(end)) => start.view() + 1..end.view() + 1,
-        (Bound::Excluded(start), Bound::Unbounded) => start.view() + 1..len,
-        (Bound::Unbounded, Bound::Unbounded) => 0..len,
-        (Bound::Unbounded, Bound::Included(end)) => 0..end.view() + 1,
-        (Bound::Unbounded, Bound::Excluded(end)) => 0..end.view(),
+#[logic(open, inline)]
+pub fn to_logic_range((lo, hi): (Bound<usize>, Bound<usize>), len: Int) -> Range<Int> {
+    pearlite! {
+        let lo = match lo {
+            Bound::Included(i) => i@,
+            Bound::Excluded(i) => i@ + 1,
+            Bound::Unbounded => 0
+        };
+        let hi = match hi {
+            Bound::Included(i) => i@ + 1,
+            Bound::Excluded(i) => i@,
+            Bound::Unbounded => len
+        };
+        lo..hi
     }
 }
 
--- a/creusot-std/src/std/slice.rs
+++ b/creusot-std/src/std/slice.rs
@@ -277,63 +277,40 @@ impl<T> SliceIndexSpec<[T]> for RangeToInclusive<usize> {
     }
 }
 
+#[logic(open)]
+pub fn to_logic_range(interval: (Bound<usize>, Bound<usize>), len: Int) -> Range<Int> {
+    match interval {
+        (Bound::Included(start), Bound::Included(end)) => start.view()..end.view() + 1,
+        (Bound::Included(start), Bound::Excluded(end)) => start.view()..end.view(),
+        (Bound::Included(start), Bound::Unbounded) => start.view()..len,
+        (Bound::Excluded(start), Bound::Excluded(end)) => start.view() + 1..end.view(),
+        (Bound::Excluded(start), Bound::Included(end)) => start.view() + 1..end.view() + 1,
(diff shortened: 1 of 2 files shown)


Halbaroth:
Looks like range expressions are expanded to `std::ops` and not `core::ops`. According to the Rust reference, https://doc.rust-lang.org/reference/expressions/range-expr.html, range expressions can be converted to `core::ops::*`. I guess that it's controlled by `#![no_std]` macro?

Halbaroth:
I tested on Compiler Explorer and adding `#![no_std]` changes the expansion of range expressions as expected. It should works as you have 
```rust
#![cfg_attr(not(feature = "std"), no_std)]
```
in `creusot-std/src/lib.rs`.

Are you sure to compile `creusot-std` without `std` feature in your test `./t build-no-std`?

jhjourdan:
> Are you sure to compile `creusot-std` without `std` feature in your test `./t build-no-std`?

We have recently had bugs uncovered by this CI job, so, yes, I'm pretty sure. Why are you concerned?

jhjourdan:
> The `SliceIndexSpec` trait is not implemented for `(Bound(_), Bound(_))` type but `SliceIndex` is implemented for these types in the standard library.
> 
> I'm concerned about usize overflow in the formula:
> 
> ```rust
> (start + 1usize..end).in_bounds(seq)
> ```


You're right to be concerned by this. Pearlite semantics on `usize` is wraparound on overflow, so this definitely not what we want here.

> 
> It seems that the Rust runtime panics on usize overflows, so we don't produce wrong formulae at least. 

It depends on the configuration. By default, it panics in debug mode, and wraps around in release mode to avoid dynamic checks.

> A solution is to implement `SliceIndexSpec` for ranges on `Int` type and write

Unfortunately, `SliceIndexSpec` inherits `SliceIndex`, which is sealed in std. So we cannot do that.

> You might have a better idea?

Well, I'm not sure it's "better", but you could:

1- first convert the pair of bounds into a `Range<Int>`, given the length of the sequence
2- in `in_bounds`, check whether the range is valid for the given sequence (i.e., start <= end <= len)
3- in `has_value`, directly use sequence indexing with int ranges (cf https://github.com/creusot-rs/creusot/pull/1858)
4- in `resolve_elswhere` quantify over all the elements that are both in the sequence but not in the range.

Halbaroth:
I followed your approach. I struggled with the visibility of my auxiliary function `to_logic_range`. I want to use `logic(open)` as writing its specification in pearlite would be annoying and its specification is exactly its code! 

You might prefer writing a trait to attach this function to the type `(Bound<usize>, Bound<usize>)`?

Halbaroth:
> We have recently had bugs uncovered by this CI job, so, yes, I'm pretty sure. Why are you concerned?

The CI failed with this error:
```
   --> creusot/creusot-std/src/std/slice.rs:319:18
    |
319 |                 (start..=end).resolve_elswhere(old, fin)
    |                  ^^^^^^^^^^^ could not find `std` in the list of imported crates
```
I believe that `(start..=end)` is expanded to `std::ops::RangeInclusive` instead of `core::ops::RangeInclusive`.

creusot-std/src/std/slice.rs
```diff
@@ -277,6 +277,44 @@ impl<T> SliceIndexSpec<[T]> for RangeToInclusive<usize> {
     }
 }
 
+#[logic(open)]
```
jhjourdan:
```suggestion
#[logic(open, inline)]
```

That's typically the kind of things you want to unfold eagerly.

creusot-std/src/std/slice.rs
```diff
@@ -277,6 +277,44 @@ impl<T> SliceIndexSpec<[T]> for RangeToInclusive<usize> {
...
+        (Bound::Included(start), Bound::Excluded(end)) => start.view()..end.view(),
+        (Bound::Included(start), Bound::Unbounded) => start.view()..len,
+        (Bound::Excluded(start), Bound::Excluded(end)) => start.view() + 1..end.view(),
+        (Bound::Excluded(start), Bound::Included(end)) => start.view() + 1..end.view() + 1,
+        (Bound::Excluded(start), Bound::Unbounded) => start.view() + 1..len,
+        (Bound::Unbounded, Bound::Unbounded) => 0..len,
+        (Bound::Unbounded, Bound::Included(end)) => 0..end.view() + 1,
+        (Bound::Unbounded, Bound::Excluded(end)) => 0..end.view(),
+    }
+}
```
jhjourdan:
```suggestion
pub fn to_logic_range((lo, hi): (Bound<usize>, Bound<usize>), len: Int) -> Range<Int> {
    pearlite! {
        let lo = match lo {
            Bound::Included(i) => i,
            Bound::Excluded(i) => i + 1,
            Bound::Unbounded => 0
        };
        let hi = match hi {
            Bound::Included(i) => i + 1,
            Bound::Excluded(i) => i,
            Bound::Unbounded => len
        };
        lo..hi
    }
}
```

jhjourdan:
> > We have recently had bugs uncovered by this CI job, so, yes, I'm pretty sure. Why are you concerned?
> 
> The CI failed with this error:
> 
> ```
>    --> creusot/creusot-std/src/std/slice.rs:319:18
>     |
> 319 |                 (start..=end).resolve_elswhere(old, fin)
>     |                  ^^^^^^^^^^^ could not find `std` in the list of imported crates
> ```
> 
> I believe that `(start..=end)` is expanded to `std::ops::RangeInclusive` instead of `core::ops::RangeInclusive`.

Well spotted! I'll submit a patch.

PR #2119, opened 2026-05-26
Add external specifications for ExactSizeIterator implementation of Iter

Halbaroth:
It is sufficient for IndexMap :)


jhjourdan:
I'm working on this, espetially giving more precise specs for this traits and the `size_hint` function.

xldenis:
It would be cool if you could prove termination of a for loop or size invariance of Map using this

jhjourdan:
Size invariance of `Map`: if you mean that `size_hint` and `len` are unchanged by `Map`: sure, this is easy.

For loop termination: this is not the primary objective, but this should work for iterators where `size_hint` has a good implementation.

PR #2123, opened 2026-05-28
Implement `RA` for tuples of size <= 4

dianegolfouse:


creusot-std/src/logic/ra/prod.rs
```diff
@@ -8,11 +8,14 @@ use crate::{
 
 // TODO: general tuples
```
jhjourdan:
```suggestion
```

jhjourdan:
This looks good to me, you can merge.

PR #2126, opened 2026-06-03
Move solvers to separate Why3 conf file

Lysxia:
- [x] doc
- [x] nix


creusot-install/creusot_why3.conf
```diff
@@ -0,0 +1,86 @@
+[prover]
```
lafeychine, line 1: Should this file be moved next to `why3find.json`? No strong opinion for this, just asking.

lafeychine (approved):
LGTM

creusot-install/creusot_why3.conf
Lysxia: Just my preference having fewer files in the root directory.

jhjourdan:
Should we merge?

PR #2129, opened 2026-06-04
fix(nix): Update FOD hash

zimward:
As the title says. The current devshell in the installation fails as the url used was referencing the head instead of the stable commit url and the file was changed upstream.

### DIFF
--- a/nix/deps/cvc5.nix
+++ b/nix/deps/cvc5.nix
@@ -44,8 +44,8 @@ let
     patches = [
       (fetchpatch {
         name = "CoCoALib-0.99800-trace.patch";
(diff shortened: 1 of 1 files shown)


lafeychine:
I agree that we should change URLs to stable ones.
There are other remaining cases that we need to fix.

However, I strongly disagree with the `sha256` hash you provided.
It appears that you computed the `sha256` hash of the unnormalised patch file.

The Nixpkgs documentation says:
> It expects patch files as a source and performs normalization on them before computing the checksum.

The normalisation phase [is available here](https://github.com/NixOS/nixpkgs/blob/317fbe5e1a29c7ec3d11744d78637c424aa9d3d0/pkgs/build-support/fetchpatch/default.nix#L66-L80).

It seems strange to me that Nix would agree with your hash. Have you tried building with it?

lafeychine:
As this issue has already been resolved in #2126, I will close it.

However, we can continue the discussion here

PR #2130, opened 2026-06-04
Fix version parsing of alt-ergo for the free version

zimward:
The free version of alt-ergo that gets pulled with the example flake setup doesn't start with a `v` which made the version detection fail.
depends on #2129 rn, but i could rebase it onto master

### DIFF
--- a/creusot-setup/src/tools.rs
+++ b/creusot-setup/src/tools.rs
@@ -108,8 +108,8 @@ fn detect_why3find_version(why3find: &Path) -> anyhow::Result<String> {
 
 fn detect_altergo_version(altergo: &Path) -> anyhow::Result<String> {
     let version_full = run(Command::new(&altergo).arg("--version"))?;
-    let version = version_full.trim_end().strip_prefix("v").map(String::from);
-    version.ok_or(anyhow!("bad Altergo version: {}", version_full))
+    let version = version_full.trim_end();
+    Ok(version.strip_prefix("v").unwrap_or(version).into())
 }
 
 // helpers: Z3


nix/deps/cvc5.nix
```diff
@@ -44,8 +44,8 @@ let
     patches = [
       (fetchpatch {
         name = "CoCoALib-0.99800-trace.patch";
-        url = "https://raw.githubusercontent.com/cvc5/cvc5/refs/heads/main/cmake/deps-utils/CoCoALib-0.99800-trace.patch";
-        sha256 = "sha256-57Ptq3ESLGJjAT9xC8Uk0zrfMoRl+KsU6cg6QxIZH6k=";
+        url = "https://raw.githubusercontent.com/cvc5/cvc5/e0514b3e454aa184654a66b7a979244babbc4c5e/cmake/deps-utils/CoCoALib-0.99800-trace.patch";
+        sha256 = "sha256-hY6/XkqLbqSunH6yUz9PmvrcPBWsd8FJYBUKhuHCYVs=";
```
Lysxia: I am pinning to the same commit here but I got a different hash, what gives? https://github.com/creusot-rs/creusot/pull/2126/changes#diff-01196c418fd9a39a73791721f20dfb266f5d6c15604d3d5c36af28fcc00d62a4R47

creusot-setup/src/tools.rs
```diff
@@ -108,8 +108,12 @@ fn detect_why3find_version(why3find: &Path) -> anyhow::Result<String> {
...
 fn detect_altergo_version(altergo: &Path) -> anyhow::Result<String> {
     let version_full = run(Command::new(&altergo).arg("--version"))?;
-    let version = version_full.trim_end().strip_prefix("v").map(String::from);
-    version.ok_or(anyhow!("bad Altergo version: {}", version_full))
+    let version = version_full.trim_end();
+    // some versions start with dont start with "v"
+    let version = version.strip_prefix("v").map(String::from).unwrap_or_else(||{
+        version.to_string()
+    });
+    Ok(version)
```
Lysxia:
```suggestion
    let version = version_full.trim_end();
    Ok(version.strip_prefix("v").unwrap_or(version).into())
```

- avoid repeating the str-to-String conversion
- use `into` (raw string conversion) instead of `to_string` (formatted string conversion)

Also do run `cargo fmt` to pass CI. (Or `./fmt` if you ever update test files.)

nix/deps/cvc5.nix
zimward: even when you run `nix-hash --to-sri --type sha256 $(nix-prefetch-url https://raw.githubusercontent.com/cvc5/cvc5/e0514b3e454aa184654a66b7a979244babbc4c5e/cmake/deps-utils/CoCoALib-0.99800-trace.patch)` ? otherwise i am not entirely sure what would be going on, as i get the same hash even on a different machine.

Lysxia: nix accepts both hashes! When I run that ^ command I do get your hash `hY6/XkqLbqSunH6yUz9PmvrcPBWsd8FJYBUKhuHCYVs=`. When I leave the hash empty the hash mismatch error message reports my hash `3IrkUx6ygIiDRJAM9OSYxH3Q/Mo+ZjXGPOauMh4uZyw=`.

Lysxia: The plot thickens: the nix on my machine accepts both, but not the one in CI...

zimward: that is to be expected. nix addresses via the hash, so if you have that hash already in your store it'll use that. basically you are performing a FOD poisoning on yourself

Lysxia: I'll let you and our local Nix expert @lafeychine figure out what's going on and continue the discussion in #2129. In the meantime I've merged #2126 which uses the other hash and passes CI.

If you keep just the other commit in this PR I'll merge it.

lafeychine: I answered in #2129

Lysxia:
@zimward Would you like to sign your commit or are you fine with an "Unverifed" commit in the wild?

zimward:
> @zimward Would you like to sign your commit or are you fine with an "Unverifed" commit in the wild?

i am fine with that. commit signatures are only useful to prove that a commit has been made by someone and i don't think this project has a reason to prove that it has been made by me.

Lysxia:
Thanks for the fix!

PR #2131, opened 2026-06-05
Check traits that are recursive via bounds of their associated types

Lysxia:


creusot/src/validate/recursive_types.rs
```diff
@@ -406,13 +406,29 @@ fn add_trait(ctx: &TranslationCtx, def_id: DefId, graph: &mut TypeGraph) {
...
         }
-        for &(clause, span) in ctx.predicates_of(item.def_id.expect_local()).predicates {
-            if let ty::ClauseKind::Trait(predicate) = clause.kind().skip_binder() {
-                graph.add_edge(
-                    node,
-                    TypeNode::Trait(predicate.trait_ref.def_id),
-                    TypeEdge::FnBound(item.def_id, span),
-                );
+        match item.kind {
+            AssocKind::Const { .. } => continue,
```
jhjourdan, line 410: Constans are always fine?

Lysxia:
How do we justify that this is legal? https://github.com/creusot-rs/creusot/blob/2121072bbbe7df3abef7e3d47d5a6522f6a04958/creusot-std/src/std/sync/atomic.rs#L19-L24

```rust
    pub trait Ordering {
        const ORDERING: atomic::Ordering;

        type Load: Ordering;
        type Store: Ordering;
    }
```

creusot/src/validate/recursive_types.rs
Lysxia: AFAICT constants can't be recursive. And even if they were problematic, that would have to be checked in `terminates.rs` (= another patch).

jhjourdan:
> How do we justify that this is legal? 

Good question. Cc @lafeychine .

jhjourdan:
Is there a way to exploit recursivity through associated types that does not involve defautl impl?

tests/should_fail/recursive_types/assoc_type_bounds.rs
jhjourdan, line 1: Another remark: it seems to me that the issue in this example is not the recursivity of the trait (which, after all, can be considered simply as a strictly positive inductive definition), but rather the recursivity of the *impl*, which uses itself to satisfy the constraint `A:Tr`.

jhjourdan: That is not solving the issue with `Ordering`, though, because the corresponding impls are recursive.

jhjourdan: Oh, and wait a second: it seems to me that if we have a trait which is recursive that way, then any impl will need to be recursive, so in the end the check on the trait is the right one.

jhjourdan:
> How do we justify that this is legal?
> 
> https://github.com/creusot-rs/creusot/blob/2121072bbbe7df3abef7e3d47d5a6522f6a04958/creusot-std/src/std/sync/atomic.rs#L19-L24
> 
> ```rust
>     pub trait Ordering {
>         const ORDERING: atomic::Ordering;
> 
>         type Load: Ordering;
>         type Store: Ordering;
>     }
> ```

This is fixed by #2132

Issue #2136, opened 2026-06-09
Cannot use the IDE : "version can't be an empty string"


dianegolfouse:
Trying to use the ide on a fresh project does not work 😢

Steps to reproduce:
- Clone creusot on master (commit https://github.com/creusot-rs/creusot/commit/65b9db6183a1afa56e7d79cc351f429421bf1c96)
- Run `./INSTALL`
- Create a new project with `cargo creusot init`
- Change the spec to be false (e.g. set the ensures to `result@ == a@`)
- Run `cargo creusot -i`

Why3find runs without issues; when creusot tries to launch the IDE though, I get an error
```
anomaly: Invalid_argument("mk_filter_prover: version can't be an empty string")
```

I tried reinstalling creusot from scratch as much as I can, removing `~/.local/creusot`, `~/.cache/creusot` and the local `target` folder in creusot's directory; it still fails.

dianegolfouse:
Weirdly enough, it works with `cargo creusot --ide-always add_one`

Lysxia:
T_T a workaround is to add version fields in `~/.local/share/creusot/creusot_why3.conf`. (I didn't do it because I wanted to avoid the string processing needed to insert the version numbers.) See below for the config to copy.

The problem comes from `why3` assigning the empty version string to provers without one, and later complaining when the version string is empty.

```
# add to the top of the file:

[prover]
name = "Alt-Ergo"
version = "2.6.2"
command = "alt-ergo --timelimit %.t %f"
driver = "alt_ergo_26"

[prover]
name = "Z3"
version = "4.15.3"
command = "z3 -smt2 -T:%t sat.random_seed=42 nlsat.randomize=false smt.random_seed=42 -st %f"
driver = "z3_4_12"

[prover]
name = "CVC5"
version = "1.3.1"
command = "cvc5 --stats-internal --tlimit=%T %f"
driver = "cvc5"

[prover]
name = "CVC4"
version = "1.8"
command = "cvc4 --stats --tlimit=%T --lang=smt2 %f"
driver = "cvc4_17"
```

suhr:
Is there a way to apply this workaround if I use a Nix environment?

Lysxia:
Before running Nix, you can write the above into `creusot-install/creusot_why3.conf`.

PR #2141, opened 2026-06-10
Update Why3

Lysxia:
- Allow empty prover version strings

Better fix for #2136 (overrides #2142).


Lysxia:
@lafeychine the nix build needs dune as a new dependency 🐪

lafeychine:
Since `why3` uses `dune` to call `./configure` at the end (`opam/why3.opam`), I would prefer to stick to the old way of building why3, only including `dune` as a new dependency

PR #2148, opened 2026-06-12
Add Chain iterator.

jhjourdan:
Cc @Halbaroth.


jhjourdan:
@Lysxia Looking at the results of the CI, it seems like the installation test is using a version of creusot-std from crates.io. Is this the expected behavior? If so, how should I add a new intrinsic?

Lysxia:
Yeah as soon as we want to change creusot-std we should bump it to a `*-dev` version: #2149

jhjourdan:
Could you please prepare a file RELEASE.md that lists all we have to do for a release, including bumping the version in master after the release?

jhjourdan:
My bad. It's already in CONTRIBUTING...

PR #2149, opened 2026-06-13
Bump version to 0.13.0-dev

Lysxia:


jhjourdan:
There are too many files to change when bumping the version... Isn't there support for variables in Cargo that we could change globally ?

Issue #2074, opened 2026-04-28
Better syntax for specifying functions that call closures


jhjourdan:
Today, the specification of `std::sync::atomic_sc::AtomicBool::load` looks like this:

```rust
#[requires(forall<c: &Committer<Self, bool, Ordering::SeqCst, Ordering::None>>
    !c.shot_store() ==> c.ward() == *self ==> f.precondition((c,))
)]
#[ensures(exists<c: &Committer<Self, bool, Ordering::SeqCst, Ordering::None>>
    !c.shot_store() && c.ward() == *self && c.val_load() == result && f.postcondition_once((c,), ())
)]
pub fn load<F>(&self, f: Ghost<F>) -> bool
```

Which is a rather verbose way to say that this function will call the function taken as a parameter with a parameter `c: &Committer<Self, bool, Ordering::SeqCst, Ordering::None>>` verifying `!c.shot_store() && c.ward() == *self` and that it will return `c.val_load()` as result.

This patterns occurs in many places in `creusot-std`, and, despite its verobosity, `std::sync::atomic_sc::AtomicBool::load` is perhaps the simplest of these instances. For example, the spec of `compare_exchange` in the case of weak memory looks like this:

```rust
            #[requires(forall<c: &mut Committer<Self, $type, _, _>>
                !c.shot_store() ==> c.ward() == *self ==>
                c.val_load().deep_model() == current.deep_model() ==>
                c.val_store() == new ==>
                f.precondition((Ok(c),)) && (f.postcondition_once((Ok(c),), ()) ==> (^c).shot_store())
            )]
            #[requires(forall<c: &Committer<Self, $type, _, _>>
                !c.shot_store() ==> c.ward() == *self ==>
                // NOTE: This following line is not present for `weak`
                c.val_load().deep_model() != current.deep_model() ==>
                f.precondition((Err(c),))
            )]
            #[ensures(
                match result {
                    Ok(result) => {
                        exists<c: &mut Committer<Self, $type, _, _>>
                            !c.shot_store() && c.ward() == *self &&
                            c.val_load().deep_model() == current.deep_model() &&
                            c.val_store() == new &&
                            result == c.val_load() &&
                            f.postcondition_once((Ok(c),), ())
                    },
                    Err(result) => {
                       exists<c: &Committer<Self, $type, _, _>>
                            !c.shot_store() && c.ward() == *self &&
                            // NOTE: This following line is not present for `weak`
[...]

PR #2151, opened 2026-06-15
Implement `such_that_decide`

dianegolfouse:
This is a function that I often need, maybe more often than `such_that`.


jhjourdan:
Should we replace `such_that` by that?

dianegolfouse:
Maybe...

Looking at the uses of `such_that` in `creusot-std` though, I find a lot of `such_that(|_| true)`. Do we want a specialized `any` function for this? Maybe that could fit into https://github.com/creusot-rs/creusot/issues/2074?

jhjourdan:
Yeah, the idea of recycling `such_that` for both uses was retrospectively not so great. When encoding `any`, we have this `|_| true` closure floating around for nothing, which is confusing and complicates the goal.

> Maybe that could fit into #2074?


I don't see how this is related. #2047 is about program/ghost code. This PR is about Pearlite.

dianegolfouse:
Subsumed by #2152

PR #2152, opened 2026-06-15
Add logic::any and logic::try_such_that

jhjourdan:
Cc @dianegolfouse 

I'm following your idea in #2151, but I used the somewhat more idiomatic name `try_such_that` (open to discussion), added `any`, and used both functions here and there in the examples.


jhjourdan:
@dianegolfouse Since this is originally your idea, I let you comment and merge if satisfied.

jhjourdan:
@dianegolfouse should I merge?

dianegolfouse:
Is `any` really always a substitute for `unreachable`? I feel like, at least from a user point of vue, they serve different purposes.
The rest I have no issue with.

PR #2153, opened 2026-06-21
Handle compiler-builtin trait impls instead of ICE-ing (lint + synthesized tuple-Clone law)

### LINKED ISSUE #2094: Creusot crashes when using builtin impls of `Clone`
Creusot crashes on:

```rust
use creusot_std::prelude::*;

struct S(i32);

impl Clone for S {
    #[ensures(result == *self)]
    fn clone(&self) -> Self {
        S(self.0)
    }
}

#[ensures(result == *x)]
pub fn f(x: &(S,)) -> (S,) {
    x.clone()
}
```

Because it does not know that `Clone` has a builtin impl for tuples.

The same is (probably) the case for closures, and probably for other elements of the type grammar.

saroupille:
> _Generated with the help of an AI agent (Claude), under human review._

## What

Stop Creusot from ICE-ing on compiler-synthesized builtin trait impls — in practice tuple `Clone` and closure `Clone` — make the resulting opacity visible via a lint, and recover full precision for tuple `Clone`.

## Why

`select_method`'s `ImplSource::Builtin` arm ended in `unimplemented!`, so any program touching e.g. `(a, b).clone()` crashed the compiler. The minimal sound fix (resolve opaquely) silently dropped all knowledge of the cloned value — a precision cliff with no signpost. For tuple `Clone` the value is statically known, so we can do better than opaque.

## How

Three layers:
1. **Un-ICE** — the builtin arm returns a new `TraitResolved::UnknownBuiltin` instead of panicking. It behaves exactly like the existing `UnknownFound` (opaque, sound) everywhere; it exists as a separate variant *only* to carry the lint signal, with an invariant (documented at the enum) that no `match` may discriminate it elsewhere.
2. **Lint** — `opaque_builtin_impl` (warn-by-default) fires when a call stays opaque because of an unmodeled builtin, so the precision loss is never silent. It is suppressed only when the synthesized law is *complete* (every leaf modeled); a tuple with an unmodeled leaf still warns. Generic `T::clone` and `dyn` reach other resolution paths and are not flagged.
3. **Synthesize tuple `Clone`** — in the backend, attach the element-wise structural postcondition to the tuple-clone val, **recursing through nested tuples to the leaves**, via the `PostconditionOnce` intrinsic — the same law `#[derive(Clone)]` produces. A single predicate, `synthesizes_builtin_law`, gates both the synthesis and the lint suppression, so they cannot desync.

Tuple `Clone` is a compiler builtin with no surface impl to attach a contract to (unlike arrays, which have a real `core` impl and already carry an `extern_spec!` in `creusot-std`), which is why this lives in the backend.

## Properties

- **Sound**: the synthesized `ensures` is a *consequence* of the per-field clone laws (tuple clone is field-wise by definition), not a new axiom on the opaque val. It never asserts anything false.
- **Fully constrained for tuples of modeled types** at any nesting depth (this gives a constraining postcondition; discharge still depends on the leaf contracts and the solver).
- **No silent residual**: if a leaf is unmodeled (a closure, or a contractless-`Clone` user type) the synthesized law is incomplete for that leaf, so the lint *fires* (naming the whole tuple) instead of being suppressed — the loss is always flagged. The modeled leaves are still fully constrained.
- Closure `Clone` stays opaque-but-flagged (it warns). Tuple `PartialEq`/`Ord`/`Hash` and array `Clone` are unaffected — they resolve via real `core`/library impls and never reach this path (tuple `==`/`<` already carry a `deep_model` law).

## Examples (now provable; each ICE-d before)

Verbatim from `tests/should_succeed/bug/tuple_clone_law.rs` (they compile and discharge):

```rust
#[ensures(result.0 == x.0 && result.1 == x.1)]
pub fn clone_flat(x: (u32, bool)) -> (u32, bool) { x.clone() }

// nested: the law recurses to the leaf (result.0.0), not just result.0
#[ensures(result.0.0 == x.0.0 && result.0.1 == x.0.1 && result.1 == x.1)]
pub fn clone_nested(x: ((u32, bool), u8)) -> ((u32, bool), u8) { x.clone() }

// generic field: composes the field's own Clone law
#[ensures(T::clone.postcondition((&x.0,), result.0))]
#[ensures(U::clone.postcondition((&x.1,), result.1))]
pub fn clone_generic<T: Clone, U: Clone>(x: (T, U)) -> (T, U) { x.clone() }

// reference leaf: <&u32 as Clone> is the shared-ref copy law
#[ensures(*result.1 == *x.1)]
pub fn clone_ref_leaf(x: (bool, &u32)) -> (bool, &u32) { x.clone() }
```

## Counter-examples

Verbatim from the committed tests (they compile):

```rust
// tests/should_fail/lints/opaque_builtin_impl.rs — the lint fires:
// (1) a bare closure clone, (2) a tuple with a closure leaf (incomplete law).
pub fn clone_closure(x: u32) -> impl Fn(u32) -> u32 {
    let f = move |y: u32| x + y;
    f.clone()
}
pub fn clone_tuple_with_closure(b: bool, x: u32) -> (bool, impl Fn(u32) -> u32) {
    let t = (b, move |y: u32| x + y);
    t.clone()
}

// tests/should_fail/bug/tuple_clone_not_too_strong.rs — soundness boundary:
// a false cross-field property is correctly NOT provable.
#[ensures(result.0 == result.1)] // does not hold for e.g. (3, 7)
pub fn clone_not_cross_field(x: (u32, u32)) -> (u32, u32) { x.clone() }
```

## Tests

Every EXAMPLE and COUNTER-EXAMPLE above is a committed test (the diff is additive — no existing golden `.coma` changes):

- `tests/should_succeed/bug/tuple_clone_law.rs` (EXAMPLES) — flat mixed-primitive, nested-to-leaf, generic-field, reference-leaf, plus arity-1 and unit-leaf edge cases; all discharge (committed `proof.json`, replayed by the why3 CI job).
- `tests/should_fail/lints/opaque_builtin_impl.rs` (Counter-examples 1 & 2) — a bare closure `Clone` and a tuple with a closure leaf both trip the lint (`#[deny]`, so no `.coma` is emitted and the why3 job skips it, like the `contractless_external_function` lint test).
- `tests/should_fail/bug/tuple_clone_not_too_strong.rs` (COUNTER-EXAMPLE 3) — the false cross-field property is correctly not provable (session records real `unknown` results from Alt-Ergo/CVC4/CVC5).
- Full UI suite green; `cargo fmt --check` clean; `creusot-std` re-translates without panic.

## Notes

1. **This PR can be split.** The first commit (un-ICE → `UnknownBuiltin`) is a tiny,
   self-contained safety fix and could land on its own; the lint, the tuple-`Clone` law
   synthesis, and the lint-completeness refinement would follow as separate PRs.

2. **Reviewable commit by commit.** The four commits are atomic — un-ICE, lint,
   synthesis, lint-completeness — and stand on their own in that order.

3. **Diff composition.** Of ~970 added lines: ~660 are generated golden test
   artifacts (`.coma`, `proof.json`, `why3session.xml`), produced by
   `./t ui --update` and checked by the why3 replay job; ~196 are compiler code
   (`creusot/src/`, a fair share doc-comments); ~76 are test sources. The behavioural
   change is one resolution arm (`unimplemented!` → `UnknownBuiltin`) plus the
   `structural_clone_post` helper and the lint plumbing.

4. **Lint completeness.** Lint suppression is gated on the synthesized law being
   *complete* (every leaf modeled), so a tuple with a closure / contractless-`Clone`
   leaf still warns rather than silently dropping precision on that leaf.


creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -215,11 +268,13 @@ impl<'a, 'ctx, 'tcx> Expander<'a, 'ctx, 'tcx> {
             TraitResolved::NotATraitItem
             | TraitResolved::Instance { .. } // The default impl is known to be the final instance
             | TraitResolved::UnknownFound // Unresolved trait method
+            | TraitResolved::UnknownBuiltin // Unmodeled builtin impl
```
jhjourdan: This should be unreachable, because builtin impls cannot be logic functions. Please remove.

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -215,11 +268,13 @@ impl<'a, 'ctx, 'tcx> Expander<'a, 'ctx, 'tcx> {
...
         );
         // The other case are impossible, because that would mean we are not guaranteed to have an instance
 
-        let opaque = matches!(trait_resol, TraitResolved::UnknownFound)
-            || !ctx.is_transparent_from(def_id, self.namer.source_id());
+        let opaque =
+            matches!(trait_resol, TraitResolved::UnknownFound | TraitResolved::UnknownBuiltin)
+                || !ctx.is_transparent_from(def_id, self.namer.source_id());
```
jhjourdan: Thus, please revert.

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -350,10 +405,12 @@ impl<'a, 'ctx, 'tcx> Expander<'a, 'ctx, 'tcx> {
             TraitResolved::NotATraitItem
             | TraitResolved::Instance { .. } // The impl is known to be the final instance
             | TraitResolved::UnknownFound // Unresolved trait const
+            | TraitResolved::UnknownBuiltin // Unmodeled builtin impl
```
jhjourdan: AFAIK, there is no builtin constant. This is unreachable.

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -350,10 +405,12 @@ impl<'a, 'ctx, 'tcx> Expander<'a, 'ctx, 'tcx> {
...
         );
-        let opaque = matches!(trait_resol, TraitResolved::UnknownFound)
-            || ctx.def_kind(def_id) == DefKind::ConstParam
-            || !ctx.is_transparent_from(def_id, self.namer.source_id());
+        let opaque =
+            matches!(trait_resol, TraitResolved::UnknownFound | TraitResolved::UnknownBuiltin)
+                || ctx.def_kind(def_id) == DefKind::ConstParam
+                || !ctx.is_transparent_from(def_id, self.namer.source_id());
```
jhjourdan: Thus, please revert (and this is anyway wrong, because if there were builtin constant, then we would likely give it a body, thus it would not be opaque.

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -720,7 +777,7 @@ fn postcondition_once_term<'tcx>(
             match TraitResolved::resolve_item(ctx.tcx, typing_env, did, subst) {
                 TraitResolved::NotATraitItem => (),
                 TraitResolved::Instance { def, .. } => (did, subst) = def,
-                TraitResolved::UnknownFound => return None,
+                TraitResolved::UnknownFound | TraitResolved::UnknownBuiltin => return None,
```
jhjourdan: Wrong. If we have a `clone` builtin impl, then we do have a spec.

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -813,7 +870,7 @@ fn postcondition_mut_term<'tcx>(
...
```
jhjourdan: Same here.

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -879,7 +936,7 @@ fn postcondition_term<'tcx>(
...
```
jhjourdan: Same.

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -950,7 +1007,7 @@ fn precondition_term<'tcx>(
...
```
jhjourdan: Same.

creusot/src/backend/resolve.rs
```diff
@@ -226,7 +226,7 @@ pub(crate) fn elaborate_resolve_def<'tcx>(
                 ))
             }
         }
-        TraitResolved::UnknownFound => {
+        TraitResolved::UnknownFound | TraitResolved::UnknownBuiltin => {
```
jhjourdan: The `TraitResolved::UnknownBuiltin` case will never happen, because `Resolve` has no builtin impl. Please move this under an `unreachable` case.

creusot/src/backend/ty_inv.rs
```diff
@@ -124,7 +124,7 @@ pub(crate) fn elaborate_tyinv_def<'tcx>(
         TraitResolved::Instance { def, .. } => {
             rhs = rhs.conj(Term::call(ctx.tcx, names.typing_env(), def.0, def.1, [subject.clone()]))
         }
-        TraitResolved::UnknownFound => {
+        TraitResolved::UnknownFound | TraitResolved::UnknownBuiltin => {
```
jhjourdan: The `TraitResolved::UnknownBuiltin` case will never happen, because `Invariant` has no builtin impl. Please move this under an `unreachable` case.

jhjourdan (changes requested):

jhjourdan:
This is essentially a fix for #2094, thanks.

I will need to review this in detail, before merging, because this touches several subtle parts of Creusot. Before my detailed review, here are a few comments:

- the Why3 test does not work, probably because a proof.json file is missing
- there needs to be some work in the termination check, because these builtin impls can be used to write recursive functions (see https://github.com/creusot-rs/creusot/pull/2006#discussion_r3203236432).
- I don't see why supporting clone of closures is difficult (and btw, you examples under "Counter-examples (three distinct boundaries)" are invalid, they don't compile). A closure is not very different from a tuple actually
- if possible, I'd like all the cases of builtin impls be factorized in `resolve.v`, and then handle differently the relevant cases in the callers. These cases are:
    + Clone builtin impl (this is what you add in this PR). This should go through a specific constructor
    + Fn/FnMut/FnOnce builtin impls for closures, FnDef types and Fn types => Currently this goes through the constructor `Instance`, but really this should use a specific constructor because this special case is debious
    + builtin impls of `dyn` => same here, the current behavior should be kept, but through a different path
- generally, I think the idea of treating builtin impls as "unkonwn" is wrong, and this should be changed: actually, the impl is well-known. It is simply builtin.

saroupille:
Thanks — and you're right that this needs a rethink rather than patches.

I've folded in the surface points: the failing why3 test (the lint test was in `should_succeed/` with a `.coma`, so the proof job tried to discharge its unprovable closure-clone VCs — it's now in `should_fail/lints/opaque_builtin_impl.rs` with `#![deny]`, like `contractless_external_function`), the inline-comment fixes, and the non-compiling snippets in the description above (my apologies for those).

But your main points are a redesign, not edits:
- model a builtin impl as a *known* `Builtin` constructor, not "unknown";
- factor all builtin cases (`Clone`, `Fn*` for closures/`FnDef`/fn-types, `dyn`) in `resolution.rs` behind dedicated constructors, handled per-caller;
- the termination-check interaction (builtin impls enable recursive functions);
- synthesize closure `Clone` too (a closure is close to a tuple).

Rather than push another version blindly, I'd prefer to iterate on the design first and get it validated. If you're OK with it, I'll open an issue sketching the proposed `TraitResolved` reshape (the `Builtin` constructor + caller handling, closure clone, termination) for you to confirm or correct, and revise the PR only once we agree on the design. Does that work?

jhjourdan:
Many thanks for reacting so quickly!

Your plan looks good.

Lysxia:
As not an AI user myself, I'm curious how much of it is written by AI? Did you basically just tell it to "fix the ICE" or did you have to provide more targeted instructions? How much work in terms of time and money does this represent?

saroupille:
> As not an AI user myself, I'm curious how much of it is written by AI? Did you basically just tell it to "fix the ICE" or did you have to provide more targeted instructions? How much work in terms of time and money does this represent?

Apart from this message written by me, all the rest was done using Claude Code (Opus 4.8 model), an AI agent: from the initial patch, up to the PR discussions and comments, through my guidance.

The problem occurred while trying to prove some specification for a smart-contract language (Michelson) that we are rewriting in Rust. While translating the AST, this error occurred and this led to the first commit of this PR that unblocked me.

On Sunday, I wanted to give it a try at upstreaming it, but thought we could do more. I went into a loop of: 1. Making suggestions to help design the feature with the agent 2. Implementing the feature (only the AI agent) 3. Reviewing the code (AI agent + human).

After a few iterations, I considered it was ready to open a PR and let the agent do it and take into account the feedback up to a point where it was obvious the approach wasn't good enough.

More than fixing this bug, this is also the first time I have the opportunity since the AI agent era to contribute to open source code and I want to better understand how I can turn this workflow I described above into a more efficient process that does not generate too much friction or AI fatigue. At the company I am working for, most of our code is now produced by AI agents. I am using AI agents quite a lot, but it is always easier when you start from scratch or on a project you already know, with a team of people you are familiar with, than on an unknown project I am not familiar with at all and a bit guided by the agent, even though I have a background in formal verification.

Right now, I do not consider my contribution a success: 1. Many obvious mistakes were pushed during the first iteration (such as counter-examples not even compiling 😮‍💨 ) 2. Missing the scope of the termination checker as part of the contribution and 3. The design document that you reviewed is too long and not straight to the point.

Those mistakes are symptomatic of the lack of harness on my side [a note on (harness engineering](https://openai.com/index/harness-engineering/). I am looking forward to making better and more efficient contributions.

jhjourdan:
Alright, after reading #2159 , I dediced to patch the PR myself, because the more I look at it, the more I figure out that there are a few subtleties I would like to be done in certain ways, and I found a few bugs along the way too.

jhjourdan:
Superseeded by #2171

PR #2158, opened 2026-06-23
Add specs for bitwise operations on references

### LINKED ISSUE #2157: Missing support for bitwise binary operations (`&`, `|`, `^`, `>>`...) with referenced integer values
Hello,

Creusot currently fails to verify this:

```rust
use creusot_std::prelude::*;

#[bitwise_proof]
#[ensures(result == *a & 0xf_u8)]
#[ensures(result@ <= 0xf)]
pub fn bitwise_and_ref(a: &u8) -> u8 {
    a & 0xf
}
```

It reports:

```text
warning: calling external function `bitand` with no contract will yield an impossible precondition
 --> src/lib.rs:7:5
  |
7 |     a & 0xf
  |     ^^^^^^^ function called here
  |
  = note: `#[warn(creusot::contractless_external_function)]` on by default
```

Using the value directly (with either `a: u8` in the function parameter or `*a & 0xf` in the body) makes Creusot successfully verify the code.

Is this a known limitation of Creusot?

For the context, I was learning Creusot and tried to annotate [crate `hex`](https://crates.io/crates/hex). This crate implements an iterator to format bytes to a hexadecimal string, using a closure that takes a parameter `byte` of type `&u8` and directly performs `>> 4` and `& 0x0F` on it ([source](https://github.com/KokaKiwi/rust-hex/blob/e25a8701f7f6b5ec4acedffcb4a45cfb69cc34eb/src/lib.rs#L99-L109)):

```rust
    fn next(&mut self) -> Option<Self::Item> {
        match self.next.take() {
            Some(current) => Some(current),
            None => self.inner.next().map(|byte| {
                let current = self.table[(byte >> 4) as usize] as char;
                self.next = Some(self.table[(byte & 0x0F) as usize] as char);
                current
            }),
        }
    }
```

... Creusot reported that external functions `bitand` and `shr` were missing contracts.

Lysxia:
Close #2157

### DIFF
--- a/creusot-std/src/std/num.rs
+++ b/creusot-std/src/std/num.rs
@@ -5,7 +5,10 @@ use crate::{
 };
 // Resolve links like [`i8::add`] in the generated documentation
 #[cfg(doc)]
-use core::ops::{Add, AddAssign, Div, DivAssign, Mul, MulAssign, Rem, RemAssign, Sub, SubAssign};
+use core::ops::{
+    Add, AddAssign, BitAnd, BitAndAssign, BitOr, BitOrAssign, BitXor, BitXorAssign, Div, DivAssign,
+    Mul, MulAssign, Rem, RemAssign, Shl, ShlAssign, Shr, ShrAssign, Sub, SubAssign,
+};
 
 macro_rules! mach_int {
     ($t:ty, $ty_nm:expr, $zero:expr, $to_int:literal) => {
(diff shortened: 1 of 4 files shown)


Lysxia:
I need to fix the doc. I am also working on a simplification which broke some other stuff.

niooss-ledger:
Thanks for working on this! By the way, for the SHR test, is there a way to write `>>` with Pearlite's `Int`? Using a division works fine and I am wondering whether I missed a way to express a shift right. For example:

```rust
#[bitwise_proof]
#[ensures(result@ == a@ / 16)]
#[ensures(result@ <= 0xf)]
pub fn bitwise_shr(a: &u8) -> u8 {
    a >> 4
}
```

jhjourdan:
Unfortunately, the SMT solvers don't like mixing mathematical integers with bitvectors. If we implemented a shift operator with `Int`, then the SMT solvers would likely not be able to prove *anything* about this operator, so we prefer not implementing them.

That being said, this operator exists in Pearlite for machine integers, so you should be able to write:

```rust
#[bitwise_proof]
#[ensures(result == a >> 4u8)]
#[ensures(result <= 0xfu8)]
pub fn bitwise_shr(a: &u8) -> u8 {
    a >> 4
}
```

niooss-ledger:
Ok. Thanks for your quick reply. It makes sense and I can use it for my attempt at writing some contracts for [`hex`](https://crates.io/crates/hex) :slightly_smiling_face:

PR #2173, opened 2026-06-25
Fix `Seq::insert`

dianegolfouse:
Previous spec was hard on provers.
I also specified the logical function with indices rather than subsequences. It seems like it makes for easier proof: when inserting into a `Seq`, we usually need to reason by case on `i == position`.


jhjourdan:
Support for sequences in Why3 is really poor...

PR #2174, opened 2026-06-25
Add PartialOrdLogic

lafeychine:


creusot-std/src/logic/ord.rs
```diff
@@ -1,111 +1,86 @@
...
-    #[ensures(x.le_log(y) == (x.partial_cmp_log(y) == Some(Ordering::Less) || x.partial_cmp_log(y) == Some(Ordering::Equal)))]
-    fn partial_cmp_le_log(x: Self, y: Self);
-
-    /// The logical `<` operation.
-    #[logic(open)]
-    fn lt_log(self, o: Self) -> bool {
-        pearlite! { self.partial_cmp_log(o) == Some(Ordering::Less) }
+    /// The logical `<=` operation.
+    #[logic(open, inline)]
+    fn le_log(self, other: Self) -> bool {
```
dianegolfouse, line 27: Why is `le_log` not `sealed` ?

jhjourdan: Some implementors (integers, real) have a builtin symbols for this. If we inlined the disjunction-based definition, then we would produce larger goals which would be more difficult to solve.

PR #2175, opened 2026-06-25
Weaken the autoloading mechanism of laws

jhjourdan:
The implementor of a trait may decide not to use the `law` attribute for a law. In this case, the autoloading mechanism is disabled when the trait instance can be resolved.

For laws of the Fn/FnMut/FnOnce traits, we simulate this mechanism for builtin instances, because the definition of the post/precondition predicate are enough and the laws are not needed.


jhjourdan:
This should greatly help when proving programs with a lot of closures; such as programs with invariants or logical atomicity.

jhjourdan:
image

PR #2180, opened 2026-06-30
Refine `size_of_logic`

dianegolfouse:
- Have more types satisfy `size_of::<T>() > 0`, based on discussion with the Rust team, and in particular this change to the reference that seems close to be accepted: https://github.com/rust-lang/reference/pull/2166:
  > This PR guarantees that for all repr(Rust) structs, unions, and inhabited enum variants, each field must fit in the type. [...]
  > The guarantee implies that (u8, !) can never be a ZST, fully closing that door.
- Restrict computation of `size_of_logic::<T>()` to types where the size is known to be fixed across compilation targets.

### DIFF
--- a/creusot/src/ctx.rs
+++ b/creusot/src/ctx.rs
@@ -861,15 +861,20 @@ fn is_nonzero_const<'tcx>(tcx: TyCtxt<'tcx>, len: ty::Const<'tcx>) -> bool {
     }
 }
 
-/// Determining whether a user-defined type is zero-sized is tricky because
-/// Rust makes very few guarantees about layout. For example, the compiler
-/// is free to omit constructors if it can determine that they are uninhabited.
+/// Determining whether a user-defined type is zero-sized is tricky because Rust
+/// makes very few guarantees about layout. For example, the compiler is free to
+/// omit constructors of `enum`s if it can determine that they are uninhabited.
 /// We use heuristics to find cases where some bits are definitely needed to
 /// represent this type.
 ///
-/// A `struct` or `enum` is non-zero sized if either:
-/// - there are two inhabited constructors;
-/// - there is one inhabited constructor with at least one non-zero sized field.
+/// An `enum` is non-zero sized if:
+/// - It has at least two inhabited constructors, or
+/// - It has an _inhabited_ constructor with at least one non-zero sized field.
+///
+/// A `struct` or a `union` is non-zero sized if it has
+/// at least one non-zero sized field. Note that since
+/// <https://github.com/rust-lang/reference/pull/2166>, this holds even if one
+/// of the fields is uninhabited.
 fn is_nonzero_sized_adt<'tcx>(
     tcx: TyCtxt<'tcx>,
     typing_env: TypingEnv<'tcx>,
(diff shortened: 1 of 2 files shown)


creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -1159,36 +1161,99 @@ fn fn_mut_hist_inv_term<'tcx>(
     }
 }
 
-/// Special definition for `::creusot_std::std::mem::size_of_logic`.
+/// Returns `true` if `ty`'s `size_of` has a guaranteed result, across platforms.
```
Lysxia:
```suggestion
/// Returns `true` if `ty`'s `size_of` has a guaranteed value independent of the platform.
```

creusot/src/ctx.rs
```diff
@@ -861,15 +861,20 @@ fn is_nonzero_const<'tcx>(tcx: TyCtxt<'tcx>, len: ty::Const<'tcx>) -> bool {
...
 ///
-/// A `struct` or `enum` is non-zero sized if either:
-/// - there are two inhabited constructors;
-/// - there is one inhabited constructor with at least one non-zero sized field.
+/// An `enum` is non-zero sized if:
+/// - It has at least two inhabited constructors, or
+/// - It has an _inhabited_ constructor with at least one non-zero sized field.
+///
+/// A `struct` or a `union` is non-zero sized if it has at
+/// least one non-zero sized field. Note that this since
```
Lysxia: "that this since"

dianegolfouse: 😅

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -1159,36 +1161,99 @@ fn fn_mut_hist_inv_term<'tcx>(
     }
 }
 
-/// Special definition for `::creusot_std::std::mem::size_of_logic`.
+/// Returns `true` if `ty`'s `size_of` has guaranteed value independent of the platform.
```
jhjourdan: Why is the independance of the platform important? Some specs in creusot-std already depend on the platform anyway.

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -1159,36 +1161,99 @@ fn fn_mut_hist_inv_term<'tcx>(
...
 ///
-/// The specification of sizes is documented at [`size_of`].
-/// Note: at this point we are guaranteed to have a `Sized` type.
+/// This is a best effort.
+fn guaranteed_size<'tcx>(tcx: TyCtxt<'tcx>, typing_env: TypingEnv<'tcx>, ty: Ty<'tcx>) -> bool {
+    use rustc_type_ir::{IntTy, TyKind, UintTy};
+    match ty.kind() {
+        TyKind::Bool | TyKind::Char | TyKind::Float(_) | TyKind::Never => true,
```
jhjourdan: For example, are we guaranteed that the size of `bool` is always 1? I am not so sure, actually?

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -1159,36 +1161,99 @@ fn fn_mut_hist_inv_term<'tcx>(
...
-        return Some(Term::int(ctx.int_ty(), layout.size.bytes() as i128));
-    }
+) -> bool {
+    use rustc_type_ir::{IntTy, TyKind, UintTy};
+    let is_sized_box = |def: &rustc_middle::ty::AdtDef<'tcx>, args| {
+        def.is_box()
+            && def
+                .all_field_tys(tcx)
+                .iter_instantiated(tcx, args)
+                .all(|ty| tcx.normalize_erasing_regions(typing_env, ty).is_sized(tcx, typing_env))
```
jhjourdan: by doing this, you are iterating over the size of `Unique<T>` and `A`, the two fields of the type `Box<T>` (see https://doc.rust-lang.org/src/alloc/boxed.rs.html#234-237). Instead, you need to look into `args[0]`.

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -1159,36 +1161,99 @@ fn fn_mut_hist_inv_term<'tcx>(
...
+        // `size_of::<Option<Box<T>>() == size_of::<usize>()`
+        // TODO: extend this to more types
+        TyKind::Adt(def, args) if tcx.is_lang_item(def.did(), rustc_hir::LangItem::Option) => def
+            .all_field_tys(tcx)
+            .iter_instantiated(tcx, args)
+            .all(|ty| match tcx.normalize_erasing_regions(typing_env, ty).kind() {
+                TyKind::Ref(_, ty, _) => ty.is_sized(tcx, typing_env),
+                TyKind::Adt(def, args) if is_sized_box(def, args) => true,
+                _ => false,
+            }),
```
jhjourdan: Is the compressed encoding of `Box` guaranteed?

jhjourdan (commented):
I appreciate your efforts in making Creusot proofs portable, but I think this is, unfortunately, doomed to failure.

Indeed, we already depend on the architecture in many places:
- The model of `usize` depends on the architecture, and we are not going to make it generic, because then many proofs will fail.
- Typechecking sometimes evaluate `const` values, and `size_of` is const (hence, typechecking is target-dependent, which you may find ugly, but this is the way it is)
- There are already quite a few instances of `#[cfg_attr(target_pointer_width = "16", ...)]` in creusot-std

So let's use the value computed by Rustc for `size_of` when it succeeds doing so, even if the result is not portable.

creusot/src/backend/clone_map/elaborator.rs
dianegolfouse: We are: https://doc.rust-lang.org/std/mem/fn.size_of.html

creusot/src/backend/clone_map/elaborator.rs
dianegolfouse: Meaning, does `Option<Box<T>> where T: Sized` always fit a `usize`? Yes, this is guaranteed. (https://doc.rust-lang.org/std/mem/fn.size_of.html)

dianegolfouse:
I unfortunately agree about platform independance, this is probably a lost battle; and we can say "verify with Creusot on each architecture you want to release on".
However there is still an issue with types whose layout is not guaranteed: I am not really comfortable saying that a type has a given size, when the next compiler version might change this...
Actually, I'm not even sure that `size_of` is stable on a given compiler version.

jhjourdan:
> I unfortunately agree about platform independence, this is probably a lost battle; and we can say "verify with Creusot on each architecture you want to release on". However there is still an issue with types whose layout is not guaranteed: I am not really comfortable saying that a type has a given size, when the next compiler version might change this...

I agree this is a problem, but again, typechecking (including trait resolution) depends on const evaluation, so this is a lost battle. We may open an issue about that, but basically this means that Creusot is really sound only if the code is linked with the exact same version of Rustc that is used for compilation.

> Actually, I'm not even sure that `size_of` is stable on a given compiler version.

I would really be surprised if that would not be the case (of course, we are assuming same compiler version with the same configuration)!

dianegolfouse:
After our discussion, I agree: I opened #2186 to track this, and slightly changed the doc of `size_of_logic`, but the computation of `size_of_logic` stays as it was.

jhjourdan (approved):
LGTM!

PR #2185, opened 2026-07-02
Fix `char` upper bound in Coma prelude

niooss-ledger:
Hello,

The documentation of Rust primitive types (https://doc.rust-lang.org/std/primitive.char.html#validity-and-layout , https://github.com/rust-lang/rust/blob/1.96.1/library/core/src/primitive_docs.rs#L336-L339) states that:

> A char is a ‘Unicode scalar value’, which is any ‘Unicode code point’ other than a surrogate code point. This has a fixed numerical definition: code points are in the range 0 to 0x10FFFF, inclusive.

The Coma prelude was excluding `char::MAX = '\u{10FFFF}'`. Include the upper bound in the inequalities.


jhjourdan:
Thanks!

jhjourdan:
Could you please run `cargo test --test why3` to update the obsolete session file?

niooss-ledger:
Done. For information, I ran `./t all` before opening this Pull Request and it succeeded. For information, I read `CONTRIBUTING.md` beforehand, but thought `all` included Why3 tests. It seems `./t all --why3-all` is the command I should have run.

PR #2190, opened 2026-07-02
Fix upper bound of integer shift operations in Coma prelude

niooss-ledger:
In Rust, the bit-shift operations are considered to overflow when the shift is greater or equal to the number of bits. This is documented for example in https://doc.rust-lang.org/reference/expressions/operator-expr.html#r-expr.operator.int-overflow.shift and in methods such as [`u8::strict_shl`](https://doc.rust-lang.org/std/primitive.u8.html#method.strict_shl):

> Computes `self << rhs`, panicking if `rhs` is larger than or equal to the number of bits in `self`.

The Coma prelude considered shifting by the number of bits as non-overflowing. This means Creusot verified this program even though it panics when built in debug in mode:

```rust
use creusot_std::prelude::*;

#[requires(s@ <= 8)]
#[ensures(true)]
pub fn shr_one(s: u32) -> u8 {
    1u8 >> s
}

#[ensures(true)]
fn main() {
    shr_one(8);
}
```

Change the upper bound of overflow checks to use a strict inequality.

I also tried adding a test to cover this case, in ` tests/should_fail/bug/int-shift-full.rs`. Why3 successfully proved the generated Coma file before the PR, and no longer proved it after. I also verified that `./t all --why3-all` succeeds.

_Disclaimer of AI usage:_ Like https://github.com/creusot-rs/creusot/pull/2188, this error was found by Claude Code running model Sonnet 4.6 while reviewing another PR I did (https://github.com/creusot-rs/creusot/pull/2185). I prompted "There was an error in prelude.coma. I am fixing it. What other errors?". I wrote the test and the fix myself, without any AI tool.


jhjourdan:
Well spotted.

PR #2191, opened 2026-07-02
Add external specifications for traits `AsRef` and `AsMut`

### LINKED ISSUE #2181: Specs for `AsRef` are missing
Hello,

While trying to add contracts to function using `AsRef<[u8]>`, I encountered a warning because some specifications are missing. Here is a minimal reproducer, using method `len` (the [actual code from `hex`](https://github.com/KokaKiwi/rust-hex/blob/e25a8701f7f6b5ec4acedffcb4a45cfb69cc34eb/src/lib.rs#L223-L232) is a bit more complex):

```rust
pub fn get_len_through_as_ref<T: AsRef<[u8]>>(bytes: T) -> usize {
    bytes.as_ref().len()
}

#[ensures(result@ == bytes@.len())]
pub fn get_len_of_slice(bytes: &[u8]) -> usize {
    get_len_though_as_ref(bytes)
}
```

Creusot reports a warning in `get_len_through_as_ref`:

```text
warning: calling external function `as_ref` with no contract will yield an impossible precondition
```

Adding some specifications in an `extern_spec!` fixes the warning:

```rust
use creusot_std::prelude::*;

extern_spec! {
    mod core {
        mod convert {
            trait AsRef<U> where Self: AsRef<U>, U: ?Sized {
                fn as_ref(&self) -> &U;
            }
        }
    }

    impl<T> AsRef<[T]> for [T] {
        #[check(ghost)]
        #[ensures(result == self)]
        fn as_ref(&self) -> &[T] {
            self
        }
    }

    impl<'a, T: ?Sized, U: ?Sized> AsRef<U> for &'a T where T: AsRef<U> {
        #[ensures(<T as AsRef<U>>::as_ref.postcondition((*self,), result))]
        fn as_ref<'b>(&'b self) -> &'b U {
            <T as AsRef<U>>::as_ref(*self)
        }
    }
}

#[ensures(exists<s: &[u8]> <T as AsRef<[u8]>>::as_ref.postcondition((&bytes,), s) && result@ == s@.len())]
pub fn get_len_through_as_ref<T: AsRef<[u8]>>(bytes: T) -> usize {
    bytes.as_ref().len()
}

#[ensures(result@ == bytes@.len())]
pub fn get_len_of_slice(bytes: &[u8]) -> usize {
    get_len_through_as_ref(bytes)
}
```

(And `cargo creusot` succeeds in verifying all contracts)

Should I open a Pull Request to add such specifications to `creusot-std`? Or would you prefer adding the missing specifications yourself?

niooss-ledger:
Copy the traits and impls from https://github.com/rust-lang/rust/blob/1.96.1/library/core/src/convert/mod.rs

Since Rust commit https://github.com/rust-lang/rust/commit/884d0e031a046af894b43180032c1803bb6d0834 (Rust 1.89.0 on 2025-08-07), traits `AsRef` and `AsMut` have been using `PointeeSized` instead of `?Sized`. While declaring the traits with `PointeeSized` work, it doesn't for the implementations. For example, using

```rust
impl<'a, T: PointeeSized, U: PointeeSized> AsRef<U> for &'a T
where
    T: AsRef<U>,
{
    #[ensures(<T as AsRef<U>>::as_ref.postcondition((*self,), result))]
    fn as_ref<'b>(&'b self) -> &'b U {
        <T as AsRef<U>>::as_ref(*self)
    }
}
```

makes Creusot report:

```text
error[E0277]: the size for values of type `T` cannot be known
  --> /creusot/creusot-std/src/std/convert.rs:31:9
   |
31 |         <T as AsRef<U>>::as_ref(*self)
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ doesn't have a known size

For more information about this error, try `rustc --explain E0277`.
```

This Pull Request introduces `PointeeSized` anyway, enabling unstable feature `sized_hierarchy` in `creusot-std`.

Fixes: https://github.com/creusot-rs/creusot/issues/2181

## PR Status

A test fails:

```console
$ cargo test -p why3tests --test why3 -- creusot-std
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.04s
     Running tests/why3.rs (target/debug/deps/why3-fd0f9d6076955d90)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/prelude-generator`
Building prelude...
Testing tests/creusot-std/creusot-std.coma ... failure
******** STDOUT ********
 1678/1685 (replay failed, with some merge miss)
******** STDERR ********
[Warning] found detached goals or theories or transformations
Replay failed.1                           
************************
Failure!
error: test failed, to rerun pass `-p why3tests --test why3`

Caused by:
  process didn't exit successfully: `/creusot/target/debug/deps/why3-fd0f9d6076955d90 creusot-std` (exit status: 1)
```

I am trying to understand whether the added annotations are wrong or whether I need to invoke something to update some files.


jhjourdan:
> 
> makes Creusot report:
> 
> ```
> error[E0277]: the size for values of type `T` cannot be known
>   --> /creusot/creusot-std/src/std/convert.rs:31:9
>    |
> 31 |         <T as AsRef<U>>::as_ref(*self)
>    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ doesn't have a known size
> 
> For more information about this error, try `rustc --explain E0277`.
> ```
> 
> This Pull Request introduces `PointeeSized` anyway, enabling unstable feature `sized_hierarchy` in `creusot-std`.
> 

Perhaps there is a missing feature that need to be enabled so that Rust stop flagging this as an error?

In any case, I would prefer if the extern specs were as faithful as possible to `std`. So, if you cannot find what should be done to fix this sizedness issue, you can simply remove the body of these extern specs. Function bodies of extern specs are just here for checking that the specs are sound. If you don't put the bodies, the specs are still used.

> 
> I am trying to understand whether the added annotations are wrong or whether I need to invoke something to update some files.

For historical reasons, the creusot-std uses Why3 IDE instead of Why3find for managing the proofs. You need to open the ide with `./ide tests/creusot-std/creusot-std`, then fix the proofs (most often, it is enough to simply select the failed goal and the press "4") and finally save the session.

creusot-std/src/std/convert.rs
```diff
@@ -1,17 +1,94 @@
...
+        #[check(ghost)]
+        #[ensures(result == self && ^self == *self)]
+        fn as_mut(&mut self) -> &mut [T] {
+            self
+        }
+    }
+
+    impl AsMut<str> for str {
+        #[check(ghost)]
+        #[ensures(result == self && ^self == *self)]
```
Lysxia:
```suggestion
        #[ensures(result == self)]
```

`^self == *self` means `self` is not mutated even through the mutable borrow(s) returned by the function.

creusot-std/src/std/convert.rs
```diff
@@ -1,17 +1,94 @@
...
+            self
+        }
+    }
+
+    impl<'a, T: ?Sized, U: ?Sized> AsMut<U> for &'a mut T
+    where
+        T: AsMut<U>,
+    {
+        #[requires(<T as AsMut<U>>::as_mut.precondition((*self,)))]
+        #[ensures(^self == *self)]
```
Lysxia:
```suggestion
        #[ensures(*^self == **self && ^^self == ^*self)]
```

Looking at the coma, it appears that `*self` is getting reborrowed instead of passed straight to the inner `as_mut`.

@jhjourdan does that look right to you?

jhjourdan:
> Perhaps there is a missing feature that need to be enabled so that Rust stop flagging this as an error?

After a bit of debugging, it seems like its because the extern spec declaration of `AsRef` assumes a trait bound that is wrong in the impl. I don't understand the magic behind the sized hierarchy.

creusot-std/src/std/convert.rs
```diff
@@ -1,17 +1,94 @@
...
+        #[ensures(<T as AsMut<U>>::as_mut.postcondition((*self,), result))]
+        fn as_mut<'b>(&'b mut self) -> &'b mut U {
+            //(*self).as_mut()
+            <T as AsMut<U>>::as_mut(*self)
+        }
+    }
+
+    impl<T> AsMut<[T]> for [T] {
+        #[check(ghost)]
+        #[ensures(result == self && ^self == *self)]
```
jhjourdan:
```suggestion
        #[ensures(result == self)]
```

Same here.

creusot-std/src/std/convert.rs
```diff
@@ -1,17 +1,94 @@
...
+        }
+    }
+
+    impl<'a, T: ?Sized, U: ?Sized> AsMut<U> for &'a mut T
+    where
+        T: AsMut<U>,
+    {
+        #[requires(<T as AsMut<U>>::as_mut.precondition((*self,)))]
+        #[ensures(^self == *self)]
+        #[ensures(<T as AsMut<U>>::as_mut.postcondition((*self,), result))]
```
jhjourdan:
```suggestion
        #[ensures(^*self == ^^self)]
        #[ensures(exists<s: &mut T> *s == **self && ^s == *^self &&
                           <T as AsMut<U>>::as_mut.postcondition((s,), result))]
```

This is indeed a reborrow, hence a simple resolve of  is not enough, because `*self` can be changed after the end of `'b`.

creusot-std/src/std/convert.rs
niooss-ledger: I agree. I added it in an attempt to fix the proofs, and in the end it was not needed.

creusot-std/src/std/convert.rs
niooss-ledger: Thanks for the solution! I was trying to debug things with `.postcondition_mut`, and this `exists` statement was exactly what was missing to get the proof to work :)

niooss-ledger:
Thanks for your reviews!
I applied the suggestions and now all tests pass :)

About `PointeeSized`, It could be applied more... except in `impl<'a, T: ?Sized, U: ?Sized> AsMut<U> for &'a mut T` where changing to `T: PointeeSized` triggered some errors in `creusot-std/src/logic/ops.rs` such as:

```text
error[E0277]: the size for values of type `T` cannot be known
  --> /w/creusot/creusot-std/src/std/convert.rs:65:54
   |
65 |         #[ensures(exists<s: &mut T> *s == **self && ^s == *^self &&
   |                                                     -^
   |                                                     ||
   |                                                     |the trait `logic::ops::Fin` is not implemented for `&mut T`
   |                                                     required by a bound introduced by this call
   |
   = note: the trait bound `&mut T: logic::ops::Fin` is not satisfied
note: required for `&mut T` to implement `logic::ops::Fin`
  --> /w/creusot/creusot-std/src/logic/ops.rs:28:17
   |
28 | impl<T: ?Sized> Fin for &mut T {
   |      -          ^^^     ^^^^^^
   |      |
   |      unsatisfied trait bound introduced here
```

The "fix" would be to replace all occurences of `?Sized` with `PointeeSized` in [`creusot-std/src/logic/ops.rs`](https://github.com/creusot-rs/creusot/blob/bef58f6aa7493ac8c8012164a8eeab462c346d1a/creusot-std/src/logic/ops.rs) and [`creusot-std/src/stubs.rs`](https://github.com/creusot-rs/creusot/blob/bef58f6aa7493ac8c8012164a8eeab462c346d1a/creusot-std/src/stubs.rs). But this would make using `creusot-std` require a nightly compiler (I tested on a simple project generated with `cargo creusot init` and `cargo build` with a stable toolchain no longer works: it would require a nightly toolchain and enabling feature `nightly` in dependency `creusot-std`).

Moreover, there is another issue with `PointeeSized`. Let's consider a new project with this `src/lib.rs`:

```rust
#![feature(sized_hierarchy)]
use creusot_std::prelude::*;
use core::marker::PointeeSized;

pub trait MyAsMut<T: PointeeSized>: PointeeSized {
    fn as_mut(&mut self) -> &mut T;
}

impl<'a, T: PointeeSized, U: PointeeSized> MyAsMut<U> for &'a mut T
where
    T: MyAsMut<U>,
{
    fn as_mut<'b>(&'b mut self) -> &'b mut U {
        <T as MyAsMut<U>>::as_mut(*self)
    }
}
```

Building with `cargo +nightly build` succeeds. But launching Creusot fails:

```text
$ cargo creusot
...

error: Unsupported type: &mut U
  --> src/lib.rs:14:40
   |
14 |         <T as MyAsMut<U>>::as_mut(*self)
   |                                        ^
```

(and replacing `PointeeSized` with `?Sized` in this example fixes this error). This "Unsupported type" which only appears with `cargo creusot` is quite puzzling.

Anyway, this is why I preferred to stick with `?Sized` in the implementations, which also enabled keeping the body of the functions.

jhjourdan:
Fair enough, let's stick with `?Sized` then.

PR #2194, opened 2026-07-09
Detect recursion between logic functions and non-terminating program functions or const items

### LINKED ISSUE #1588: Termination check does not take into account requires/ensures clauses
It's not clear to me whether one can create unsoundnesses from this, but at least this can create cycles in the clone map, which should not happen.

### LINKED ISSUE #2193: Recursion through `const`
`const` definitions are not supposed to be recursive and Rust rejects many of them, but not always: Rust allows recursive occurrences inside unused closures such as those emitted by `proof_assert!` (or `ensures` with #2054).

Counterexample that works on master:

```rust
extern crate creusot_std;
use creusot_std::prelude::*;

pub struct False;

impl Invariant for False {
    #[logic]
    fn invariant(self) -> bool {
        false
    }
}

pub const C: False = {
    proof_assert!{ let _ = f(); false };
    False
};

#[logic]
#[ensures(false)]
pub fn f() {
    proof_assert!( let _ = C; false );
}
```

The fix is for the termination checker to track dependencies of `const`, including *logic* functions.

Lysxia:
Close #2193, close #1588

Previously, program functions and const items were simply ignored by the termination checker. However those can be used in logic functions (program functions via `f.postcondition`, and consts are always accessible).
See `tests/should_fail/terminates/const.rs` and `tests/should_fail/terminates/ensures.rs`.

When visiting the body of a function, we now distinguish recursive calls (which may be allowed in terminating functions given a variant) from other recursive occurrences (which lose track of the variant).

For terminating program functions, this also fixes a bug where their pre- and postconditions could refer to themselves.
See `tests/should_fail/terminates/ensures_terminates.rs`.


jhjourdan:
Does this also fix #1588?

Lysxia:
Yes it does!

jhjourdan:
Thanks. We do not often two soundness bugs in the same PR...

PR #2213, opened 2026-07-31
Update toolchain to nightly-2026-07-29

Lysxia:
Main changes

- Some Rust intrinsics can only be used in `const` blocks and this is now strictly enforced. This affects us because we have extern specs on them (because these intrinsics are exposed in constants; otherwise users are not supposed to use these intrinsics directly), so the generated wrapper must now wrap its body in a `const` block. I have added an attribute to that effect.
- `TyKind::Alias` now has a `IsRigid` field that indicates whether the alias is "rigid". If I understand correctly, normalization either unfolds the alias (when possible) or sets that flag to `IsRigid::Yes`, whereas instantiation (`EarlyBinder::bind` + `instantiate`) sets that flag everywhere to `IsRigid::No`. Basically this means we have to call normalization more consistently than before to correctly identify types (we should never try to translate an alias with `IsRigid::No`).
- Only one test changed: `examples/iterators/04_skip.coma` has an extra resolve statement which looks like it should have been there from the beginning, so I deduce that was a bugfix in Rust's borrowchecker.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -171,7 +171,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-creusot"
-version = "0.13.0"
+version = "0.14.0-dev"
 dependencies = [
  "anyhow",
  "cargo_metadata",
@@ -293,7 +293,7 @@ dependencies = [
 
 [[package]]
 name = "creusot"
-version = "0.13.0"
+version = "0.14.0-dev"
 dependencies = [
  "creusot-args",
  "creusot-metadata",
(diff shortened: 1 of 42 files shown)


jhjourdan:
> - Only one test changed: `examples/iterators/04_skip.coma` has an extra resolve statement which looks like it should have been there from the beginning, so I deduce that was a bugfix in Rust's borrowchecker.

That is weird actually. This is resolving variable `r` whose contents serves as a return value of the function just after. This is correct in this case because `r` contains `None`, but I'd like to make sure nothing bad happens here in general.

PR #2214, opened 2026-08-03
Properly error on async

### LINKED ISSUE #2212: ICE: 'entered unreachable code' in pre_sig_of on any crate containing an async fn
Any crate that loads `creusot_std` and contains an `async fn` aborts the compiler, regardless of what is being verified.

```
thread 'rustc' panicked at creusot/src/translation/specification.rs:423:61:
internal error: entered unreachable code
error: the compiler unexpectedly panicked. This is a bug
```

## Minimal reproduction

`Cargo.toml`:

```toml
[package]
name = "cr_asyncf"
version = "0.1.0"
edition = "2021"

[dependencies]
creusot-std = "0.12"
```

`src/lib.rs`:

```rust
use creusot_std::prelude::*;

pub async fn an_async_fn(x: u32) -> u32 {
    x + 1
}

pub fn probe() {
    proof_assert! { true };
}
```

`cargo creusot` → ICE.

Replacing the `async fn` with a closure is enough to make it go away — the same crate with

```rust
pub fn uses_a_closure(v: Vec<u32>) -> usize {
    v.iter().filter(|x| **x > 2).count()
}
```

compiles and reaches a real proof goal (`Goal Coma.vc_uses_a_closure: ✘ (1/2)`, as expected for an uncontracted external call). So it is specific to coroutines, not to closures generally.

The `proof_assert!` is incidental — it is only there because `creusot_std` has to be loaded for the crate to be translated at all. Nothing in the harness refers to the async function.

## Cause

`pre_sig_of`:

```rust
if ctx.def_kind(def_id) == DefKind::Closure {
    let fn_ty = ctx.type_of(def_id).instantiate_identity().skip_normalization();
    let TyKind::Closure(_, subst) = fn_ty.kind() else { unreachable!() };   // <-- 423:61
```

`DefKind::Closure` also covers coroutines and coroutine-closures — an `async fn` body is a coroutine, so it answers that def-kind while its `TyKind` is `Coroutine`, and the `else` arm fires. The closure handling that follows (deriving `ClosureKind`, building `closure_env_ty`, substituting `self` into pre/post conditions) has nothing to operate on for a coroutine anyway, since there is no closure environment.

## Suggested fix

Selecting on the type rather than the def-kind lets coroutines fall through with their signature unchanged, and leaves the closure path exactly as it was:

```rust
let closure_ty = (ctx.def_kind(def_id) == DefKind::Closure)
    .then(|| ctx.type_of(def_id).instantiate_identity().skip_normalization())
    .filter(|ty| matches!(ty.kind(), TyKind::Closure(..)));

if let Some(fn_ty) = closure_ty {
    let TyKind::Closure(_, subst) = fn_ty.kind() else { unreachable!() };
    // ... unchanged ...
```

I am testing this locally and can open a PR if the approach looks right to you. If the preferred behaviour is instead a diagnostic saying Creusot does not support coroutines, that would also be a large improvement over the ICE — the current failure gives no indication that `async` is involved.

## Versions

- Creusot `v0.12.0`; `master` at `318615b` ("Release 0.13.0") carries the identical `unreachable!()` at that line, so this does not appear to be fixed there.
- `rustc 1.97.0-nightly (66da6cae1 2026-04-20)`, toolchain `nightly-2026-04-21`.
- `aarch64-unknown-linux-gnu`.

## Why it matters beyond the repro

It rules out any crate with async anywhere in it — a web service, anything on tokio — even when the properties being verified are in ordinary synchronous code and never touch the async parts. That was how we hit it: a synchronous decision function we wanted to verify, in a crate that happens to also contain an axum server.

Lysxia:
Close #2212

### DIFF
--- a/creusot/src/translation.rs
+++ b/creusot/src/translation.rs
@@ -128,6 +128,9 @@ fn force_translation(ctx: &TranslationCtx) {
         if is_no_translate(ctx.tcx, def_id) && !is_extern_spec(ctx.tcx, def_id) {
             continue;
         }
+        if ctx.def_kind(def_id) == DefKind::Closure {
+            continue;
+        }
         let _ = ctx.sig(def_id);
         if is_logic(ctx.tcx, def_id) && !is_opaque(ctx.tcx, def_id) {
             let _ = ctx.term(def_id);
--- a/creusot/src/translation/specification.rs
+++ b/creusot/src/translation/specification.rs
@@ -429,12 +429,21 @@ pub(crate) fn pre_sig_of<'tcx>(ctx: &TranslationCtx<'tcx>, def_id: DefId) -> Pre
     let mut presig = contract_of(ctx, def_id);
 
     if ctx.def_kind(def_id) == DefKind::Closure {
-        let fn_ty = ctx.type_of(def_id).instantiate_identity();
-        let fn_ty = ctx.normalize_erasing_regions(ctx.typing_env(def_id), fn_ty);
-        let TyKind::Closure(_, subst) = fn_ty.kind() else { unreachable!() };
-
-        let kind = subst.as_closure().kind();
+        let fn_ty = ctx.normalize_erasing_regions(
+            ctx.typing_env(def_id),
+            ctx.type_of(def_id).instantiate_identity(),
+        );
+        let subst = match fn_ty.kind() {
+            TyKind::Closure(_, subst) => subst,
+            TyKind::Coroutine(..) | TyKind::CoroutineClosure(..) => {
+                ctx.crash_and_error(ctx.def_span(def_id), "async is not yet supported")
+            }
+            _ => unreachable!(),
+        };
+        let closure = subst.as_closure();
+        let kind = closure.kind();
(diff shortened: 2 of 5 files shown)


jhjourdan:
Does it crashes with a trusted async function? It should not.

Lysxia:
> Does it crashes with a trusted async function? It should not.

Fixed!

creusot/src/translation.rs
```diff
@@ -128,6 +128,9 @@ fn force_translation(ctx: &TranslationCtx) {
         if is_no_translate(ctx.tcx, def_id) && !is_extern_spec(ctx.tcx, def_id) {
             continue;
         }
+        if ctx.def_kind(def_id) == DefKind::Closure {
```
jhjourdan, line 131: This fix is a bit risky, because it will disable forcing signature evaluation for async functions when we will support them.

Could you at least document that we are doing that for async functions?

Lysxia: The toplevel async function still gets processed here. This only disables that pass for closure-like things, such as the coroutine that the async function returns.

jhjourdan: I see... The encoding of coroutines seems particularly complex.

PR #2218, opened 2026-08-04
Logical functions aliases, reloaded

mcoulmance:
This PR continues the work started by @dianegolfouse in #1346. It adds two new features to creusot:
- logic aliases between program functions and logic functions. A logic alias can be specified either by a 
  path to a logic function, or by a call to a logic function
- hybrid functions, which is basically a mechanism to simultaneously declare a program function and an identical logic
  function, with a logic alias between them.

# Example

```rust
// An alias defined with a path to a logic function
#[logic(open)]
fn is_zero_logic(a: i64) -> bool {
    a == 0i64
}

#[logic_alias(is_zero_logic)]
fn is_zero(x: i64) -> bool {
    x == 0
}



// An alias defined with a call to a logic function
#[logic(open)]
fn is_one_logic(a: Int) -> bool {
    pearlite! {
        a == 1
    }
}

#[logic_alias(is_one_logic(x@))]
fn is_one(x: i64) -> i64 {
    x == 1
}


// A hybrid declaration
#[hybrid]
#[requires(x@ < i64::MAX@ - 1)]
fn add_one(x: i64) -> i64 {
    x + 1i64
}


fn foo() {
  let mut i = 0;
  proof_assert!(is_zero(i));
  i = add_one(i);
  proof_assert!(is_one(i));
  proof_assert!(add_one(i) == 2i64);
}
```

It is still quite experimental, but it is already possible to use aliasing in traits. I would like to open a discussion on this PR about the semantic we want to give to aliasing in traits, since it is a pretty important design decision (see header doc in creusot/src/logic_alias.rs).


creusot/src/validate/purity.rs
```diff
@@ -142,6 +143,26 @@ pub(crate) fn validate_purity<'tcx>(
...
+                .raise_fatal();
+        }
+        let logic_id = logic_alias::get_logic_id(ctx, alias_id);
+        if !is_logic(ctx.tcx, logic_id) {
+            ctx.dcx()
+                .struct_span_err(span, "Only logic functions can be aliased")
+                .with_note(format!("`{}` is not a logic function", ctx.def_path_str(logic_id)))
+                .emit()
+                .raise_fatal();
+        }
```
Lysxia, line 164: Isn't this already implied by the purity check of ensures clauses?

creusot-std-proc/src/creusot/specs.rs
```diff
@@ -261,6 +284,93 @@ pub fn bitwise_proof(_: TS1, tokens: TS1) -> TS1 {
...
+            Ok(Term::Call(term_call)) => {
+                quote!(result == #term_call)
+            }
+            Ok(invalid_term) => {
+                return syn::Error::new(
+                   invalid_term.span(),
+                   "`logic_alias` should contain a path to a logic function with the same signature, or \
+                   a logic function call with same return type"
+               ).to_compile_error().into();
+            }
```
Lysxia: Maybe we should allow arbitrary terms here. This would simplify the current code that assumes logic aliases are function calls, which required adding a `raw_term` function.

 In particular, an identity builtin can then be written as a logical alias

```rust
#[logic_alias((x))]
fn identity(x: T) { x }
```

In this example I'm using extra parentheses around `x` to make the expression not a path which would be interpreted as a function name. Another idea is to use a different syntactic form `#[logic_alias = x]` (I'm not sure this is legal) or `#[logic_alias(fn path)]` (with `fn` as a keyword).

creusot/src/logic_alias.rs
```diff
@@ -0,0 +1,537 @@
...
+//! |  impl type        |   impl trait      |    valid if it's the same type                     |
+//! |-------------------+-------------------+----------------------------------------------------|
+//! |  impl trait       |   fun             |    valid if `fun` comes from the current crate     |
+//! |-------------------+-------------------+----------------------------------------------------|
+//! |  impl trait       |   trait           |    valid if it's the same trait (**)               |
+//! |-------------------+-------------------+----------------------------------------------------|
+//! |  impl trait       |   impl type       |    valid if `type` implements `trait`              |
+//! |-------------------+-------------------+----------------------------------------------------|
+//! |  impl trait       |   impl trait      |    valid if it's the same trait (**)               |
+//! +-------------------+-------------------+----------------------------------------------------+
```
Lysxia: What's the motivation for the restrictions in the third column?

I do agree that we should only allow aliases on program functions of the current crate, this prevents having two crates defining inconsistent aliases on an external function.

I don't see a reason to restrict the RHS (the "logic part") of the alias.

And we don't want logic aliases in "impl trait" because it makes aliasing depend on type instantiations which is extremely tricky both to implement and to understand.

creusot-std-proc/src/creusot/specs.rs
jhjourdan, line 173: What about consts? Either handle logic_alias or emit an error.

creusot-std-proc/src/creusot/specs.rs
```diff
@@ -261,6 +284,93 @@ pub fn bitwise_proof(_: TS1, tokens: TS1) -> TS1 {
...
+            Err(err) => {
+                return syn::Error::new(
+                    err.span(),
+                    "`logic_alias` should contain a path to a logic function with the same signature, or \
+                    a logic function call with same return type"
+                ).to_compile_error().into();
+            }
```
jhjourdan, line 322:
```suggestion
            r => {
                let span = r.map_or_else(|t| t.span(), |e| e.span());
                return syn::Error::new(
                   span,
                   "`logic_alias` should contain a path to a logic function with the same signature, or \
                   a logic function call with same return type"
               ).to_compile_error().into();
            }
```

creusot-std-proc/src/creusot/specs.rs
```diff
@@ -261,6 +284,93 @@ pub fn bitwise_proof(_: TS1, tokens: TS1) -> TS1 {
...
+        }
+    };
+
+    ensures_inner(ensures_contract.into(), tokens, true)
+}
+
+pub fn hybrid(attrs: TS1, tokens: TS1) -> TS1 {
+    match syn::parse::<ContractSubject>(tokens.clone()) {
+        Ok(ContractSubject::FnOrMethod(prog)) => {
```
jhjourdan, line 331:
```suggestion
    match syn::parse::<FnOrMethod>(tokens.clone()) {
        Ok(prog) =>
```

creusot-std-proc/src/creusot/specs.rs
```diff
@@ -261,6 +284,93 @@ pub fn bitwise_proof(_: TS1, tokens: TS1) -> TS1 {
...
+                sig: prog.sig.clone(),
+                body: prog.body.clone(),
+                semi_token: prog.semi_token.clone(),
+            };
+            log.sig.ident = logic_ident.clone();
+            let logic_alias = if log.sig.receiver().is_some() {
+                quote!(Self::#logic_ident)
+            } else {
+                quote!(#logic_ident)
+            };
```
jhjourdan, line 347: That is not a good criterion: there exist methods in impl blocks that do not have a receiver (i.e., static methods).

jhjourdan: Also, have you tested polymorphic hybrid functions?

creusot-std-proc/src/creusot/specs.rs
```diff
@@ -261,6 +284,93 @@ pub fn bitwise_proof(_: TS1, tokens: TS1) -> TS1 {
...
+            let attrs = TokenStream::from(attrs);
+            quote! {
+                #[logic(#attrs)]
+                #log
+
+                #[logic_alias(#logic_alias)]
+                #prog
```
jhjourdan, line 354: What's happening if `#[hybrid]` is used before `#[ensures` or `#[requires` clauses?
What's happening if `#[hybrid]` is used after `#[ensures` or `#[requires` clauses?

It seems to me that these would not be the same behavior, which is a big problem.

creusot-std/src/ghost/resource/auth.rs
```diff
@@ -71,7 +71,7 @@ impl<R: UnitRA> Authority<R> {
     ///
     /// This is the same as [`Self::id`], but for ghost code.
     #[check(ghost)]
-    #[ensures(result == self.id())]
+    #[logic_alias(Self::id(*self))]
     pub fn id_ghost(&self) -> Id {
```
jhjourdan, line 75:
```suggestion
    #[logic_alias(Self::id_logic(*self))]
pub fn id(&self) -> Id {
```

Now that we can use `id` both in ghost and pearlite code, let's follow the same convention as everywhere.

creusot-std/src/ghost/resource/auth.rs
```diff
@@ -166,7 +166,7 @@ impl<R: UnitRA> Fragment<R> {
     ///
     /// This is the same as [`Self::id`], but for ghost code.
     #[check(ghost)]
-    #[ensures(result == self.id())]
+    #[logic_alias(Self::id(*self))]
```
jhjourdan, line 169:
```suggestion
    #[logic_alias(self.id())]
```

Or perhaps it's not possible for some reason? Why?

creusot-std/src/logic/fmap.rs
```diff
@@ -278,14 +278,15 @@ impl<K, V> FMap<K, V> {
...
     #[check(ghost)]
-    #[ensures(result == self.len())]
+    #[logic_alias(Self::len(*self))]
     pub fn len_ghost(&self) -> Int {
         panic!()
     }
 
     #[trusted]
     #[check(ghost)]
     #[ensures(result == (self.len() == 0 && forall<k> !self.contains(k)))]
```
jhjourdan:
```suggestion
    #[ensures(result == (self.len() == 0))]
```

AND this should be moved to the logical version.

creusot-std/src/lib.rs
```diff
@@ -358,6 +358,11 @@ pub mod macros {
     /// ```
     pub use base_macros::logic;
 
+    /// TODO(mael) Document this
```
jhjourdan, line 362: Indeed.

creusot-std/src/logic/fmap.rs
jhjourdan: (I prepared the work at #2237 )

examples/message_passing/relacq_options.coma
```diff
@@ -430,8 +447,13 @@ module M_message_passing
...
   
-  let new_Perm_PermCell_i32 (x: t_Perm_PermCell_i32) (return (x'0: t_Perm_PermCell_i32)) = any
-    [ return (result: t_Perm_PermCell_i32) -> {[@stop_split] [@expl:new ensures] result = x} (! return {result}) ]
+  function new_Perm_PermCell_i32 (x: t_Perm_PermCell_i32) : t_Perm_PermCell_i32
+  
+  axiom new_Perm_PermCell_i32_spec: forall x: t_Perm_PermCell_i32 [new_Perm_PermCell_i32 x]. new_Perm_PermCell_i32 x = x
+  
+  let new_Perm_PermCell_i32'0 (x: t_Perm_PermCell_i32) (return (x'0: t_Perm_PermCell_i32)) = any
+    [ return (result: t_Perm_PermCell_i32) -> {[@stop_split] [@expl:new ensures] result = new_Perm_PermCell_i32 x}
+      (! return {result}) ]
```
jhjourdan: These diffs should be removed by #2238. Please check.

creusot/src/translation/pearlite.rs
```diff
@@ -168,6 +168,8 @@ pub enum TermKind<'tcx> {
         id: DefId,
         subst: GenericArgsRef<'tcx>,
     },
+    /// Be careful when building this case, that it respects
+    /// [logic aliases](TranslationCtx::logical_alias)
```
jhjourdan, line 172: Make things more explicitely: what should we care about?

creusot/src/validate/purity.rs
jhjourdan: Indeed. But I agree with @mcoulmance that it is a better error message.

creusot/src/translation/pearlite/from_thir.rs
```diff
@@ -434,6 +435,12 @@ impl<'tcx> ThirTerm<'_, 'tcx> {
                             .iter()
                             .map(|arg| self.expr_term(*arg))
                             .collect::<Result<_, _>>()?;
+
+                        let (id, args, subst) = match logic_alias::subst_call(self.ctx, self.typing_env, id, subst, args.clone()) {
+                            Some((aid, aargs, asubst)) => (aid, aargs, asubst),
+                            None => (id, args, subst)
+                        };
+
```
jhjourdan, line 443: Do we check somewhere that all this is well-typed? I.e., does the alias have the same type as the original function?

creusot/src/validate/purity.rs
```diff
@@ -272,6 +293,13 @@ impl<'a, 'tcx> Visitor<'a, 'tcx> for PurityVisitor<'a, 'tcx> {
                             .unwrap();
 
                     let fn_purity = self.purity(func_did, args);
+                    let base_purity = fn_purity; // backup to get more consistent error messages
+                    let fn_purity = match self.ctx.logic_alias(func_did) {
+                        Some((_, alias_id)) if !self.context.can_call(fn_purity) => {
+                            self.purity(logic_alias::get_logic_id(self.ctx, alias_id), args)
+                        }
+                        _ => fn_purity,
+                    };
```
jhjourdan, line 302: I don't quite understand this logic. Why cannot you do the replacement with the alias only in logic mode?

Then, we would not have to do the `base_purity` backup.

creusot/src/validate/purity.rs
jhjourdan, line 310: One of the benefit of logical function aliases is that we should be able to get rid of these special cases by adding a logical alias to `deref` and `deref_mut`. Same for `Box::new`, `Snapshot::deref`, and perhaps others.

creusot/src/ctx.rs
```diff
@@ -371,6 +378,32 @@ impl<'tcx> TranslationCtx<'tcx> {
             .as_ref()
     }
 
+    /// Same as `Self::term`, but does not normalize term.
+    /// This is used for logic aliases, since we do not want our alias closure to be
+    /// optimized out (e.g. builtin("identity") replaces the call to the logic function
+    /// by `self`, which breaks the aliasing machinery).
```
jhjourdan, line 386: I really don't like seeing several subtly different variants of the same function proliferating.

I would rather propose:
- Doing the aliasing substitution during normalization
- Substitute with an already normalized term

Note that it is OK to substitute a function call with something which is not a function call (like in the `identity` case).

creusot/src/ctx.rs
```diff
@@ -714,6 +761,46 @@ impl<'tcx> TranslationCtx<'tcx> {
...
+                let mut err = self
+                    .dcx()
+                    .struct_span_err(sid, "Implicit logic alias in trait impl is not supported");
+                err.span_note(sp, "Trait item has a `#[logic_alias]` attribute");
+                err.span_suggestion_hidden(
+                    sid,
+                    "Trait item and Impl item must have the same `#[logic_alias]` attribute",
+                    "",
+                    rustc_errors::Applicability::MachineApplicable,
+                );
```
jhjourdan, line 800: While I do agree that if an trait item has an alias, then the impls should have the same alias, I think the converse does not apply. Trait impls should be allowed to have aliases that the trait does not have. The typical example is `deref`, which is a program method without alias by default, but in some impls we do have an alias (ex: Snapshot, Box, Ghost, ...).

tests/should_fail/logic_alias/alias_defined_in_different_impls.stderr
```diff
@@ -0,0 +1,20 @@
+error: Cannot use #[logic_alias] in this context, because logic function is defined in trait T1, and program function is defined for trait T2.
```
jhjourdan, line 1: What's the issue here? As long as the impl context implies that the instance exists, we are good.

tests/should_fail/logic_alias/alias_defined_in_different_impls2.stderr
```diff
@@ -0,0 +1,20 @@
+error: Cannot use #[logic_alias] in this context, because program function is bound to type S2 and logic function is bound to type S1
```
jhjourdan, line 1: Again, what's the issue here? I agree it is somewhat bad practice to do that, but I don't see a good reason to systematically reject that.

tests/should_fail/logic_alias/alias_defined_in_different_trait.stderr
```diff
@@ -0,0 +1,20 @@
+error: A trait method can only be aliased by a logic method from the same trait. (program method defined in trait T2, logic alias defined in trait T1)
```
jhjourdan, line 1: Same here. I don't see why we would reject that, since the instance always exists.

Actually, it seems to be possible that some trait give a program implementation to something that only exists logically in a supertrait.

tests/should_fail/logic_alias/alias_to_extern_crate_item.stderr
```diff
@@ -0,0 +1,20 @@
+error: The logic function must be defined in the same crate as the aliased program function.
```
jhjourdan, line 1: Any good justification for this? I can imagine situations where we want to put the logical definition somewhere, and the implementation somewhere else.

tests/should_fail/logic_alias/function_aliased_to_trait.stderr
```diff
@@ -0,0 +1,20 @@
+error: A program function can only be aliased by another program function.
```
jhjourdan, line 1: Same here: I don't see a good reason to reject this.

tests/should_fail/logic_alias/missing_trait_impl.stderr
```diff
@@ -0,0 +1,20 @@
+error: Cannot use #[logic_alias] in this context, because type S2 does not implement trait IsZeroLog.
```
jhjourdan, line 1: Same here. No reason to forbid this.

tests/should_fail/logic_alias/missing_trait_impl2.stderr
```diff
@@ -0,0 +1,20 @@
+error: Cannot #[logic_alias] in this context, because logic function is bound to type S2 and program function is bound to type S1
```
jhjourdan, line 1: Same.

tests/should_fail/logic_alias/not_a_funcall.stderr
```diff
@@ -0,0 +1,8 @@
+error: `logic_alias` should contain a path to a logic function with the same signature, or a logic function call with same return type
```
jhjourdan, line 1: This is confusing. `.into()` is a logic function call with the same return type, isn't it?

tests/should_fail/logic_alias/prog_and_alias_in_different_traits.stderr
```diff
@@ -0,0 +1,20 @@
+error: A trait method can only be aliased by a logic method from the same trait. (program method defined in trait T2, logic alias defined in trait T1)
```
jhjourdan, line 1: Same here. I don't see why we should reject that.

tests/should_fail/logic_alias/trait_alias_to_non_trait.stderr
```diff
@@ -0,0 +1,20 @@
+error: Cannot use #[logic_alias] in this context, because type S does not implement trait T.
```
jhjourdan, line 1: Same here. I don't see why we should reject this.

creusot/src/logic_alias.rs
```diff
@@ -0,0 +1,537 @@
...
+//! fn add_one_log(x: i64) -> i64 { ... }
+//!
+//! #[ensures(result == add_one_log(x))]
+//! fn add_one(x: i64) -> i64 { ... }
+//!
+//! #[ensures(result == add_one_log(add_one_log(a)))]
+//! fn add_two(a: i64) -> bool { ... }
+//! ```
+//!
+//! TODO: move this to user documentation
```
jhjourdan, line 46: Indeed.

creusot/src/logic_alias.rs
```diff
@@ -0,0 +1,537 @@
...
+//! ## Step 3: substitution
+//! See [creusot::translation::pearlite::from_thir::expr_term]
+//! and [creusot::logic_alias::subst_call]
+//!
+//! # Validity of aliasing
+//!
+//! The following table shows the rules for aliasing:
+//! (NB: This is still a work in progress, and I do not yet guarantee that `check_validity`
+//!  enforces these rules properly, since they have been subject to quite some change
+//!  recently following discussions with Diane and Li-Yao. I rather wait for a consensus
```
jhjourdan, line 93: Ah, ok, then I need to enter the discussion as well.

jhjourdan (commented):
All in all, this is good work!

Of course, I have plenty of comments, see in the review.

A few general remarks:
- How does all this play with termination check? Could we have test checking that recursing through an alias does not lead to unsoundness?
- I would really love to use this to sole the issue with deref/deref_mut/new of Snapshot/Ghost/Box.
- I don't know what to thing about `#[hybrid]`. First, it would need two variants depending on whether we are in an `impl` block or not. Second, it has a potential to be very confusing, since the body of the function needs to be syntactically interpreted in two quite different languages at the same time (Pearlite and Rust).

PR #2220, opened 2026-08-04
Add `extern_spec` for `NonZero`

### LINKED ISSUE #2217: Add `extern_spec` for `NonZero<T>`
`NonZero` types are used in programs that want to also leverage the type system to prove correctness and to get layout optimizations. Currently `View` and `DeepModel` are not implemented for the type.

~~This appeared easy enough to add myself but I've hit a weird error:~~

```
 no method named `__creusot_view_stub` found for type parameter `T` in the current scope
```

~~It's unclear why this happens since `ZeroablePrimitive` is refined to require `View` and `DeepModel` so these bounds should be implied but it still behaves as if they weren't. That method is implemented for all types implementing `View`, so should be available.~~

~~The code I tried: https://github.com/Kixunil/creusot/pull/1/changes#diff-395074bc7336c8f926175cd143cf9f44ca74b57af5bf866708883b2f578be005 (Don't bother reading PR comments and other files. I used clanker to help me write the code but I did steer it heavily and reviewed the code myself. I'm stuck here too as a human.)~~

Figured it out.

Kixunil:
`NonZero` is a standard library type that is used for layout optimizations as well as using the type system to prove basic invariants. So far creusot didn't support it.

This adds basic support for the type, so far only adding `View`, `DeepModel` and `Invariant` impls as well as constructors and the `get` method.

Closes #2217


creusot-std/src/std/num.rs
```diff
@@ -603,3 +603,60 @@ spec_abs_diff!(u32, i32);
...
+use core::num::{NonZero, ZeroablePrimitive};
+
+#[cfg(creusot)]
+extern_spec! {
+    mod core {
+        mod num {
+            trait ZeroablePrimitive: View<ViewTy = Int> + DeepModel<DeepModelTy = Int> {}
+        }
+    }
+}
```
jhjourdan: This is a no-op. Extern specs cannot be used to add a super trait to a trait. Could you please remove this?

creusot-std/src/std/num.rs
```diff
@@ -603,3 +603,60 @@ spec_abs_diff!(u32, i32);
...
+
+#[cfg(creusot)]
+impl<T: ZeroablePrimitive + View<ViewTy = Int> + DeepModel<DeepModelTy = Int>> View for NonZero<T> {
```
jhjourdan:
```suggestion
impl<T: ZeroablePrimitive> View for NonZero<T> {
```

ZeroablePrimitive types are a fixed bounded set of types which are already all Int view and deep model. Could you please remove these constraints everywhere they are unnecessary?

creusot-std/src/std/num.rs
```diff
@@ -603,3 +603,60 @@ spec_abs_diff!(u32, i32);
...
+impl<T: ZeroablePrimitive + View<ViewTy = Int> + DeepModel<DeepModelTy = Int>> Invariant for NonZero<T> {
+    #[logic(open)]
+    fn invariant(self) -> bool {
+        pearlite! { self@ != 0 }
+    }
+}
+
+#[cfg(creusot)]
+extern_spec! {
+    impl<T: ZeroablePrimitive + View<ViewTy = Int> + DeepModel<DeepModelTy = Int>> NonZero<T> {
```
jhjourdan:
```suggestion
    impl<T: ZeroablePrimitive + View<ViewTy = Int>> NonZero<T> {
```

jhjourdan (dismissed):
Thanks for the contribution! Could you please address the comments before we merge?

jhjourdan:
Also you need to fix formatting. Please run `./fmt`.

creusot-std/src/std/num.rs
Kixunil: Oh, sure, I thought it might be unneeded and forgot.

jhjourdan: Then could you remove this?

jhjourdan:
I thought a bit more about this, and I think it should be useful (e.g., for bitwise operations) to have an opaque logical function from `NonZero<T>` to `T` and then to specify all the other functions (view, deep model, program functions) in terms of that. Could you do that? Thanks!

Kixunil:
Is this what you had in mind? I also remembered that the code could be `check(ghost)` but it seems there's some problem with it that I can't understand. I'm still a beginner when it comes to creusot/formal verification in general.

creusot-std/src/std/num.rs
```diff
@@ -603,3 +603,84 @@ spec_abs_diff!(u32, i32);
...
+#[cfg(creusot)]
+impl<T: ZeroablePrimitive> Plain for NonZero<T> {
+    #[trusted]
+    #[ensures(*result == *snap)]
+    #[check(ghost)]
+    #[allow(unused_variables)]
+    fn into_ghost(snap: Snapshot<Self>) -> Ghost<Self> {
+        Ghost::conjure()
+    }
+}
```
jhjourdan, line 637: Do you really need that?

I am not very confortable with this instance, because it implies that there is no null value of `NonZero` *at the logical level*, which in turns implies that we cannot model `NonZero<T>` with `T` directly (if we do that, then we exclude 0 at the program/ghost level, but not at the logical level). Modeling `NonZero<T>` with `T` directly is a simple way of specifying that `get_logic` is injective (which is sometimes useful but tedious in general).

Kixunil: Ah, so users should just call `get()` in ghost code and use the resulting value for proofs? The reason I added this was that I came across the trait, saw it implemented on a bunch of primitives with the rationale that they are not pointers so since this isn't a pointer either I though it could be added for completeness. But I don't *personally* need this.

Also note that `char` has a niche and also impls `Plain` so maybe that one is also wrong?

creusot-std/src/std/num.rs
Kixunil: Done in the newest push, I forgot to commit it. :see_no_evil:

creusot-std/src/std/num.rs
jhjourdan: The problem is not with having a niche or not. The point is that to implement `Plain`, the logical meaning of the type should be the same as the meaning in programs. It's not the case for pointer types because they carry ownership in programs. It's neither the case here because of the niche, which is not enforced in the logic because type invariants are not enforced in the logic.

That being said, I think we should consider adding the type invariant as a precondition of `into_ghost`, but this is another story...

jhjourdan: Anyway, please remove this instance and I'll merge.

jhjourdan:
I fixed the soundness of `into_ghost`, and made `NonZero<T>` identical to `T` at the logical level. Merging.

Kixunil:
Thanks! I was busy until recently and I was just planning to look into this in upcoming days.

Issue #2223, opened 2026-08-11
Support using variant constructors as function values (e.g., `.map(E::A)`)


greeng3:
Using an enum tuple-variant constructor as a function value ICEs the driver.

Minimal repro (creusot-std 0.12.0, current cargo-creusot, rustc 1.97.0-nightly 66da6cae1):

```rust
use creusot_std::prelude::*;

pub enum E {
    A(u32),
}

pub fn f() -> Vec<E> {
    (0..3).map(E::A).collect()
}
```

```console
error: internal compiler error: .../rustc_middle/src/hir/mod.rs:499:13: fn_arg_idents:
unexpected item DefId(... ~ repro[...]::E::A::{constructor#0})
```

Rewriting the constructor-as-value to a closure — `.map(|x| E::A(x))` — avoids the ICE (verified on the real crate this was found in), so the crash is specific to the synthesized `{constructor#0}` item reaching `fn_arg_idents`.

Found while measuring how far `#[trusted]` annotation gets a ~40-file crate toward translation; the constructor was `parser.map(Expr::Atom)` in an ordinary recursive-descent parser.

If useful I can also file the second ICE we hit immediately after: a Why3-backend panic (`backend::clone_map::elaborator::Expander::expand`, query stack `[mir_for_ctfe] caching mir of std::env::consts::OS`) when a function references `std::env::consts::OS` — not yet reduced.

greeng3:
Update: **this no longer ICEs on `master`** — reporting so the issue can be closed or repurposed rather than left standing as a live ICE.

I re-tested with a smaller repro than the one I filed (no `Vec`/iterator needed):

```rust
extern crate creusot_std;

pub enum E {
    A(u32),
}

pub fn g(o: Option<u32>) -> Option<E> {
    o.map(E::A)
}
```

**0.12.0** (what I originally filed against) — ICE, as reported:

```text
error: internal compiler error: rustc_middle/src/hir/mod.rs:499:13: fn_arg_idents:
unexpected item DefId(0:6 ~ ctorprobe[a00c]::E::A::{constructor#0})
```

**`master` @ 7c3d1e83** — a clean diagnostic instead:

```text
error: expanding "E::A" failed because of unexpected kind Unsupported(Ctor(Variant, Fn)) while translating "g"
  --> constructor_as_value.rs:12:1
```

A tuple-*struct* constructor gives the analogous `Unsupported(Ctor(Struct, Fn))`, and the fn-pointer
[...]

PR #2224, opened 2026-08-11
Add regression test for a constructor used as a function value

greeng3:
This is the regression test I offered in
[#2223](https://github.com/creusot-rs/creusot/issues/2223#issuecomment-5249418140). Test-only — no
translation changes.

I filed #2223 as an ICE against 0.12.0: using a tuple-variant constructor as a *function value*
rather than applying it directly reached rustc's `fn_arg_idents`, which has no entry for the
synthesized `{constructor#0}` item, so it hit the `span_bug!`:

```text
error: internal compiler error: rustc_middle/src/hir/mod.rs:499:13: fn_arg_idents:
unexpected item DefId(0:6 ~ ctorprobe[a00c]::E::A::{constructor#0})
```

On `master` it is already refused with a proper diagnostic, so there is nothing to fix — but nothing
pins the behaviour either, and the crashing path is one `item_type` arm away. This adds a
`should_fail` test recording the refusal:

```text
error: expanding "E::A" failed because of unexpected kind Unsupported(Ctor(Variant, Fn)) while translating "g"
```

Checked on `master` @ 7c3d1e83: `./t ui` 472/472, `./t fmt --fmt-check` clean.

Close this without ceremony if you'd rather not carry the extra test — I won't be offended, and #2223
can just be closed as fixed either way.

### DIFF
--- a/tests/should_fail/constructor_as_value.rs
+++ b/tests/should_fail/constructor_as_value.rs
@@ -1,0 +1,14 @@
+extern crate creusot_std;
+
+pub enum E {
+    A(u32),
+}
+
+// A tuple-variant constructor used as a function value, rather than applied
+// directly, reaches Creusot as an item of kind `Ctor`, which it does not
+// translate. Creusot must reject it with a diagnostic: asking rustc for the
+// argument idents of the synthesized `{constructor#0}` item instead ICEs,
+// as it did in 0.12.0 (creusot-rs/creusot#2223).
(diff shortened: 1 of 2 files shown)


Lysxia:
Indeed, for features that are relatively easy to implement I'd rather see them done than add a "missing feature" test.

greeng3:
Fair enough.

PR #2225, opened 2026-08-11
Support aarch64 Linux in creusot-install

greeng3:
On aarch64 Linux, `creusot-install` does not compile — `URLS` is `#[cfg]`-gated to linux/x86_64 and
macos/aarch64 only, so there is no definition for the platform:

```text
error[E0425]: cannot find value `URLS` in this scope
```

This adds the missing linux/aarch64 table. Z3 and cvc5 both publish aarch64 Linux binaries, so those
are wired up directly. Alt-Ergo and CVC4 do not, so the per-tool URLs become `Option<Url>` and
`install_provers` skips a `None` tool with a message pointing at `--external <TOOL>`:

```text
Skipping Alt-Ergo: no binary release for this platform. Install it separately and re-run with `--external altergo`.
```

Alt-Ergo is an OCaml program, so `opam install alt-ergo` + `--external alt-ergo` covers it. CVC4 is
archived upstream (last release 2020) and never shipped an aarch64 binary; cvc5 supersedes it.

The x86_64 Linux and aarch64 macOS tables are unchanged apart from being wrapped in `Some`, and I've
deliberately left the `PROVERS` list and the `why3find.json` template alone — see the caveat below.

### Checks

- `cargo build -p creusot-install` on aarch64 Linux: succeeds (it did not compile before this).
- `cargo check -p creusot-setup --target x86_64-unknown-linux-gnu`: clean.
- `cargo check -p creusot-setup --target aarch64-apple-darwin`: clean.
- `./t fmt --fmt-check`: clean.
- Both new SHA256s verified by fresh download of the release archives, not copied from anywhere.
- This has been running in anger: I've built and used Creusot on aarch64 Linux with this change for
  about three weeks (`creusot-install why3` / `creusot-rustc` / `prelude cargo-creusot provers
  why3-conf`, with `--external alt-ergo`), and Z3 + cvc5 install and discharge goals normally.

### One caveat, deliberately not fixed here

`CVC4` stays in `PROVERS` and in the `why3find.json` template, so on a platform where it is skipped,
`why3find` can still try to *calibrate* it and report `can not calibrate prover cvc4` on a goal it
cannot discharge quickly — which reads as an error rather than a clean unproved result. Locally I
drop CVC4 from both, but that is a prover-policy change affecting every platform, and it seemed
wrong to smuggle it into a PR about aarch64 support. Happy to send it separately, or to scope the
prover list per-platform instead, whichever you'd prefer.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -172,6 +172,45 @@ jobs:
       with:
         path: .why3find
         key: ${{ runner.os }}-why3find-${{ env.TODAY }}-${{ hashFiles('.why3find/**') }}
+  install-aarch64:
+    runs-on: ubuntu-24.04-arm
+    steps:
+    - uses: actions/checkout@v4
+    - name: Cache Opam switch
+      uses: actions/cache@v4
+      with:
+        path: ~/.local/share/creusot/_opam
+        key: ${{ runner.os }}-${{ runner.arch }}-opam-${{ hashFiles('creusot-deps.opam') }}
+        restore-keys: ${{ runner.os }}-${{ runner.arch }}-opam-
+    - run: sudo apt update && sudo apt install -y opam libzmq3-dev
+    - name: setup opam
+      run: |
+        opam init --bare --disable-sandboxing --no-setup -y
+        opam --cli=2.1 var --global in-creusot-ci=true
+    # The switch must exist before Alt-Ergo can be installed into it, so the components are taken in
+    # two passes rather than one `./INSTALL`.
+    - name: install why3 (creates the opam switch)
+      run: cargo run --bin creusot-install -- why3
+    - name: install alt-ergo from opam
+      run: opam install -y alt-ergo.2.6.2 --switch "$HOME/.local/share/creusot"
+    - name: install the rest, with alt-ergo supplied externally
+      run: |
+        eval $(opam env --switch "$HOME/.local/share/creusot" --set-switch)
+        cargo run --bin creusot-install -- --external alt-ergo \
+          creusot-rustc prelude cargo-creusot provers why3-conf
+    # Stat the results rather than trusting a zero exit: the point of the job is that the binaries
+    # an aarch64 user needs actually land.
+    - name: the installed toolchain is really there
+      run: |
+        D="$HOME/.local/share/creusot"
+        test -x "$D/bin/z3"
+        test -x "$D/bin/cvc5"
+        test -x "$D/bin/alt-ergo"
+        test -x $(echo "$D"/toolchains/*/bin/creusot-rustc)
+        test -x "$HOME/.cargo/bin/cargo-creusot"
+    - name: Minimize Opam cache
+      run: opam clean --all-switches --download-cache --logs --repo-cache
+
   install:
     strategy:
       matrix:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -172,6 +172,58 @@ jobs:
       with:
         path: .why3find
         key: ${{ runner.os }}-why3find-${{ env.TODAY }}-${{ hashFiles('.why3find/**') }}
+  # Creusot did not build on aarch64 Linux at all until the linux/aarch64 `URLS` table existed, so
+  # nothing here would have caught a regression in it. This job is deliberately narrow: it proves
+  # the *installation* completes on aarch64 and leaves the translation suite to the jobs above.
+  #
+  # `ubuntu-24.04-arm` is a native arm64 hosted runner, free on public repos — no QEMU, so this
+  # costs about what the x86_64 install job costs rather than many hours.
+  #
+  # Alt-Ergo publishes no aarch64-linux binary, so `creusot-install` skips it and the prover has to
+  # come from opam instead; that is what `--external alt-ergo` is for, and installing it into
+  # Creusot's own switch is what makes it resolvable. CVC4 is likewise absent on this platform
+  # (archived upstream, x86_64 only) and is simply not installed — cvc5 supersedes it.
+  install-aarch64:
+    runs-on: ubuntu-24.04-arm
+    steps:
+    - uses: actions/checkout@v4
+    - name: Cache Opam switch
+      uses: actions/cache@v4
+      with:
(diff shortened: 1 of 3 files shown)


Lysxia (commented):
Thanks for this. I think it's fine to merge this. We do have some tests that are solved only by CVC4, but as long as you don't need to touch those, things should be okay.

Can you add a CI script to just test that the installation succeeds on aarch64? Github Actions supports it, right?

greeng3:
Added in 837bc6a. And yes — GitHub Actions supports it directly: `ubuntu-24.04-arm` is a native
arm64 hosted runner, free on public repositories. No QEMU, so it costs about what the x86_64 install
job costs rather than the many hours emulation would take.

I made it a separate `install-aarch64` job rather than adding the runner to the existing `install`
matrix, for two reasons:

- every Linux step in that job is gated on `matrix.os == 'ubuntu-latest'`, so a new matrix entry
  would silently skip `apt install opam` and then fail at `setup opam`;
- its cache keys are `${{ runner.os }}-…`, and `runner.os` is `Linux` on both x86_64 and arm64, so
  the two would share a cache and restore each other's incompatible opam switch. The new job puts
  `runner.arch` in the key.

Happy to fold it into the matrix instead if you'd rather — it just means touching the conditionals
and keys of a job that currently works.

The job is deliberately narrow: it proves the *installation* completes on aarch64 and leaves
translation to the existing jobs. Alt-Ergo has no aarch64-linux binary, so it comes from opam and is
handed over with `--external alt-ergo`; that needs the switch to exist first, which is why the
components are taken in two passes rather than one `./INSTALL`. It ends by stat-ing z3, cvc5,
alt-ergo, `creusot-rustc` and `cargo-creusot` rather than trusting a zero exit.

Verified before pushing here, on a real arm64 runner in my fork: `install-aarch64` passed in 8m01s,
with all 10 jobs in the run green.

On CVC4 — understood, and this PR leaves `PROVERS` and the `why3find.json` template alone precisely
so those tests keep their prover. Worth flagging one consequence I hit locally: on a platform where
CVC4 is skipped, `why3find` can still try to *calibrate* it and report `can not calibrate prover
cvc4` on a goal it cannot discharge quickly, which reads as an error rather than a clean unproved
result. Not touched here; say the word if you'd like it addressed separately, per-platform rather
than globally.

.github/workflows/rust.yml
```diff
@@ -172,6 +172,58 @@ jobs:
...
+  # nothing here would have caught a regression in it. This job is deliberately narrow: it proves
+  # the *installation* completes on aarch64 and leaves the translation suite to the jobs above.
+  #
+  # `ubuntu-24.04-arm` is a native arm64 hosted runner, free on public repos — no QEMU, so this
+  # costs about what the x86_64 install job costs rather than many hours.
+  #
+  # Alt-Ergo publishes no aarch64-linux binary, so `creusot-install` skips it and the prover has to
+  # come from opam instead; that is what `--external alt-ergo` is for, and installing it into
+  # Creusot's own switch is what makes it resolvable. CVC4 is likewise absent on this platform
+  # (archived upstream, x86_64 only) and is simply not installed — cvc5 supersedes it.
```
Lysxia:
```suggestion
```

greeng3:
Thanks!

> Can you add a CI script to just test that the installation succeeds on aarch64? Github Actions supports it, right?

It does, and that job is already in this PR — commit 837bc6ae, `install-aarch64` in `.github/workflows/rust.yml`. It is what the comment block you flagged was describing, so removing the comment is not removing the job.

It runs on `ubuntu-24.04-arm`, a **native** arm64 hosted runner (free on public repos), so there is no QEMU involved and it costs about what the x86_64 `install` job costs rather than many hours. It is deliberately narrow, exactly as you describe: it installs and then stats the binaries an aarch64 user actually needs (`z3`, `cvc5`, `alt-ergo`, `creusot-rustc`, `cargo-creusot`) rather than trusting a zero exit. The translation suite is left to the existing jobs.

Two details in it that are not obvious, in case they matter on review:

- The opam cache key includes `runner.arch`. `runner.os` is `Linux` on both x86_64 and arm64 runners, so without it the two jobs would share a cache and restore each other's incompatible switch.
- Alt-Ergo publishes no aarch64-linux binary, so it comes from opam instead and the install is taken in two passes (`why3` first to create the switch, then the rest with `--external alt-ergo`).

> We do have some tests that are solved only by CVC4, but as long as you don't need to touch those, things should be okay.

Those are safe. This PR does not change the prover list, and it does not remove CVC4 from any platform that has it — it only widens `cvc4: Url` to `Option<Url>` so a platform can say it has none. Every existing entry becomes `Some(...)` and keeps exactly what it had; `None` appears only in the new linux-aarch64 row, where CVC4 was never available in the first place (archived upstream in 2020, x86_64 only). So on the runners that solve those tests, nothing about CVC4 changes.

Accepted on the comment block — dropping it.

Lysxia (approved):

.github/workflows/rust.yml
```diff
@@ -172,6 +172,47 @@ jobs:
...
+  install-aarch64:
+    runs-on: ubuntu-24.04-arm
+    steps:
+    - uses: actions/checkout@v4
+    - name: Cache Opam switch
+      uses: actions/cache@v4
+      with:
+        path: ~/.local/share/creusot/_opam
+        # `runner.os` is `Linux` on both x86_64 and arm64 runners, so the arch has to be in the key
+        # or the two would share a cache and restore each other's incompatible switch.
```
jhjourdan: Please remove this comment. The code is self discribing.

greeng3:
Dropped in 90225f76 (on top of your merge, so nothing of yours is rewritten). Thanks for the merge
and the approval.

Two comments of mine are still in that job, both of the same species as the ones you and @Lysxia
have already had me remove — the `./INSTALL` two-pass note and the "stat the results" note above the
final check. I left them because neither restates a line of YAML, but say the word and they go too;
I would rather over-trim than keep sending you comment review.

Lysxia:
Looks okay to me. Can you squash the commits?

greeng3:
Squashed to a single commit (`77dd30a`), rebased onto current master in the process.

PR #2227, opened 2026-08-12
Support a constructor used as a function value

### LINKED ISSUE #2223: Support using variant constructors as function values (e.g., `.map(E::A)`)

greeng3:
Implements the feature request in #2223, taking the second of the two suggestions there — generating
the pre/post for the constructor directly, since they are very simple logical terms.

`o.map(E::A)` — a tuple-struct or tuple-variant constructor passed as a function value rather than
applied — was refused with `unexpected kind Unsupported(Ctor(Variant, Fn))`, and before 0.13.0 it
ICEd in `fn_arg_idents`.

## What was in the way

Three things, each fixed where every caller routes through rather than at the use site:

- **`item_type`** did not classify a constructor at all, so the elaborator took its `Unsupported`
  arm. A constructor behaves like a program function with a trivial contract, so it is classified
  as one.
- **`inputs_and_output`** asked rustc for the constructor's argument idents. A constructor is
  synthesized and has no HIR parameters, so that query raises `fn_arg_idents: unexpected item` —
  the original ICE. Its arguments are positional and therefore anonymous, which the existing
  unnamed-parameter path already handles.
- **`typing_env_with`** chose `post_borrowck_analysis`, because a constructor has shim MIR and so
  passes `is_mir_available`. That mode needs `opaque_types_defined_by`, which rustc refuses for a
  constructor. A constructor defines no opaque types, so the non-body mode is both correct and the
  only one available.

## Why that alone wasn't enough

With just the above it translates, but with a `true` postcondition:

```
predicate postcondition_A (self: ()) (args: UInt32.t) (result: t_E) = let _1 = args in true
```

It would compile and prove nothing about the result, which seemed worse than refusing it — the
vacuity is silent. So `contract_of` now gives a constructor the postcondition it obviously has:

```
predicate postcondition_A (self: ()) (args: UInt32.t) (result: t_E) = let _1 = args in result = A _1
```

## Test

Written so it cannot pass unless that postcondition is really there:

```rust
#[ensures(forall<x: u32> o == Some(x) ==> result == Some(E::A(x)))]
#[ensures(o == None ==> result == None)]
pub fn g(o: Option<u32>) -> Option<E> {
    o.map(E::A)
}
```

- `./t ui` — 473/473. Worth noting: the UI tests compare exact generated Coma, and no other
  baseline moved, so nothing outside constructors-as-values changed translation.
- `./t why3` — `tests/should_succeed/constructor_as_value.coma ... proved`.
- `./t fmt --fmt-check` — clean.
- Also run on x86_64 in my fork before sending: the full workflow is green, including `why3`
  proving the new test. One earlier attempt on the same commit failed `why3` on
  `examples/insertion_sort.coma` and `examples/parallel_add/sc_generic.coma`; re-running the same
  commit passed. Different examples fail on different runs there, including on an unmodified
  `master`, so I read it as prover timeouts on those runners rather than anything here — but you
  will see it against your own CI, so flagging it rather than leaving you to wonder.

## One edge I did not fix

`contract_of` pushes `requires false` for items from a non-Creusot crate, so a constructor imported
from another crate should still be poisoned. I have not tested that, it is not covered by the new
test, and I did not want to claim behaviour I have not measured. Happy to extend this to cover it if
you would like it in the same PR.

Closes #2223.

### DIFF
--- a/creusot/src/backend/ty_inv.rs
+++ b/creusot/src/backend/ty_inv.rs
@@ -7,7 +7,7 @@ use crate::{
     resolution::TraitResolved,
     translation::{
         pearlite::{Ident, Pattern, Term, TermKind, Trigger},
-        specification::{Condition, PreSignature},
+        specification::{Condition, PreSignature, ctor_term},
     },
 };
 use rustc_abi::{FieldIdx, VariantIdx};
@@ -338,6 +338,16 @@ pub(crate) fn sig_add_type_invariant_spec<'tcx>(
         }
     });
     pre_sig.contract.requires.splice(0..0, new_requires);
+
+    // A constructor has no body, so the result invariant below is a goal nothing else can
+    // discharge: the field invariants above do not imply the built value's. Require it, which
+    // moves the obligation to the call site, where the fields are known.
+    if let Some(built) = ctor_term(ctx, def_id, &pre_sig.inputs, pre_sig.output)
+        && let Some(term) = inv_call(ctx, typing_env, scope_id, built)
+    {
+        let expl = format!("expl:{} result type invariant", fn_name);
+        pre_sig.contract.requires.push(Condition { term: term.span(ctx.def_span(def_id)), expl });
+    }
 
     let ret_ty_span: Option<Span> = try { ctx.hir_get_fn_output(def_id.as_local()?)?.span() };
     if (ctx.def_kind(def_id) == DefKind::ConstParam || !is_open_inv_result(ctx.tcx, def_id))
--- a/creusot/src/backend/ty_inv.rs
+++ b/creusot/src/backend/ty_inv.rs
@@ -7,7 +7,7 @@ use crate::{
     resolution::TraitResolved,
     translation::{
         pearlite::{Ident, Pattern, Term, TermKind, Trigger},
-        specification::{Condition, PreSignature, ctor_term},
+        specification::{Condition, PreSignature},
     },
 };
 use rustc_abi::{FieldIdx, VariantIdx};
@@ -338,16 +338,6 @@ pub(crate) fn sig_add_type_invariant_spec<'tcx>(
         }
     });
     pre_sig.contract.requires.splice(0..0, new_requires);
-
-    // A constructor has no body, so the result invariant below is a goal nothing else can
-    // discharge: the field invariants above do not imply the built value's. Require it, which
-    // moves the obligation to the call site, where the fields are known.
-    if let Some(built) = ctor_term(ctx, def_id, &pre_sig.inputs, pre_sig.output)
-        && let Some(term) = inv_call(ctx, typing_env, scope_id, built)
-    {
-        let expl = format!("expl:{} result type invariant", fn_name);
-        pre_sig.contract.requires.push(Condition { term: term.span(ctx.def_span(def_id)), expl });
-    }
 
     let ret_ty_span: Option<Span> = try { ctx.hir_get_fn_output(def_id.as_local()?)?.span() };
     if (ctx.def_kind(def_id) == DefKind::ConstParam || !is_open_inv_result(ctx.tcx, def_id))
--- a/creusot/src/ctx.rs
+++ b/creusot/src/ctx.rs
@@ -569,7 +569,8 @@ impl<'tcx> TranslationCtx<'tcx> {
         param_env: ParamEnv<'tcx>,
     ) -> TypingEnv<'tcx> {
         // FIXME: is it correct to pretend we are doing a non-body analysis?
-        // A constructor has shim MIR, so it passes `is_mir_available`, but it is not a body owner:
+        // `contract_of` calls this for a constructor used as a function value (`o.map(E::A)`). A
+        // constructor has shim MIR, so it passes `is_mir_available`, but it is not a body owner:
         // `post_borrowck_analysis` needs `opaque_types_defined_by`, which rustc refuses for one.
         let mode = if self.is_mir_available(def_id)
             && def_id.is_local()
--- a/creusot/src/ctx.rs
+++ b/creusot/src/ctx.rs
@@ -35,7 +35,7 @@ use rustc_data_structures::steal::Steal;
 use rustc_errors::{Diag, DiagMessage, FatalAbort};
 use rustc_hir::{
     HirId,
-    def::DefKind,
+    def::{CtorKind, DefKind},
     def_id::{CrateNum, DefId, LOCAL_CRATE, LocalDefId, ModId},
 };
 use rustc_index::bit_set::DenseBitSet;
@@ -569,7 +569,14 @@ impl<'tcx> TranslationCtx<'tcx> {
         param_env: ParamEnv<'tcx>,
     ) -> TypingEnv<'tcx> {
         // FIXME: is it correct to pretend we are doing a non-body analysis?
-        let mode = if self.is_mir_available(def_id) && def_id.is_local() {
+        // A tuple-struct/tuple-variant constructor has shim MIR, so it passes `is_mir_available`,
+        // but it is not a body owner: rustc refuses to compute `opaque_types_defined_by` for one,
+        // which `post_borrowck_analysis` needs. It defines no opaque types, so the non-body mode is
+        // both correct and the only one available (#2223).
+        let mode = if self.is_mir_available(def_id)
+            && def_id.is_local()
+            && !matches!(self.def_kind(def_id), DefKind::Ctor(..))
+        {
             TypingMode::post_borrowck_analysis(self.tcx, def_id.as_local().unwrap())
         } else {
             TypingMode::non_body_analysis()
@@ -761,6 +768,10 @@ impl<'tcx> TranslationCtx<'tcx> {
             DefKind::AssocTy => ItemType::AssocTy,
             DefKind::Field => ItemType::Field,
             DefKind::Variant => ItemType::Variant,
+            // A tuple-struct/tuple-variant constructor used as a *function value* (`o.map(E::A)`)
+            // reaches us as an item of its own. It behaves like a program function whose contract
+            // is trivial, so it is translated as one (#2223).
+            DefKind::Ctor(_, CtorKind::Fn) => ItemType::Program,
             dk => ItemType::Unsupported(dk),
         }
     }
--- a/creusot/src/ctx.rs
+++ b/creusot/src/ctx.rs
@@ -569,10 +569,8 @@ impl<'tcx> TranslationCtx<'tcx> {
         param_env: ParamEnv<'tcx>,
     ) -> TypingEnv<'tcx> {
         // FIXME: is it correct to pretend we are doing a non-body analysis?
-        // A tuple-struct/tuple-variant constructor has shim MIR, so it passes `is_mir_available`,
-        // but it is not a body owner: rustc refuses to compute `opaque_types_defined_by` for one,
-        // which `post_borrowck_analysis` needs. It defines no opaque types, so the non-body mode is
-        // both correct and the only one available (#2223).
+        // A constructor has shim MIR, so it passes `is_mir_available`, but it is not a body owner:
+        // `post_borrowck_analysis` needs `opaque_types_defined_by`, which rustc refuses for one.
         let mode = if self.is_mir_available(def_id)
             && def_id.is_local()
             && !matches!(self.def_kind(def_id), DefKind::Ctor(..))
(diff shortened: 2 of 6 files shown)


creusot/src/translation/specification.rs
```diff
@@ -368,6 +372,32 @@ pub(crate) fn contract_of<'tcx>(ctx: &TranslationCtx<'tcx>, def_id: DefId) -> Pr
...
+            let fields =
+                inputs.iter().map(|(ident, _, ty)| Term::var(ident.0, *ty)).collect::<Box<[_]>>();
+            let built = Term {
+                ty: output,
+                kind: TermKind::Constructor { variant, fields },
+                span: ctx.def_span(def_id),
+            };
+            let term = Term::var(name::result(), output).eq(ctx.tcx, built);
+            contract
+                .ensures
```
xldenis, line 401: I think we need to require the type invariant holds no?

xldenis (commented):
This hits the right places. Could you please tidy the comments? They are a bit claudish still.

creusot/src/translation/specification.rs
```diff
@@ -566,10 +596,18 @@ pub fn inputs_and_output<'tcx>(
...
-                .cloned()
+            // A tuple-struct/tuple-variant constructor is a synthesized item with no HIR
+            // parameters, so asking rustc for its argument idents would ICE (#2223). Its arguments
+            // are anonymous by construction — the fields are positional — so name them like any
+            // other unnamed parameter.
+            let arg_idents: Vec<Option<rustc_span::Ident>> =
+                if matches!(tcx.def_kind(def_id), DefKind::Ctor(..)) {
+                    vec![None; sig.inputs().len()]
+                } else {
+                    tcx.fn_arg_idents(def_id).to_vec()
```
Lysxia:
```suggestion
                    &vec![None; sig.inputs().len()]
                } else {
                    tcx.fn_arg_idents(def_id)
```

tests/should_succeed/constructor_as_value.rs
```diff
@@ -0,0 +1,14 @@
...
+    A(u32),
+}
+
+/// A tuple-variant constructor used as a function value. The postcondition is only provable if the
+/// constructor carries its meaning — `result = A x` — rather than an opaque one.
+#[ensures(forall<x: u32> o == Some(x) ==> result == Some(E::A(x)))]
+#[ensures(o == None ==> result == None)]
+pub fn g(o: Option<u32>) -> Option<E> {
+    o.map(E::A)
+}
```
Lysxia, line 14: Some more tests (an external type, and different arities):

```rust
extern crate creusot_std;
use creusot_std::prelude::*;

#[ensures(result == o.map_logic(|x| Some(x)))]
pub fn g<A>(o: Option<A>) -> Option<Option<A>> {
    o.map(Some)
}

pub struct T0();

pub fn h() -> T0 {
    let t = T0;
    t()
}

#[allow(dead_code)]
pub struct T2(u32, u32);

#[ensures(result == T2(0u32, 1u32))]
pub fn i() -> T2 {
    let t = T2;
    t(0, 1)
}
```

Lysxia:
> contract_of pushes requires false for items from a non-Creusot crate, so a constructor imported
from another crate should still be poisoned. I have not tested that, it is not covered by the new
test, and I did not want to claim behaviour I have not measured. Happy to extend this to cover it if
you would like it in the same PR.

Yes, please support this case.

jhjourdan:
Thanks for the contribution. I need to review this absolutely. Could you please wait before I come back from vacation?

One thing that needs to be taken care of is type invariants: one need to make sure that the precondition ensures it.

xldenis:
> One thing that needs to be taken care of is type invariants: one need to make sure that the precondition ensures it

https://github.com/creusot-rs/creusot/pull/2227#discussion_r3764494833

creusot/src/ctx.rs
jhjourdan, line 566: Another comment which I think would be more interesting is to document when we are actually calling this function for constructors.

greeng3:
Thanks, all four points are addressed in 60f2d86d. @jhjourdan, no rush from my side; this is here for whenever you are back.

### Type invariants

You were both right, and it was worse than the edge case I had flagged. I measured it before changing anything:

```rust
pub struct Pos(u32);
impl Invariant for Pos {
    #[logic] fn invariant(self) -> bool { pearlite! { self.0@ > 0 } }
}
pub fn f(o: Option<u32>) -> Option<Pos> { o.map(Pos) }
```

The constructor promised the result invariant with nothing required to establish it:

```
let f_Pos (_1: UInt32.t) (return (x: t_Pos)) = any
  [ return (result: t_Pos) ->
  {... [@expl:Pos result type invariant] inv_Pos result) /\ (... result = { f0 = _1 })} ...

predicate precondition_Pos (self: ()) (args: UInt32.t) = let _1 = args in true
```

It is not unsound, which was my first guess and was wrong: the goal is emitted and simply fails, `Goal vc_f_Pos: ✘`. But it means every constructor of a type with a non-trivial invariant is unprovable, and nothing in the suite noticed because no existing test constructs one that way.

The invariant of the built value does not follow from the invariants of the fields, so requiring the fields' invariants is not enough. `sig_add_type_invariant_spec` now requires the built value's, which is exactly "the precondition ensures it":

```
let f_Pos (_1: UInt32.t) ... = {[@expl:Pos result type invariant] inv_Pos { f0 = _1 }}
predicate precondition_Pos (self: ()) (args: UInt32.t) = let _1 = args in inv_Pos { f0 = _1 }
```

The obligation lands at the call site, where the fields are known. It is in `sig_add_type_invariant_spec` rather than `contract_of` because that is where `inv_call` has the typing env and scope, and it means the rule applies to constructors from any crate. The term for the built value is now shared between the postcondition and the required invariant, so the two cannot drift apart.

### External constructors

Done. A constructor was getting `requires false` from the `non_creusot_crate` branch, which made `o.map(Some)` unusable. A constructor needs no written spec, so it is now exempt from that branch and its contract is identical whichever crate it comes from. `Some` now carries the field invariant, the built-value invariant, and `result = Some _1`, with no `requires false`.

### Tests and comments

Added the three you asked for: the external generic type, zero arity, and two arguments. All four modules prove.

Comments: took the point, and cut them back. Also applied the `fn_arg_idents` suggestion, and documented that `item_type` only sees a constructor when it is used as a value, since applying one is an aggregate in MIR.

### Evidence

`./t ui` passes in full. Since it compares exact Coma, that also shows the `sig_add_type_invariant_spec` change moved no other baseline, which was my main worry given every signature goes through it. `./t why3 constructor_as_value` proves all four modules (10 goals, alt-ergo). `rustfmt --check` clean on the touched files.

Note I have only run this on aarch64. I have an x86_64 harness on my fork and can run it if you would like that before you look.

creusot/src/backend/ty_inv.rs
```diff
@@ -339,6 +339,16 @@ pub(crate) fn sig_add_type_invariant_spec<'tcx>(
     });
     pre_sig.contract.requires.splice(0..0, new_requires);
 
+    // A constructor has no body, so the result invariant below is a goal nothing else can
+    // discharge: the field invariants above do not imply the built value's. Require it, which
+    // moves the obligation to the call site, where the fields are known.
```
jhjourdan: Remove this comment. The code is self discribing and the comment is rather obscure.

creusot/src/backend/ty_inv.rs
```diff
@@ -339,6 +339,16 @@ pub(crate) fn sig_add_type_invariant_spec<'tcx>(
...
+    if let Some(built) = ctor_term(ctx, def_id, &pre_sig.inputs, pre_sig.output)
+        && let Some(term) = inv_call(ctx, typing_env, scope_id, built)
+    {
+        let expl = format!("expl:{} result type invariant", fn_name);
+        pre_sig.contract.requires.push(Condition { term: term.span(ctx.def_span(def_id)), expl });
+    }
+
```
jhjourdan: This function is not the right place for this code. The point of this code is to add implicit type invariant specifications that appear for every functions. Function constructors should not be a special case.

Please move this code next to the code that synthesizes the postconditon.

creusot/src/ctx.rs
```diff
@@ -761,6 +766,11 @@ impl<'tcx> TranslationCtx<'tcx> {
             DefKind::AssocTy => ItemType::AssocTy,
             DefKind::Field => ItemType::Field,
             DefKind::Variant => ItemType::Variant,
+            // Reached only for a constructor used as a function value (`o.map(E::A)`): applying
```
jhjourdan, line 770: Please in addition document which caller(s) of typing_env_with may call it with a function constructor.

creusot/src/translation/specification.rs
```diff
@@ -359,7 +391,10 @@ pub(crate) fn contract_of<'tcx>(ctx: &TranslationCtx<'tcx>, def_id: DefId) -> Pr
             contract,
         }
     } else {
-        if ctx.non_creusot_crate(def_id.krate) {
+        let ctor = ctor_term(ctx, def_id, &inputs, output);
```
jhjourdan: If you move the synthesis of the pre here, you don't need the `ctor_term` function.

creusot/src/translation/specification.rs
```diff
@@ -566,8 +610,15 @@ pub fn inputs_and_output<'tcx>(
                 typing_env,
                 tcx.fn_sig(def_id).instantiate_identity().skip_normalization(),
             );
-            let inputs = tcx
-                .fn_arg_idents(def_id)
+            // A constructor has no HIR parameters, so `fn_arg_idents` raises `unexpected item` for
+            // one. Its fields are positional, so they are unnamed like any other anonymous
+            // parameter.
```
jhjourdan: The code is self discribing. Please remove this comment.

creusot/src/ctx.rs
jhjourdan: Please address this.

greeng3:
Thanks — all addressed in 2aea1fcb.

- **`ty_inv.rs` is back to what it was on master.** You're right that it is the wrong place: it adds
  the invariants every function gets implicitly, and a constructor is not a special case of that.
  The result invariant is now required in `contract_of`, directly above the postcondition.
- **`ctor_term` is gone**, inlined at that one site as you predicted, so the precondition and the
  postcondition are built from the same term without a helper to keep them in agreement.
- **`typing_env_with`** now names the caller that reaches it with a constructor. Measured rather
  than guessed — I instrumented it with a backtrace, and the only two chains that get there are
  `Expander::expand_program -> sig -> pre_sig_of -> contract_of` and
  `PurityVisitor::purity -> sig -> pre_sig_of -> contract_of`, so `contract_of` is the honest
  answer for both.
- The two comments you called self-describing are deleted.

One thing worth recording, since the move is not obviously behaviour-preserving: in
`sig_add_type_invariant_spec` the invariant was built with the *use-site* typing env, and in
`contract_of` it is built with the constructor's own. I expected that to change which `inv` calls
survive `is_tyinv_trivial`. It does not, at least not anywhere the suite reaches: `./t ui` passes
with **no baseline moved at all**, so the generated Coma is byte-identical to the previous commit.

Evidence: `./t ui` all 473 pass, `./t why3 constructor_as_value` proved,
`rustfmt --check --edition 2024` clean. (Local runs are aarch64; happy to point the fork's x86_64
harness at it too if you'd like that before merge.)

creusot/src/ctx.rs
Lysxia: In #2230 I managed to avoid calling `typing_env` altogether for constructors (it is called in `sig` to normalize contracts, but constructors have no user-provided contracts to normalize).

Lysxia:
Subsumed by #2230

greeng3:
Thanks — #2230 is the better fix. Avoiding `typing_env` for constructors entirely is cleaner than the guard I put in `typing_env_with`: mine worked around the `opaque_types_defined_by` query being refused for a ctor, whereas yours removes the reason to ask in the first place.

Glad the review caught the type-invariant point too; that was the part I had wrong initially and it would have shipped an unprovable constructor for any type with a non-trivial invariant.

PR #2230, opened 2026-08-18
Support constructor closures

### LINKED ISSUE #2223: Support using variant constructors as function values (e.g., `.map(E::A)`)

Lysxia:
Close #2223

Generate a simplified body of the closure (directly call the continuation after asserting the expected invariant):

```
let c_Some [@coma:extspec] (x: a) (return (x: option_a) = {inv (Some x)} return {Some x}
```

Then reuse the pre- and post-conditions inferred by Coma (`c_Some'pre`, `c_Some'post'return`).

### DIFF
--- a/creusot/src/backend.rs
+++ b/creusot/src/backend.rs
@@ -91,6 +91,7 @@ impl<'tcx> Why3Generator<'tcx> {
                 format!("unsupported definition kind {:?} {:?}", def_id, dk),
             ),
             ItemType::Closure
+            | ItemType::CtorFn
             | ItemType::Trait
             | ItemType::Type
             | ItemType::AssocTy
--- a/creusot/src/backend/clone_map/elaborator.rs
+++ b/creusot/src/backend/clone_map/elaborator.rs
@@ -2,16 +2,17 @@ use crate::{
     backend::{
         self, Why3Generator,
         clone_map::{CloneNames, DepGraphBuilder, Dependency, Kind, Namer},
-        closures::{closure_hist_inv, closure_post, closure_pre},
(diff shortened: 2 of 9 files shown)


jhjourdan:
Out of curiosity, why did you not choose to synthesize the pre- and postconditions?

Lysxia:
Looking for an excuse to use `coma:extspec`.

PR #2233, opened 2026-08-20
tests: Use modular output for creusot-std

Lysxia:
This enables using why3find for most goals and why3 for the few remaining that are difficult to automate. I also added some `#[logic(inline)]` to automate some more of those.

While I was at it, I refactored the tests to remove a dependency (`assert_cmd`) and to allow multiple patterns to select test cases.

I've put the main logical changes in the first two commits, isolating the big diff of test files in the third commit.


creusot-std/src/std/iter/map_inv.rs
```diff
@@ -48,7 +48,7 @@ impl<I: IteratorSpec, B, F: FnMut(I::Item, Snapshot<Seq<I::Item>>) -> B> Iterato
 impl<I: IteratorSpec, B, F: FnMut(I::Item, Snapshot<Seq<I::Item>>) -> B> Invariant
     for MapInv<I, F>
 {
-    #[logic(open, prophetic)]
+    #[logic(open, inline, prophetic)]
```
jhjourdan, line 51: I'm not happy about this one. ` Self::reinitialize` and `Self::next_precondition` are also inline and have large definitions, so that each time the type invariant of `MapInv` appears, a large term is unfolded.

PR #2243, opened 2026-09-03
Guarded borrows

dianegolfouse:
Add borrows with an invariant, that
- is asserted (in particular) at the end of the borrow's lifetime
- can be broken locally, in particular across function boundary

```rust
let mut x = 1;
let guarded = GuardedBorrow::new(&mut x, snapshot!(|v| v == 1i32));

let borrow = guarded.borrow; // access the contained borrow, via a _place_: no invariant asserted
write_zero(borrow); // break the invariant, ok
*borrow = 1i32; // restore the invariant
```


examples/persistent_array.rs
```diff
@@ -164,22 +162,22 @@ pub mod implementation {
...
-                        proof_assert!(false)
-                    }
-                    pa.perms.insert_ghost(snapshot!(permcell), perm.into_inner());
-                    pa.depth = snapshot!(pa.depth.set(permcell, pa.depth[*self.permcell@] + 1));
-                    let mut frag = Fragment::new_unit(pa.auth.id_ghost());
-                    pa.auth.update(&mut frag, FMapInsertLocalUpdate(snapshot!(permcell), new_ag));
-                    frag
+            let frag = ghost! {
+                let pa = self.inv.open_guarded(tokens.into_inner());
+                let pa = &mut *pa.borrow;
```
jhjourdan, line 167:
```suggestion
                let pa = &mut *self.inv.open_guarded(tokens.into_inner()).borrow;
```

Perhaps we should have a macro for that? It looks like we are always going to do this after `open_guarded`.

creusot-std/src/ghost/invariant.rs
jhjourdan, line 545: Am I right that this can now be defined in terms of `open_guarded`?

jhjourdan:
That's very nice. I still need to do a more thorough review, but this should fix a lot of our issues.

In particular, could you make `Perm` sized? That would simplify a LOT of examples.

Also, could we have something like `GuardedFullBorrow`, a variant of `FullBorrow` in order to finally merge #2075?

Are there any other use of this feature we discussed?

creusot-std/src/ghost/invariant.rs
dianegolfouse: Unfortunately, I don't think so: something like
```rust
let guarded = ghost!(this.into_inner().open_guarded(tokens.into_inner()));
f(ghost!(guarded.borrow))
```
Fails borrow checking. Even `ghost_let` doesn't save us I think: `f` expects a lifetime `'a`, but `guarded.borrow` is necessarily bounded by the lifetime of `guarded` itself.

dianegolfouse:
Okay, as we discussed I changed the definition to have a `Guarded` type instead, that can accommodate multiple kinds of borrows. I only instantiated it with `&mut T` for now.
I also made permissions sized; @Lysxia can you check if this is still ok with the slice specs? We can also discuss this in person next week.

creusot-std/src/invariant/subset.rs
```diff
@@ -0,0 +1,196 @@
+use crate::prelude::*;
+#[cfg(creusot)]
+use crate::resolve::structural_resolve;
+use core::ops::{Deref, DerefMut};
+
+/// A type implements `InhabitedInvariants` when its type invariant is inhabited.
```
Lysxia:
```suggestion
/// A type implements `InhabitedInvariant` when its type invariant is inhabited.
```

creusot-std/src/std/slice.rs
```diff
@@ -121,11 +121,15 @@ impl<T> SliceExt<T> for [T] {
 
     /// Convert `&mut [T]` to `*mut T` and a mutable ownership token.
     #[check(terminates)]
-    #[ensures(result.0 as *const T == *result.1.ward() as *const T)]
-    #[ensures(&*self == result.1.val())]
-    #[ensures(&^self == (^result.1).val())]
+    #[ensures(exists<p: *const [T]>
+         p as *const T == result.0 as *const T &&
+         p.len_logic()@ == self@.len() &&
+         result.1.guard() == |b: &mut Perm<*const [T]>| *b.ward() == p
```
Lysxia: How about

```suggestion
    #[ensures(
         result.1.guard() == |b: &mut Perm<*const [T]>| *b.ward() as *const T == result.0 as *const T && b.ward().len_logic()@ == self@.len()
```

Lysxia (commented):
Looks good to me.

creusot-std/src/ghost/invariant.rs
jhjourdan: Hugh... But this is due to `Guarded` being `Drop`, right?

Seems like we need another hack to make sure that nobody destructures `Guarded`, because this is going to bite us again!

Could you please remind me why we are implementing `Drop` exactly? What is the risk?

creusot-std/src/invariant/guarded.rs
```diff
@@ -0,0 +1,108 @@
...
+    ///
+    /// The type invariant of the `Guarded` ensures that the current value
+    /// of [`borrow`](Self::borrow) satisfies this guard.
+    #[logic(open, inline)]
+    pub fn guard(self) -> Mapping<T, bool> {
+        *self._guard
+    }
+}
+
+impl<'a, T: ?Sized> Invariant for GuardedBorrow<'a, T> {
```
jhjourdan, line 31:
```suggestion
impl<'a, T: ?Sized> Invariant for Guarded<a, T> {
```

creusot-std/src/std/ptr.rs
```diff
@@ -596,12 +594,14 @@ impl<T: ?Sized> Perm<*const T> {
...
     /// ```
     #[trusted]
-    #[check(terminates)] // can overflow the number of available pointer adresses
-    #[ensures(*result.1.ward() == result.0)]
-    #[ensures(*result.1.val() == *r)]
-    #[ensures(*(^result.1.inner_logic()).val() == ^r)]
+    // can overflow the number of available pointer adresses
+    #[check(terminates)]
+    // Ensures that the permission is not stolen (via e.g. `std::mem::replace`)
+    #[ensures(result.1.guard() == |p: &mut Perm<*const T>| *p.ward() == result.0)]
```
jhjourdan, line 600: Shouldn't we say something about the prophecy of `p` too?

creusot-std/src/std/ptr.rs
```diff
@@ -729,15 +729,17 @@ impl<T> Perm<*const [T]> {
     #[trusted]
     #[check(ghost)]
     #[requires(0 <= index && index <= self.len())]
-    #[ensures(self.ward().thin() == result.0.ward().thin())]
-    #[ensures(self.ward().thin().offset_logic(index) == result.1.ward().thin())]
-    #[ensures(self.val()@[..index] == result.0.val()@)]
-    #[ensures(self.val()@[index..] == result.1.val()@)]
+    #[ensures(result.0.guard() == |p: &mut Perm<*const [T]>| p.ward().thin() == self.ward().thin())]
+    #[ensures(result.1.guard() == |p: &mut Perm<*const [T]>| p.ward().thin() == self.ward().thin().offset_logic(index))]
```
jhjourdan, line 733: Same here: shouldn't we say something about the prophecy of `p` too?

creusot-std/src/std/slice.rs
```diff
@@ -121,11 +121,14 @@ impl<T> SliceExt<T> for [T] {
...
     /// Convert `&mut [T]` to `*mut T` and a mutable ownership token.
     #[check(terminates)]
-    #[ensures(result.0 as *const T == *result.1.ward() as *const T)]
-    #[ensures(&*self == result.1.val())]
-    #[ensures(&^self == (^result.1).val())]
+    #[ensures(result.1.guard() == |b: &mut Perm<*const [T]>| {
+        *b.ward() as *const T == result.0 as *const T &&
+        b.ward().len_logic()@ == self@.len()
+     })]
+    #[ensures(*self == *result.1.borrow.val_unsized())]
```
jhjourdan, line 128: Same here.

PR #2244, opened 2026-09-04
Refactor `term()` method to streamline naming of parameters and other cleanups

Lysxia:
This refactoring prepares for the introduction of modes, mainly by centralizing the logic of renaming function parameters into Why3 `Ident`.

- `inputs_and_output` is memoized to keep names stable.
- the `ctx.terms` field now only contains memoized bodies of `#[logic]` functions. Other terms that appear in contracts are now either memoized and exported elsewhere, as part of `PreSignature` (for local idents) or `ExternSpec`.
- various minor refactors.


creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -841,27 +841,22 @@ fn postcondition_once_term<'tcx>(
...
-        typing_env,
-        EarlyBinder::bind(ctx.tcx, ctx.sig(Intrinsic::PostconditionOnce.get(ctx)).inputs[2].2),
-    );
-    let res = Term::var(result, ty_res);
+    assert_eq!(result.0.0, name::result());
+    let ty_self = self_.2;
+    let self_ = Term::var(self_.0, self_.2);
+    let args = Term::var(args.0, args.2);
+    let res = Term::var(result.0, result.2);
     match ty_self.kind() {
```
jhjourdan:
```suggestion
    match self_.ty.kind() {
```

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -912,35 +907,29 @@ fn postcondition_mut_term<'tcx>(
...
-        EarlyBinder::bind(ctx.tcx, ctx.sig(Intrinsic::PostconditionMut.get(ctx)).inputs[3].2),
-    );
-    let res = Term::var(result, ty_res);
+    assert_eq!(result.0.0, name::result());
+    let ty_self = self_.2;
+    let self_ = Term::var(self_.0, self_.2);
+    let args = Term::var(args.0, args.2);
+    let result_state = Term::var(result_state.0, result_state.2);
+    let res = Term::var(result.0, result.2);
     match ty_self.kind() {
```
jhjourdan:
```suggestion
    match self_.ty.kind() {
```

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -1002,27 +991,22 @@ fn postcondition_term<'tcx>(
...
-        typing_env,
-        EarlyBinder::bind(ctx.tcx, ctx.sig(Intrinsic::Postcondition.get(ctx)).inputs[2].2),
-    );
-    let res = Term::var(result, ty_res);
+    assert_eq!(result.0.0, name::result());
+    let ty_self = self_.2;
+    let self_ = Term::var(self_.0, self_.2);
+    let args = Term::var(args.0, args.2);
+    let res = Term::var(result.0, result.2);
     match ty_self.kind() {
```
jhjourdan:
```suggestion
    match self_.ty.kind() {
```

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -1093,30 +1077,30 @@ fn post_fndef<'tcx>(
...
+        )
     };
-    let ty_self = subst.type_at(1);
-    let self_ = Term::var(self_, ty_self);
-    let args = Term::var(args, subst.type_at(0));
-
+    let ty_self = self_.2;
+    let self_ = Term::var(self_.0, self_.2);
+    let args = Term::var(args.0, args.2);
     match ty_self.kind() {
```
jhjourdan:
```suggestion
    match self.ty.kind() {
```

creusot/src/backend/clone_map/elaborator.rs
```diff
@@ -1191,50 +1175,42 @@ fn pre_fndef<'tcx>(
...
-        panic!("hist_inv must have 2 arguments. This should not happen. Found: {bound:?}")
+    let &[self_, future] = inputs else {
+        unreachable!("hist_inv must have 2 arguments. This should not happen. Found: {inputs:?}")
     };
-    let ty_self = subst.type_at(1);
-
+    let ty_self = self_.2;
+    let self_ = Term::var(self_.0, self_.2);
+    let future = Term::var(future.0, future.2);
     match ty_self.kind() {
```
jhjourdan:
```suggestion
    match self_.ty.kind() {
```

Cargo.lock
```diff
@@ -1773,7 +1773,3 @@ dependencies = [
  "log",
  "simd-adler32",
 ]
-
-[[patch.unused]]
-name = "creusot-std"
-version = "0.14.0-dev"
```
jhjourdan: What is this?

I keep seeing this in my diffs locally, and I don't add it in my commits. Is this dangerous? What does it comes from?

tests/should_succeed/open_inv.coma
```diff
@@ -21,7 +21,7 @@ module M_impl_Clone_for_IsZero__clone (* <IsZero as std::clone::Clone> *)
   
   meta "rewrite_def" predicate inv_ref_IsZero
   
-  predicate postcondition_clone (self: ()) (args: Int32.t) (result: Int32.t) = let self_ = args in result = self_
+  predicate postcondition_clone (self: ()) (args: Int32.t) (result: Int32.t) = let self'0 = args in result = self'0
```
jhjourdan, line 24: This diff seems quite dangerous. It seems like the same identifier `self` is bound twice (the Why3 generator then uses a different name, but in Creusot, there is only one identifier).

I agree this should not be problematic in this specific case, but could you double-check that there is no other instance of this?

jhjourdan (approved):
Generally I approve these changes. This makes the handling of identifiers in logical terms clearer, which is desirable.

However, I won't be able to do an in-depth review of this PR, which is too large. I'll trust you on this one ;)

Cargo.lock
Lysxia: I'm not sure how this ended up in creusot but `[[patch.unused]]` is an undesired side effect of adding a `[[patch.crates-io]]` to the global `~/.cargo/config.toml`. Building a package that *doesn't* use `creusot-std` emits a warning about it and adds `[[patch.unused]]` to its `Cargo.lock`. When I have the time I want to look into why cargo does that and find a workaround.

tests/should_succeed/open_inv.coma
Lysxia: It's the `self` argument of `postcondition` vs the `self` argument of `clone` for `i32`, which are tagged with different integer (every `DefId` generates a fresh `Ident` for its arguments). The latter was previously named `self_`  that comes from the argument of the extern spec wrapper of `clone` rather than `clone` itself.

PR #2245, opened 2026-09-07
Add simplified version of `std::sync::mpmc`

lafeychine:


cargo-creusot/src/main.rs
jhjourdan, line 1: I guess this change should be reverted?

creusot-std/src/logic/ord.rs
```diff
@@ -352,3 +352,22 @@ impl<A: OrdLogic, B: OrdLogic> OrdLogic for (A, B) {
         let _ = B::lt_log_total;
     }
 }
+
+pub trait SemiLattice: PartialOrdLogic + Sized {
```
jhjourdan, line 356: Do we really need the `Sized` requirement? It is not very problematic, but it looks weird.

cargo-creusot/src/main.rs
lafeychine: Good catch, this was never intended to be pushed

creusot-std/src/logic/ord.rs
lafeychine: `fn join` returns a `Self`, hence needing the `Sized` requirement

